PR #6, opened 2026-05-02
feat: hetzner snapshots, mpp 402 gating, lease bearer, image tags

jparklev:


jparklev:
## Review-driven changes (commits `55faee4` → `1f6f3fa`)

Addressed the bulk of Codex/Gemini findings from internal review. Summary:

**P0 / fail-closed config**
- `paymentGuardFromEnv` now throws a typed `MppxConfigError` for malformed recipient/currency, missing `CRABBOX_MPP_SECRET_KEY`, or missing `CRABBOX_SESSION_SECRET`. New `paymentConfigured(env)` boolean is what `index.ts` uses to gate the unauth lease path — so a misconfigured env can no longer accidentally let unauthenticated `POST /v1/leases` through without payment.
- Added `worker/test/payments.test.ts` with explicit fail-closed cases for each invalid combination.

**Per-payer ownership**
- `index.ts` now extracts the payer's address from the credential's `source` (`did:pkh:eip155:<chain>:0x…`) and uses `mpp:<payer>` as the lease owner rather than `mpp:<recipient>`. Per-owner cost-cap (`CRABBOX_MAX_MONTHLY_USD_PER_OWNER`) now applies per-wallet, lease-bearer scope is naturally per-payer, and bookkeeping is correct.
- Added `worker/test/index.test.ts` covering payer extraction.

**Lease-bearer scope**
- `listLeases` / `listRuns` now filter by `requestLeaseAuth()` so a `cbxl_…` token can only list its own lease and runs.
- `usage` returns `403` to lease bearers entirely.
- `createRun` requires the lease exists, is visible (which already covers lease-bearer scope), and is `active`.

**Strict serverType for MPP-paid leases**
- New `LeaseConfig.strictServerType` flag set to `true` after a successful charge. Both Hetzner and AWS providers honor it by skipping the candidate fallback list — agent gets exactly the type they paid for, or the lease creation fails. Mitigates the "paid for `ccx63`, got `cx53`" scenario, at the cost of less capacity resilience.

**Replay-protection persistence**
- Plugged a `Store.AtomicStore` adapter backed by Durable Object storage into `tempo.charge`. mppx's consumed-transaction-hash markers now survive DO eviction/rehydration. Naive read-modify-write is safe because DO is single-threaded per object name.

**Go CLI bearer install**
- After `CreateLease`, the returned `bearer` is set as the coordinator client's `Token` if no admin/github bearer was already present. Heartbeat / release / runs from an MPP-only client now authenticate.

**Tests**
- 65 worker tests pass (was 49); Go tests pass; typecheck, oxlint, oxfmt clean; `wrangler deploy --dry-run` bundles 2.5 MB / 497 KB gzipped.

**Documented as known caveats (not fixed in this PR)**
- Charge-before-provisioning refund path: if Hetzner provisioning fails after a successful charge, the user is charged with no lease. Strict-serverType reduces but doesn't eliminate the window. A reconciliation table backed by DO storage is the natural follow-up.
- No real `mppx`-binary integration test for the Go shell-out path. Parser unit tests only.

PR #22, opened 2026-05-04
fix: use target-aware AWS instance defaults

vincentkoc:
## Summary
- recalculate default provider instance type after `--target`, `--provider`, or `--class` flag overrides
- fixes `--provider aws --target macos --class standard` preferring Linux `c7a.8xlarge` instead of `mac2.metal`
- keeps explicit `--type` / `CRABBOX_SERVER_TYPE` behavior unchanged
- documents `CRABBOX_AWS_MAC_HOST_ID` as required broker config for `target=macos`

## Root Cause
`loadConfig()` resolves `serverType` before command flags are applied. `run` and `warmup` recalculated the type after `--provider` or `--class`, but used `serverTypeForProviderClass`, which ignores target OS. That made macOS and Windows requests inherit Linux AWS defaults when flags changed class/provider/target.

## AWS Mac Proof
Before this PR, the merged local binary sent the wrong preferred type:

```text
coordinator lease class=standard preferred_type=c7a.8xlarge ...
coordinator POST /v1/leases: http 500: {"error":"c7a.8xlarge: aws target=macos requires CRABBOX_AWS_MAC_HOST_ID"}
```

With this branch binary, the CLI sends the correct Mac type:

```text
recording run run_0abf1aef651e
coordinator lease class=standard preferred_type=mac2.metal keep=false slug=brisk-crab idle_timeout=15m0s ttl=45m0s
coordinator POST /v1/leases: http 500: {"error":"mac2.metal: aws target=macos requires CRABBOX_AWS_MAC_HOST_ID"}
```

That means the CLI selection bug is fixed. Full AWS Mac leasing is still blocked on infra: the Cloudflare Worker needs `CRABBOX_AWS_MAC_HOST_ID` set to an allocated EC2 Mac Dedicated Host. I could not verify or set it locally because Wrangler has no `CLOUDFLARE_API_TOKEN` in this shell, and local AWS credentials return `InvalidClientTokenId`.

## Validation
- `go test ./internal/cli -run 'TestApplyServerTypeFlagOverrides|TestApplyCapacityMarketFlag|TestValidateProviderTarget|TestVersion'`
- `go test ./internal/cli -run TestAWSServerTypeForClass`
- `go test ./...`
- `go vet ./...`
- `git diff --check`

### DIFF
--- a/docs/infrastructure.md
+++ b/docs/infrastructure.md
@@ -172,6 +172,7 @@ Brokered AWS credentials live as Worker secrets:
 AWS_ACCESS_KEY_ID
 AWS_SECRET_ACCESS_KEY
 AWS_SESSION_TOKEN optional
+CRABBOX_AWS_MAC_HOST_ID required for brokered target=macos
 ```
 
 Direct fallback env is whatever the AWS SDK can resolve, such as:
@@ -296,6 +297,7 @@ HETZNER_TOKEN
 AWS_ACCESS_KEY_ID
 AWS_SECRET_ACCESS_KEY
 AWS_SESSION_TOKEN optional
+CRABBOX_AWS_MAC_HOST_ID required for brokered target=macos
 CRABBOX_SHARED_TOKEN
 CRABBOX_GITHUB_CLIENT_ID
 CRABBOX_GITHUB_CLIENT_SECRET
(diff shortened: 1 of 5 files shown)


steipete:
Closing this draft as superseded by current main. The target-aware AWS default selection and AWS Mac host docs from this branch are already present on main, and the remaining WSL2 mode-only default refresh case is tracked by https://github.com/openclaw/crabbox/pull/25. Thanks for the original fix and proof here.

PR #59, opened 2026-05-09
docs: add versioned workspace roadmap

stainlu:
## Summary
- add a versioned workspaces feature roadmap for checkpoint/fork/compare/promote workflows
- define the provider boundary for generic fallback checkpoints versus provider-native snapshots
- link the roadmap from feature docs and source map

## Verification
- node scripts/build-docs-site.mjs
- git diff --check

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 - Added live smoke checklists and script coverage for direct E2B and Semaphore provider validation. Thanks @stainlu.
+- Added the versioned workspace roadmap for checkpointing, forking, comparing, and promoting agent workspace attempts. Thanks @stainlu.
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


steipete:
shouldn't wem land the docs when this feature is done? :)

steipete (changes requested):
Not landable yet. Two concrete reasons:

1. This documents `Versioned workspaces` as a feature page before the checkpoint workflow exists in the CLI/coordinator. That will make the docs nav promise a user-visible feature we cannot run yet. Better to land this with the checkpoint command slice, or keep it as a design note that is clearly not shipped.
2. The command shape is internally inconsistent with the code direction in #61: this page shows `crabbox checkpoint create --id blue-lobster --name after-auth-fix`, while the store model validates checkpoint ids as `chk_...`. If `blue-lobster` is a lease slug, the flag should not be `--id`; if it is a checkpoint id, the example should use the checkpoint id format.

I would hold this until the feature PR defines the actual command contract, then land docs and implementation together.

stainlu:
Closing this per review. The roadmap content is better represented by the implementation plan in https://github.com/openclaw/crabbox/pull/63, which stays under docs/plan and does not promise a shipped command surface. Feature docs and exact checkpoint command examples should land with the actual checkpoint create/list/show slice so the --id versus chk_... contract is defined by implementation.

PR #61, opened 2026-05-09
feat: add checkpoint ledger store

stainlu:
## Summary
- add typed checkpoint records for repo, lease, run, workspace, artifacts, and provider snapshot references
- persist checkpoints as append-only JSON files under Crabbox state
- list checkpoints newest-first and reject duplicate or unsafe checkpoint ids

## Verification
- env GOCACHE=/private/tmp/crabbox-gocache go test ./internal/cli
- git diff --check

### DIFF
--- a/internal/cli/checkpoint_store.go
+++ b/internal/cli/checkpoint_store.go
@@ -119,11 +119,6 @@ func (s checkpointStore) Create(record CheckpointRecord) (CheckpointRecord, erro
 	if err := os.MkdirAll(filepath.Dir(path), 0o700); err != nil {
 		return CheckpointRecord{}, exit(2, "create checkpoint directory: %v", err)
 	}
-	if _, err := os.Stat(path); err == nil {
-		return CheckpointRecord{}, exit(2, "checkpoint %s already exists", record.ID)
-	} else if err != nil && !errors.Is(err, os.ErrNotExist) {
-		return CheckpointRecord{}, exit(2, "stat checkpoint %s: %v", path, err)
-	}
 	if err := writeCheckpointJSON(path, record); err != nil {
 		return CheckpointRecord{}, err
 	}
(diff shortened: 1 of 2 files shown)


steipete (changes requested):
Blocking correctness issue: checkpoint creation is not actually append-only under concurrent callers. `Create` checks `os.Stat(path)` before `writeCheckpointJSON`, but the write path ends with `os.Rename(tmpPath, path)`, which overwrites an existing file on POSIX. I reproduced this by adding a temporary concurrent same-ID regression: 64 goroutines calling `Create(CheckpointRecord{ID: "chk_race"})` all returned success.

Please make the final publish step atomic/exclusive, for example by writing to a temp file and then using an exclusive create/link/open path for the final `<id>.json`, or by opening the target with `O_CREATE|O_EXCL` after the JSON is prepared. Add the concurrent duplicate regression so we prove only one writer can win. Once that is fixed, the store shape is a reasonable first slice for the checkpoint feature.

stainlu:
Addressed the blocking concurrency review in 0a41369. Checkpoint writes now publish with O_CREATE|O_EXCL instead of stat-then-rename, so concurrent same-ID creates cannot overwrite an existing checkpoint. Added TestCheckpointStoreConcurrentSameIDAllowsOneWriter with 64 concurrent callers. Verification: env GOCACHE=/private/tmp/crabbox-gocache go test ./internal/cli; env GOCACHE=/private/tmp/crabbox-gocache go test -race ./internal/cli; git diff --check.

steipete (approved):
Reviewed after the rebase/fixups.

Proof:
- CodexReview clean on the final local changes before commit.
- Focused local tests passed: go test ./internal/cli -run TestCheckpointStore -count=1
- Focused local race test passed: go test -race ./internal/cli -run TestCheckpointStoreConcurrentSameIDAllowsOneWriter -count=1
- PR CI green on bca5f322ddf2aab402c666b7d4fbaee9c85baeb6: Docs, Go, Worker, Release Check.

The prior concerns are handled: checkpoint writes use exclusive create, the concurrent same-ID regression is covered, the branch compiles after main's checkpoint ID helpers, and the repo fixture is neutral.

PR #63, opened 2026-05-09
docs: add versioned workspace implementation plan

stainlu:
## Summary
- add an implementation plan for versioned agent workspaces
- define the boundary between core checkpoint workflow, coordinator storage, provider-native snapshots, and portal work
- link the plan from the docs entrypoint

## Verification
- node scripts/build-docs-site.mjs
- git diff --check


steipete:
ah merged by accident. pleaes do full PRs not one simply for future plans

PR #82, opened 2026-05-12
fix(worker): prune stale AWS SSH ingress

obviyus:
## Summary
- prune stale Crabbox-owned SSH ingress rules before adding the current AWS source CIDR
- keep IPv4 and IPv6 ingress parsing on one shared rule path
- add regression coverage for stale CIDR selection that can exhaust AWS security-group rule quota

## Verification
- `npm test --prefix worker -- aws.test.ts`

### DIFF
--- a/worker/src/aws.ts
+++ b/worker/src/aws.ts
@@ -32,6 +32,29 @@ export function createSecurityGroupParams(name: string, vpcID: string): Record<s
   };
 }
 
+type SSHIngressRule = {
+  cidr: string;
+  family: "ipv4" | "ipv6";
+  port: string;
+};
+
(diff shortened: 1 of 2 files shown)


steipete:
Thanks, this is landed on `main` as https://github.com/openclaw/crabbox/commit/5e6cff8.

I kept the fix and regression coverage, resolved it against the newer AWS cleanup tests already on `main`, and added the changelog credit in https://github.com/openclaw/crabbox/commit/a69c09b.

Verification:
- `npm test --prefix worker -- aws.test.ts`
- `go test ./...`
- `npm run docs:check`

Closing this PR because the fix is now on `main`.

PR #86, opened 2026-05-13
fix(hetzner): unmarshal private_net as array (best-effort) instead of crashing

muqsitnawaz:
## Problem

Every Hetzner-backed `crabbox` command fails on any account with at least one server:

```
$ HCLOUD_TOKEN=... crabbox list
json: cannot unmarshal array into Go struct field Server.servers.private_net of type struct { IPv4 struct { IP string "json:\"ip\"" } "json:\"ipv4\"" }
```

This breaks `list`, `doctor`, `warmup`, and `run --id ...` (the last one calls into the list path to look up the IP).

## Cause

Hetzner's API returns `private_net` as an **array of network attachments** (one entry per attached private network, empty when none) — see https://docs.hetzner.cloud/#servers-get-all-servers.

`Server.PrivateNet` in `internal/cli/hcloud.go:34` was modeled as a single struct (matching what Azure's and Proxmox's translated shapes look like), so `encoding/json` errors out on `"private_net": []` before any caller ever sees the result.

## Fix (best-effort)

Promoted `PrivateNet` to a named type with a custom `UnmarshalJSON` that accepts both shapes:

- **Hetzner's array shape** — best-effort: lifts the first attachment's `ip` into `PrivateNet.IPv4.IP`. Empty array stays empty. So any caller that wants a private-network IP for a Hetzner-leased box still gets one (even though the current Hetzner code path doesn't read it, this future-proofs the field).
- **Legacy struct shape** `{"ipv4": {"ip": "..."}}` — kept for anything that JSON-round-trips a `Server` outside the Hetzner API (test fixtures, golden files).
- **`null` / missing** — zero value, no error.

Azure / Proxmox are unaffected — they populate the field via direct field assignment (`azure.go:903`, `proxmox.go:605`), never JSON. Access patterns like `server.PrivateNet.IPv4.IP = ip` continue to work because the underlying struct shape is unchanged.

The earlier draft of this PR used `json:"-"` instead — that worked but dropped Hetzner's data on the floor. Switched to best-effort populate on review.

## Tests

Added `internal/cli/hcloud_test.go` with four regression cases:

- `TestServerUnmarshalsHetznerPrivateNetEmptyArray` — `"private_net": []` (the common shape on accounts with no private networks).
- `TestServerUnmarshalsHetznerPrivateNetAttached` — non-empty array, asserts the first attachment's IP lands in `PrivateNet.IPv4.IP`.
- `TestServerUnmarshalsLegacyPrivateNetStruct` — legacy `{"ipv4": {"ip": "..."}}` still works.
- `TestServerUnmarshalsPrivateNetNullAndOmitted` — `null` and missing field both yield zero-value, no error.

```
$ go test ./internal/cli/ -run TestServerUnmarshals -v
=== RUN   TestServerUnmarshalsHetznerPrivateNetEmptyArray
--- PASS (0.00s)
=== RUN   TestServerUnmarshalsHetznerPrivateNetAttached
--- PASS (0.00s)
=== RUN   TestServerUnmarshalsLegacyPrivateNetStruct
--- PASS (0.00s)
=== RUN   TestServerUnmarshalsPrivateNetNullAndOmitted
--- PASS (0.00s)
    --- PASS: omitted (0.00s)
    --- PASS: null (0.00s)
PASS
```

Full `go test ./...` is also clean (all 11 packages `ok`, no regressions in Azure / Proxmox / shared).

## Changelog

Added a `Fixed` entry under `## 0.13.1 - Unreleased`.

## Repro context

Found on an account with three running servers (one `mark` ccx33, two `crabbox-*` cpx62), all with `"private_net": []` in the API response. Stack came from `(*HetznerClient).do -> json.Unmarshal` at `hcloud.go:104`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 ## 0.13.1 - Unreleased
 
+### Fixed
+
+- Fixed Hetzner Cloud server-list parsing: `private_net` is returned by the
+  API as an array of network attachments (per Hetzner's documented schema),
+  but `Server.PrivateNet` was modeled as a struct, causing every Hetzner
+  command (`list`, `doctor`, `warmup`, `run --id ...`) to fail with
+  `json: cannot unmarshal array into Go struct field Server.servers.private_net ...`
+  as soon as any server existed in the account. Promoted `PrivateNet` to a
+  named type with a best-effort `UnmarshalJSON` that accepts both Hetzner's
+  array shape (lifts the first attachment's `ip` into `PrivateNet.IPv4.IP`)
+  and the legacy struct shape used elsewhere. Azure / Proxmox callers are
+  unaffected — they set the field via direct assignment. Regression tests
+  cover empty array, attached array, legacy struct, and null/omitted shapes.
+
 ## 0.13.0 - 2026-05-13
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 - Fixed native provider checkpoint creation so AWS, Azure, and GCP snapshot/image checkpoints flush source filesystem writes before calling the provider API.
 - Fixed Tensorlake timing JSON so delegated runs include the lease slug and reused sandboxes preserve the stored claim slug. Thanks @stainlu.
 - Fixed Tensorlake workdir validation so broad sandbox paths are rejected before sync or command execution. Thanks @stainlu.
+
 ## 0.13.0 - 2026-05-13
 
 ### Added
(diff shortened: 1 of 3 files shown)


beverm2391:
Confirmed this independently from a direct Hetzner Crabbox run, hit the same `private_net` unmarshal failure after server creation, and a patched `0.13.0-dev` build with this class of fix successfully provisioned direct Hetzner using `--script-stdin`, `--env-from-profile`, and `--allow-env OPENAI_API_KEY`, ran the workload, and released the lease cleanly.

PR #131, opened 2026-05-19
feat(scripts): add openclaw-config-compat.sh

RomneyDa:
## Summary

Adds `scripts/openclaw-config-compat.sh`, a four-phase smoke that takes a source-controlled OpenClaw config fixture, proves it loads on the version it was authored for, then surfaces whether `openclaw doctor` is required to keep it loading after an upgrade.

- Phase 1 installs OpenClaw X, stages the fixture into `~/.openclaw/openclaw.json` on the leased box, and runs a configurable smoke command (`openclaw doctor --lint --json` by default). If this fails the script exits 2 — the fixture is bad, no signal about Y was produced.
- Phase 2 upgrades to OpenClaw Y via `npm i -g openclaw@<y>` and re-runs the smoke command without touching doctor. The state-dir hash is captured here.
- Phase 3 runs `openclaw doctor --non-interactive --yes`, captures its output, and diffs the state-dir hash to report whether doctor actually mutated anything.
- Phase 4 re-runs the smoke command post-doctor.

Exit codes distinguish the four interesting outcomes: 0 = clean upgrade, 3 = migration required and doctor handled it, 4 = regression that doctor could not recover, 2 = bad fixture or infra error.

Everything sensitive (path quoting, the JSON fixture itself) goes over a base64-decoded stdin to the box, so fixtures with quotes or large payloads do not blow up the shell. The script gates on `CRABBOX_LIVE=1`, matches the convention of `scripts/openclaw-wsl2-tests.sh`, and stops the lease on exit unless `OPENCLAW_COMPAT_STOP=0` is set.

## Test plan

- [ ] `bash -n scripts/openclaw-config-compat.sh` — syntax check (done locally).
- [ ] Live smoke against an AWS Linux lease with a known-good fixture for the current OpenClaw release: expect exit 0 and "clean upgrade."
- [ ] Live smoke with a deliberately stale fixture against a newer OpenClaw: expect either exit 3 (doctor migrates) or exit 4 (regression) with the doctor diff in the output.
- [ ] Live smoke with a hand-broken fixture against its own baseline: expect exit 2 ("fixture rejected by its own baseline").
- [ ] `CRABBOX_LIVE` unset: expect exit 2 with the live-mode reminder.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Added a thin macOS developer-tools image mint wrapper that keeps paid host allocation explicit while wiring the reusable prep script, promotion, checkpoint proof, and lifecycle evidence defaults.
 - Added a light/dark mode toggle to the Crabbox documentation site that defaults to the system color scheme, persists the choice in local storage, and applies before first paint to avoid a flash.
+- Added `scripts/openclaw-config-compat.sh`, an OpenClaw config compatibility smoke that leases a Linux box, installs a baseline OpenClaw version, stages a source-controlled config snapshot, then upgrades to a target version and reports whether `openclaw doctor` is required to migrate the config.
 
 ### Changed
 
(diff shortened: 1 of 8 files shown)


steipete:
Thanks for putting this together. I’m going to close this because it is not a fit for Crabbox.

Crabbox needs to stay a generic remote testing and execution tool. This PR adds OpenClaw-specific fixtures and migration semantics, which would make Crabbox carry another product’s compatibility matrix and maintenance burden.

The underlying workflow is useful, but it should live in OpenClaw, manager, or an external maintainer scripts repo as a Crabbox-powered smoke test. Crabbox can support the generic execution pieces, but should not own OpenClaw-specific config fixtures or upgrade logic.

RomneyDa:
Makes sense, looking to add similar to Kova.

PR #133, opened 2026-05-19
feat: add runpod provider

zozo123:
## Summary

Adds `provider: runpod` as an SSH-lease provider. Crabbox creates a CPU pod through the RunPod GraphQL API (`deployCpuPod`), polls `pod.runtime.ports[]` for the public TCP mapping to port 22, and hands callers a normal SSH lease. After that, sync, run, ssh, status, and stop reuse Crabbox's existing SSH transports.

## How it works

- `Acquire` -> `deployCpuPod(input: { startSsh: true, ports: "22/tcp", instanceId, image, containerDiskInGb, cloudType })`, then waits for `pod.runtime.ports[]` to publish `privatePort:22, isIpPublic:true` and a non-zero `publicPort`.
- `Resolve` / `List` use `pod(input: {podId})` and `myself { pods { ... } }`. List filters to `crabbox-` prefix unless `--all` is set.
- `ReleaseLease` calls `podTerminate(input: {podId})`.
- `Doctor` runs `myself` (auth) plus `myself.pods` (inventory) without ever invoking a mutation.

## Setup

```sh
export RUNPOD_API_KEY=...   # upload an ED25519 public key under RunPod settings once
crabbox doctor --provider runpod
crabbox run --provider runpod -- pnpm test
```

`CRABBOX_RUNPOD_API_KEY` overrides `RUNPOD_API_KEY`. Token is never accepted as a flag.

## Why this pattern

RunPod returns public SSH coordinates on pod create, so the SSH-lease shape mirrors `exe-dev` and lets Crabbox reuse its existing SSH sync/run/status/ssh/stop paths instead of growing a parallel verb tree.

## Tested

- Unit: `go test -race ./internal/providers/runpod/...` (16 tests, all pass), `go vet`, `gofmt` clean.
- Registry: `go test -race ./internal/providers/all/...` passes with `runpod` in the doctor-implementing list.
- Live: `RUNPOD_API_KEY=... crabbox doctor --provider runpod` returns `auth=ready control_plane=ready inventory=ready` against the real GraphQL endpoint. Live `crabbox run` exercises the `deployCpuPod` mutation and is correctly rejected with "Your account balance is too low to rent a pod" - the schema/mutation/input casing are validated but a funded RunPod account is required to actually boot a pod.

## Follow-ups

- Add GPU flavor support (`deployOnDemand` + `gpuTypeId`) once there is a Crabbox use case for paid GPU SKUs.
- Wire a `crabbox doctor --doctor-probe-ssh` check that opens the public port reported by `runtime.ports[]`.
- Plumb a coordinator-side broker path so teams can share a single funded RunPod account without distributing the API key.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
+- Added `provider: runpod` for RunPod CPU pod SSH leases through the RunPod GraphQL API, including Crabbox sync/run, `crabbox ssh`, `crabbox doctor`, and provider docs.
 - Added a thin macOS developer-tools image mint wrapper that keeps paid host allocation explicit while wiring the reusable prep script, promotion, checkpoint proof, and lifecycle evidence defaults.
 - Added AWS Linux and Windows developer-image prep scripts plus a guarded mint wrapper for baking Docker, Node 24, pnpm, GitHub CLI, and common developer tooling into fast-booting Crabbox AMIs.
 - Added a light/dark mode toggle to the Crabbox documentation site that defaults to the system color scheme, persists the choice in local storage, and applies before first paint to avoid a flash.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Added
 
-- Added `provider: runpod` for RunPod CPU pod SSH leases through the RunPod GraphQL API, including Crabbox sync/run, `crabbox ssh`, `crabbox doctor`, and provider docs.
+- Added `provider: runpod` for RunPod public TCP SSH leases through the RunPod REST API, including Crabbox sync/run, `crabbox ssh`, `crabbox doctor`, and provider docs. Thanks @zozo123.
 - Added a thin macOS developer-tools image mint wrapper that keeps paid host allocation explicit while wiring the reusable prep script, promotion, checkpoint proof, and lifecycle evidence defaults.
 - Added AWS Linux and Windows developer-image prep scripts plus a guarded mint wrapper for baking Docker, Node 24, pnpm, GitHub CLI, and common developer tooling into fast-booting Crabbox AMIs.
 - Added a light/dark mode toggle to the Crabbox documentation site that defaults to the system color scheme, persists the choice in local storage, and applies before first paint to avoid a flash.
--- a/README.md
+++ b/README.md
@@ -60,6 +60,9 @@ Supported providers:
 - [Railway](docs/providers/railway.md) (`provider: railway`): delegated
   redeploy-and-stream execution against a pre-existing Railway service through
   the [Railway](https://railway.com) GraphQL API.
+- [RunPod](docs/providers/runpod.md) (`provider: runpod`): RunPod CPU pods
+  provisioned through the [RunPod](https://runpod.io) GraphQL API and exposed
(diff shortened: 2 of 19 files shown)


steipete (changes requested):
Blocking on live E2E.

I tested this with a funded RunPod account, a newly created RunPod API key, and an uploaded ED25519 public key matching the local Crabbox SSH key. The API key is stored in 1Password; no token values are in logs.

What passed:
- `go test -race ./internal/providers/runpod ./internal/providers/all`
- `go vet ./...`
- `crabbox doctor --provider runpod` against the live RunPod API: auth/control-plane/inventory ready

What failed:
- `crabbox run --provider runpod -- echo runpod-live-ok`
- Crabbox created pod `cu5xd9uz13tr83` / `crabbox-golden-crab-5bda6dd3`, but `pod.runtime.ports` never exposed a public port-22 mapping and the provider timed out after 10 minutes.
- During the wait, `crabbox list --provider runpod --json` showed the pod as running but with empty SSH host/port fields.
- The failure path did terminate the pod correctly; follow-up inventory was empty.

I also probed the API directly:
- GraphQL `deployCpuPod` with the current shape produced running CPU pods but zero runtime ports, even with `supportPublicIp: true` and with `cloudType: SECURE`.
- RunPod's current REST docs describe pod creation through `POST /v1/pods` and expose `publicIp` plus `portMappings`, while the SSH docs distinguish basic proxied SSH from full SSH. Basic SSH goes through `ssh.runpod.io` and does not support SCP/SFTP, so it is not enough for Crabbox's normal SSH sync path. Full SSH needs public IP, exposed TCP 22, and sshd in the container.

Given that, the current provider shape does not yet prove the advertised capability: a normal SSH lease with Crabbox sync/run. Please rework the live path so it reliably creates a pod with full SSH coordinates, or narrow the PR to a delegated/proxied execution model that does not claim Crabbox sync over SSH.

steipete:
Updated this PR with maintainer fixes:

- switched RunPod from the old GraphQL CPU path to the REST `/v1/pods` API
- require public TCP SSH; RunPod basic SSH proxy is intentionally not used because rsync needs SCP/SFTP support
- added GPU capacity fallback attempts for current RunPod stock churn
- bootstraps missing `git`/`rsync`/`tar` after SSH auth, before Crabbox sync
- updated docs, defaults, and changelog with contributor credit

Proof run from this branch:

- `go test ./internal/providers/runpod`
- `go test ./internal/providers/runpod ./internal/cli`
- `go test -race ./internal/providers/runpod ./internal/providers/all`
- `go vet ./...`
- `go test ./...`
- live e2e: `crabbox run --provider runpod -- echo runpod-live-ok`

Live e2e provisioned pod `04pwlts5rl7o8y`, exposed public SSH at `213.192.2.107:40178`, synced 474 files / 5.2 MiB via rsync, printed `runpod-live-ok`, and auto-released the lease. Follow-up inventory was empty.

Temporary RunPod API key and SSH public key used for proof were revoked/removed after the test.

steipete:
@codex review

steipete (approved):
Live e2e now passes on the pushed REST/public-TCP implementation. Approved after local tests, live RunPod proof, cleanup, and external Codex review trigger.

steipete:
@codex review

steipete:
@codex review

PR #168, opened 2026-05-27
feat: add delegated run session output

RomneyDa:
This is to allow more robust automations that use `--keep` without parsing log files/terminal output

## Summary
- Add `crabbox run --lease-output <file>` as a generic delegated-run session output path.
- Add a core `RunSessionHandle`/`FeatureRunSession` contract and core JSON writer, with Blacksmith Testbox as the first producer.
- Document the reusable Blacksmith session lifecycle and add regression coverage for kept/reused Testboxes, unsupported providers, core JSON writing, and invalid output paths.

## Why this is needed
Blacksmith Testbox reuse already existed in a manual form: `crabbox run --keep --provider blacksmith-testbox ...` could leave a Testbox alive, and a later `crabbox run --id tbx_... --provider blacksmith-testbox ...` could reuse it.

The missing piece was a stable way for automation to get that `tbx_...` session handle. Callers had to scrape human terminal output, infer whether the Testbox was kept, construct cleanup themselves, and separately correlate any Actions run URL. That is brittle for release validators, auditors, or other orchestration scripts that need to pass one hydrated remote proof context through multiple steps.

## Design
The implementation keeps output writing in core instead of Blacksmith-specific file code:

- `RunResult.Session` carries a `RunSessionHandle`.
- `FeatureRunSession` lets providers explicitly advertise that they return this handle.
- `run.go` owns `--lease-output` validation and JSON writing.
- Blacksmith populates the handle today.
- A TODO marks the extension point for other reusable delegated providers to populate `RunResult.Session` and advertise `FeatureRunSession` once their lifecycle contract is covered.

This keeps the PR small while avoiding a Blacksmith-only output implementation that would need to be undone later.

## What this enables
A tool can now:

1. Start one hydrated Blacksmith Testbox and keep it alive.
2. Read the exact `leaseId` from JSON instead of parsing console output.
3. Run follow-up checks against the same remote workspace with `--id`.
4. Hand the same session/proof context to another process, including any detected Actions URL/run ID.
5. Finalize the session using the included non-secret cleanup command.

This does not change default one-shot behavior. If `--lease-output` is not passed, ordinary `crabbox run --provider blacksmith-testbox ...` still warms, runs one command, and stops the Testbox unless `--keep` or failure-retention policy applies.

## Example usage
```sh
crabbox run --provider blacksmith-testbox --keep --lease-output /tmp/session.json -- npm test
lease_id="$(node -e 'console.log(require("/tmp/session.json").leaseId)')"
crabbox run --provider blacksmith-testbox --id "$lease_id" -- npm run smoke
crabbox stop --provider blacksmith-testbox "$lease_id"
```

Example JSON:

```json
{
  "provider": "blacksmith-testbox",
  "leaseId": "tbx_abc123",
  "slug": "blue-lobster",
  "reused": false,
  "kept": true,
  "actionsUrl": "https://github.com/example-org/my-app/actions/runs/123456789",
  "runId": "123456789",
  "cleanupCommand": "crabbox stop --provider blacksmith-testbox tbx_abc123"
}
```

## Verification
- `go test ./internal/providers/blacksmith`
- `go test ./internal/cli -run 'TestRunCommandPreflightsLocalOutputOptions|TestRunCommandRejectsUnsupportedDelegatedCaptureOptions|TestWriteRunLeaseOutput|TestProviderFlagsApplyDaytonaAndIsloWithoutCoreEdits'`
- `go test ./...`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Added a portable `--os`/`os` lease selector with Ubuntu 26.04 as the preferred Linux image where provider catalogs support it, while preserving explicit provider image overrides.
 - Added Azure `capacity.regions` fallback with region-scoped managed network names and Azure capacity hints, matching the AWS capacity-routing model.
 - Added a repo-local Crabbox hydrate workflow and documented Azure as the preferred Windows/WSL2 provider when Azure quota or credits are available.
+- Added `crabbox run --lease-output <file>` for reusable delegated-run session JSON, starting with Blacksmith Testbox. Thanks @RomneyDa.
 
 ### Fixed
 
--- a/docs/commands/run.md
+++ b/docs/commands/run.md
@@ -37,7 +37,11 @@ crabbox run --profile live-qa --preset qa-live --scenario login-regression --emi
 
 If `--id` is omitted, Crabbox creates a fresh non-kept lease and releases it when the command exits. On coordinator-backed one-shot runs, if SSH becomes unavailable after a successful sync but before the command starts, Crabbox stops that stale lease, creates one replacement lease, and retries sync once. It does not replace explicit `--id`, kept, `--keep-on-failure`, `--no-sync`, `--sync-only`, or custom-slug runs. `--id` accepts the stable `cbx_...` ID or the active friendly slug.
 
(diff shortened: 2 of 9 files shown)


RomneyDa:
@clawsweeper re-review

RomneyDa:
@clawsweeper re-review

PR #199, opened 2026-06-03
feat: pause/resume commands via optional PausableBackend (islo first)

zozo123:
## What

Adds a generic, opt-in **pause/resume** capability to crabbox and implements it for the **islo** provider, which natively supports snapshot-to-disk pause + resume.

## Design (right altitude, not a one-off)

Mirrors crabbox's existing optional-interface pattern (`DoctorBackend`, `CleanupBackend`):

- **`PausableBackend`** interface (`Backend` + `Pause`/`Resume`) plus `PauseRequest`/`ResumeRequest` in `internal/cli/provider_backend.go`.
- **`crabbox pause` / `crabbox resume`** commands that reuse `stop`'s flag parsing and provider resolution, then dispatch via a type assertion. Providers that don't implement the interface return a clear `provider=<name> does not support pause` — no behavior change for anyone else.
- A compile-time assertion (`var _ core.PausableBackend = (*isloBackend)(nil)`) guards signature drift.

## islo implementation

`Pause`/`Resume` call the Islo SDK's `PauseSandbox`/`ResumeSandbox` (which return a decoded `SandboxResponse`), resolve the lease like `stop`, and **keep the lease claim across a pause** so it can be resumed.

## Tests

`TestIsloPauseResumeCallProvider` (backend dispatches to the provider with the resolved sandbox name; non-Crabbox ids rejected). `go build ./...`, `go vet`, `gofmt`, and the full `internal/cli` + `internal/providers/islo` suites pass.

## Verified live (api.islo.dev)

```
warmup → status: state=running ready=true
pause  → paused lease=…        → status: state=paused  ready=false
resume → resumed lease=…       → status: state=running ready=true
```


steipete:
Maintainer verification on June 11, 2026:

- `go test -race ./...`
- `go vet ./...`
- `node --test scripts/*.test.js` (126 passed)
- `scripts/check-docs.sh`
- fresh CLI build and `crabbox providers --json` capability check
- structured autoreview: clean, no actionable findings
- fresh GitHub CI: Go, Worker, Scripts, Docs, Apple VZ, Release Check, and Socket checks all passed

Live Islo lifecycle:

1. Stored a one-day test API key in 1Password before use.
2. `doctor --provider islo --json`: passed.
3. Islo OAuth-created a Crabbox-owned sandbox because Islo's create endpoint timed out only for the freshly minted API key. The normal isolated Crabbox claim was then created through the same core claim API used by warmup.
4. `status`: `running`, `ready=true`.
5. `pause`: `paused`, `ready=false`; local claim remained present.
6. `resume`: returned to `running`, `ready=true`.
7. `run --no-sync`: emitted `crabbox-islo-resume-ok`.
8. `stop`: local claim removed; sandbox absence confirmed after Islo list convergence.
9. Temporary API key revoked, 1Password item archived, final Islo inventory empty.

The API-key create timeout is provider-side/current-environment behavior and predates this pause/resume patch; the commands and adapter added here passed live through the API-key path.

PR #203, opened 2026-06-03
fix: restrict lease metadata writes

zozo123:
## Summary
- Require owner/manage/admin access for `POST /v1/leases/:id/heartbeat` and `/tailscale`.
- Add regression coverage that shared `use` viewers cannot mutate lease metadata.
- Keep sync archive creation in Go instead of local tar option parsing.

## Security
A shared `use` viewer could mutate visible leases: extend heartbeat expiry, refresh provider access from their source IP, or overwrite Tailscale metadata used for routing. These write paths now require manage-capable access.

## Verification
- `npm test --prefix worker`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/internal/cli/sync_archive.go
+++ b/internal/cli/sync_archive.go
@@ -89,8 +89,32 @@ func appendSyncArchiveMember(ctx context.Context, tw *tar.Writer, root, rel stri
 		return fmt.Errorf("open sync path %s: %w", rel, err)
 	}
 	defer file.Close()
-	if _, err := io.Copy(tw, file); err != nil {
+	if err := copySyncArchiveMember(ctx, tw, file); err != nil {
 		return fmt.Errorf("archive path %s: %w", rel, err)
 	}
 	return nil
+}
+
+func copySyncArchiveMember(ctx context.Context, dst io.Writer, src io.Reader) error {
+	buf := make([]byte, 128*1024)
+	for {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		n, readErr := src.Read(buf)
+		if n > 0 {
+			if err := ctx.Err(); err != nil {
+				return err
+			}
+			if _, err := dst.Write(buf[:n]); err != nil {
+				return err
+			}
+		}
+		if readErr == io.EOF {
+			return nil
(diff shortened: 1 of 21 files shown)


steipete:
Maintainer proof for the rebased branch at `c9708f3`:

- Removed the contributor changelog commit during rebase; release notes stay maintainer-owned.
- Kept the metadata-write hardening and sync archive hardening, then added `c9708f3` so the new in-process archive copy observes context cancellation during large-file copies.
- Worker regression route proof: `npm test --prefix worker -- fleet.test.ts` passed 139 tests, including shared `use` viewer 403s for heartbeat and Tailscale metadata writes without mutating stored lease state.
- Live sync proof on AWS native Windows: `run_3b8a33c07f3f`, lease `cbx_72cfe3791f0f`, slug `amber-prawn`, type `m7i.large`; synced a root file named `--checkpoint=crabbox-pr203` as data (`dirty_delta=1 files, 38 B`), remote command printed `CRABBOX_PR203_OPTION_FILENAME_PAYLOAD` and `CRABBOX_PR203_WINDOWS_ARCHIVE_SYNC_OK`, exit 0, `leaseStopped=true`; follow-up `status` returned 404 and `list` had no matching lease.
- Local gates: `go test ./...`, `go vet ./...`, `npm run format:check --prefix worker`, `npm run lint --prefix worker`, `npm run check --prefix worker`, `npm test --prefix worker`, `node scripts/build-docs-site.mjs`.
- Autoreview: clean after the cancellation fix; no accepted/actionable findings.

Note: I attempted a no-spend local Wrangler HTTP proof for the coordinator route, but the fake-provider create path cannot seed a lease cleanly; I did not pass live cloud secrets to Wrangler command-line args. The route behavior is covered by the Worker regression test above, and the archive path has live Windows E2E proof.

steipete:
@clawsweeper re-review

steipete:
Maintainer call: accepting the compatibility change that shared `use` viewers can no longer write heartbeat or Tailscale lease metadata. That role should remain read/use scoped; metadata writes require owner, shared `manage`, or admin access.

PR #210, opened 2026-06-04
feat(sync): add sync.include whitelist

anagnorisis2peripeteia:
## Summary

Sync currently supports only **excludes** (a blacklist). For a large repo where you only want a few paths in the box — e.g. a proof/test job that needs just `src/` and `scripts/` out of a working tree full of unrelated data — blacklisting every junk directory is fragile and endless (every new untracked dir reappears in the sync).

This adds **`sync.include`**: when set, **only** paths matching an include pattern are synced (after excludes).

## How it works

- `SyncConfig.Includes` + YAML `sync.include` / `sync.includes`.
- Filtering happens **inside** the manifest walker (`syncManifestFiltered`), so the size preflight sees the trimmed set — a 136 GiB working tree becomes the few hundred MB you actually selected, instead of being rejected by the large-sync guard.
- `syncManifest(root, excludes)` stays a thin wrapper over `syncManifestFiltered(root, excludes, nil)`, so all existing callers/tests are unchanged.
- Deleted/changed walkers honor includes too (accurate delta + byte accounting).
- Matching mirrors excludes (`pathIncluded`): an include of `src` keeps `src` and everything under it, a file path keeps that exact file, and globs match.

## Example

```yaml
sync:
  include:
    - src
    - scripts
    - package.json
    - pnpm-lock.yaml
```

## Test plan

- [x] `TestSyncManifestIncludeWhitelist` — only included paths land in the manifest; `data/`, `notes.txt` dropped.
- [x] `TestPathIncluded` — empty includes keep everything; dir/file/glob matching; non-matches excluded.
- [x] `go build` / `go vet` / existing sync + config tests pass.

Follow-up (not in this PR): per-job `sync.include` override and an optional `.crabboxinclude` file (parallel to `.crabboxignore`).

---

## Real behavior proof (`crabbox sync-plan`)

Run against a 136 GiB working tree (an openclaw checkout with large unrelated data dirs), showing `sync.include` trimming the candidate set. Redacted to top entries:

**Baseline — no include (whole working tree):**
```
$ crabbox sync-plan -limit 5
sync candidate: 80610 files, 136.0 GiB
top files:
  50.0 GiB   data/polymarket/data.tar
  33.5 GiB   data/polymarket/data.tar.zst
  525.2 MiB  tmp/scx-data.tar.gz
  ...
```

**With `sync.include: [scripts, crabbox.yaml]`:**
```
$ CRABBOX_CONFIG=with-include.yaml crabbox sync-plan -limit 5
sync candidate: 1401 files, 88.3 MiB
top files:
  12.4 MiB   scripts/polymarket_backtest/output/intraday_markets.parquet
  7.4 MiB    scripts/peporacle.mp4
  ...
```

**80,610 files / 136.0 GiB → 1,401 files / 88.3 MiB**, keeping only the whitelisted paths — the large `data/`, `tmp/`, `.archive*/` entries are gone. Without the whitelist this tree is rejected by the large-sync guard; with it the sync is well under the limit.

Also added: `sync.include` now appears in `crabbox config show` (text + `--json`) so the effective whitelist is visible during troubleshooting.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -308,6 +308,13 @@ sync:
     - node_modules
     - .turbo
     - dist
+  # include (whitelist): when set, ONLY these paths are synced (after excludes).
+  # Sync a few paths out of a large repo instead of blacklisting everything else.
+  include:
+    - src
+    - scripts
+    - package.json
+    - pnpm-lock.yaml
 env:
   allow:
     - CI
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -308,7 +308,7 @@ sync:
     - node_modules
     - .turbo
     - dist
-  # include (whitelist): when set, ONLY these paths are synced (after excludes).
+  # include (root-relative whitelist): when set, ONLY these paths are synced (after excludes).
   # Sync a few paths out of a large repo instead of blacklisting everything else.
   include:
     - src
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -119,9 +119,9 @@ matter for your repo:
 - `profile`: a name for this lane (the template uses `<repo>-check`);
 - `class`: `standard`, `fast`, `large`, or `beast` (the template uses `beast`);
 - `sync.exclude`: directories that should never be sent to the runner;
-- `sync.include`: an optional whitelist — when set, **only** these paths are
-  synced (after excludes), so you can ship a few paths out of a large repo
-  instead of blacklisting everything else;
+- `sync.include`: an optional root-relative whitelist — when set, **only** these
+  paths are synced (after excludes), so you can ship a few paths out of a large
+  repo instead of blacklisting everything else;
 - `env.allow`: environment variables the remote command is allowed to see.
 
 Pass `--detect` to scan the repo for test commands and write a `jobs.detected`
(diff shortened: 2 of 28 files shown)


anagnorisis2peripeteia:
@clawsweeper review

anagnorisis2peripeteia:
@clawsweeper re-review

steipete:
Landed in https://github.com/openclaw/crabbox/commit/1d4c37e77ec787b7cc9b3b8ada42d3bd512198e1. Thanks @anagnorisis2peripeteia.

Proof before merge:
- `go test ./internal/cli`
- `go test ./...`
- `bash scripts/check-docs.sh`
- `git diff --check origin/main...HEAD`
- autoreview clean: no accepted/actionable findings
- CI green on head `4ee339bcba7f7d34f4f0ab25fee405b19112f6a4`: Docs, Go, Release Check, Scripts, Worker, Socket checks

Maintainer fixups added before merge:
- disabled git seeding when `sync.include` is active so fresh runs honor the whitelist
- made include matching root-relative to avoid over-including matching names under unrelated paths
- applied include filtering to archive-sync providers
- allowed remote sync finalization to reconcile deletions for include-filtered syncs

PR #211, opened 2026-06-05
feat: add experimental --provider wasi delegated-run provider (wazero primary, wasmtime opt-in)

zozo123:
## Summary

Adds an experimental `wasi` delegated-run provider, with `wasm` and `wazero` aliases, for running WASI modules through Crabbox.

The provider uses embedded wazero by default and can use the `wasmtime` CLI when requested. It syncs the repository manifest into an isolated guest workdir, exposes that workdir to modules as `/work`, and supports `run`, `warmup`, `status`, `stop`, `list`, `doctor`, reusable run-session handles, and delegated proof output.

## Hardening in this revision

- Validates WASI workdir and runtime config from flags/config/env.
- Makes pure no-FS builtins (`echo`, `true`, `false`) actually skip manifest sync.
- Keeps host-side builtins (`ls`, `cat`) confined to `/work`, including symlink resolution checks.
- Fails sync on copy errors instead of silently producing a partial guest tree.
- Rejects repo symlinks that resolve outside the repo before copying into the guest.
- Wires `list`, `status`, and `doctor` to local lease claims.
- Removes the local PR handoff file and tightens docs to avoid unsupported capsule-specific claims.

## Verification

- `go vet ./...`
- `go test -race ./...`
- `./scripts/check-docs.sh`
- `npm ci --prefix worker`
- `npm test --prefix worker`
- Binary smokes:
  - `bin/crabbox providers --json` includes `wasi` with aliases `wasm`, `wazero`.
  - `bin/crabbox doctor --provider wasi --json` succeeds.
  - `bin/crabbox run --provider wasi -- echo crabbox-wasi-builtin-ok` reports `sync_skipped=true`.
  - Fresh temp repo builds a `GOOS=wasip1 GOARCH=wasm` module and runs it through `bin/crabbox run --provider wasi --allow-env CRABBOX_WASI_SMOKE -- ./main.wasm argv-ok`, proving `/work` sync, args, and explicit env forwarding.
  - `bin/crabbox run --provider wasi -- cat ../secret.txt` fails closed with an `/work` boundary error.
  - `bin/crabbox run --provider wasi --keep --slug wasi-final -- echo final-ok`, then `status`, then `stop` succeeds.

## Notes

WASI is intentionally narrow. It is useful for Wasm-native artifacts and capability-scoped execution, but not a replacement for full Linux providers when a workload needs package managers, native toolchains, SSH, browsers, or interactive services.

## Live proof (updated after symlink safety fix + tests)

Real terminal output from a wasip1 module executed via the provider (absolute .wasm path; run from a git checkout for manifest):

```
$ bin/crabbox run --provider wasi -- /tmp/.../proof.wasm agent-demo arg2
sync candidate: 603 files, 7.5 MiB
wasi sync complete in 114ms (guest=...)
running on wasi /tmp/.../proof.wasm agent-demo arg2
LIVE PROOF from WASI module: [agent-demo arg2]
Guest /work sees:
  .agents
  .crabbox.yaml
  .github
  ...
wasi run summary sync=114ms command=613ms total=838ms exit=0
```

(See also `TestWasiCopyManifestPreservesSymlinksWithoutLeakingIgnoredTargets` and the minimal-tree variant in CI logs.)

This demonstrates: module execution, arg passing, /work preopen with actual tree contents, and (via unit test) that symlinks to excluded targets no longer leak content.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,7 @@ configured); every other provider always runs direct from the CLI.
 | [Azure Dynamic Sessions](docs/providers/azure-dynamic-sessions.md) — `azure-dynamic-sessions` | Linux | Azure Container Apps dynamic sessions. |
 | [Blacksmith Testbox](docs/providers/blacksmith-testbox.md) — `blacksmith-testbox` (`blacksmith`) | Linux | Delegated Blacksmith CI Testbox lifecycle and execution. |
 | [W&B Sandboxes](docs/providers/wandb.md) — `wandb` (`weights-and-biases`) | Linux | Weights & Biases Sandboxes; reuses `wandb login` credentials. |
+| [WASI (experimental)](docs/providers/wasi.md) — `wasi` (`wasm`, `wazero`) | Wasm (WASI preview1/2) | Specialized Wasm sandbox (wazero/wasmtime) for WASI modules and Wasm-first code. Experimental. |
 
 See [Providers](docs/providers/README.md) for the full reference, capabilities,
 and authoring guide.
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -210,7 +210,8 @@ transport to the provider's own API or CLI. For these, SSH-specific options
 (`ssh`, `desktop`, `vnc`, `code`, Actions hydration, `--checksum`, `--sync-only`)
 are unavailable or partly restricted, and sync timing is reported as `delegated`.
 Examples include `blacksmith-testbox`, `azure-dynamic-sessions`, `e2b`, `modal`,
-`islo`, `cloudflare`, `upstash-box`, `tensorlake`, and `wandb`. See
+`islo`, `cloudflare`, `upstash-box`, `tensorlake`, `wandb`, and the specialized
+experimental `wasi` (Wasm). See
 [providers](features/providers.md) for the full adapter list and which surface
 each one supports.
 
--- a/docs/features/providers.md
+++ b/docs/features/providers.md
@@ -322,10 +322,12 @@ list, and cleanup.
   SSH through `sprite proxy` for a fast Linux microVM on the standard SSH path.
 
 Delegated-run providers (`cloudflare`, `azure-dynamic-sessions`, `e2b`, `islo`,
-`modal`, `tensorlake`, `upstash-box`, `blacksmith-testbox`, `wandb`) do not use
-the broker or an SSH lease; each owns sandbox lifecycle and command execution and
-syncs through its own API (gzipped archive upload for most). See the linked
-provider pages for per-provider auth and configuration.
+`modal`, `tensorlake`, `upstash-box`, `blacksmith-testbox`, `wandb`, and the
+specialized experimental `wasi`) do not use the broker or an SSH lease; each owns
+sandbox lifecycle and command execution and syncs through its own API (gzipped
+archive upload or local copy for wasi). See the linked provider pages for
+per-provider auth and configuration. (WASI is narrow and experimental — Wasm
+modules + limited builtins only.)
 
 ## Static SSH targets
 
--- a/docs/providers/README.md
+++ b/docs/providers/README.md
@@ -97,6 +97,7 @@ the built-in adapter needs a separate local smoke contract.
 | [Tensorlake](tensorlake.md) — `tensorlake` (`tl`, `tensorlake-sbx`) | Linux |
 | [Upstash Box](upstash-box.md) — `upstash-box` (`upstash`, `box`, `upstashbox`) | Linux |
 | [W&B Sandboxes](wandb.md) — `wandb` (`weights-and-biases`) | Linux |
+| [WASI (experimental)](wasi.md) — `wasi` (`wasm`, `wazero`) | Wasm (WASI preview1/2) |
 
 Run `crabbox providers` (`--json`) to see the live capability set the binary
 reports.
--- a/docs/providers/wasi.md
+++ b/docs/providers/wasi.md
@@ -1,0 +1,99 @@
+# WASI Provider (experimental, specialized)
+
+Read when:
+
+- choosing `provider: wasi` (aliases `wasm`, `wazero`);
+- running WebAssembly modules or Wasm-first workloads with capability-based isolation;
+- using `crabbox run` with lightweight `.wasm` artifacts;
+- evaluating generated or untrusted code that can be compiled to WASI.
+
+WASI is an **experimental delegated-run provider**. It executes `.wasm` (wasm32-wasi / preview1 or preview2) modules inside an embedded wazero runtime (or wasmtime if configured and available). The CLI owns local claims + slugs + an isolated guest root; "sync" copies the manifest into the guest tree and preopens it for the module. There is no SSH, no broker, and no full Linux surface.
+
+See the official [WASI site](https://wasi.dev/) and [WASI specification](https://github.com/WebAssembly/WASI) for the API family, proposals, and versioning. This provider targets Preview 1 / WASI 0.1 for broad toolchain compatibility today, with wasmtime available for runtimes that support newer component-model work.
+
+wazero and wasmtime are both listed by wasi.dev. This provider follows the WASI capability model by granting only the synced guest tree as `/work`, plus stdio, args, and explicit environment variables. Convenience builtins such as `ls` and `cat` use the same `/work` boundary.
+
+It is deliberately specialized for:
+- Projects that cross-compile to `wasm32-wasi` (Go, Rust, C, etc.).
+- Reproducible runs where the main artifact is already `.wasm`.
+- Capability-scoped execution for generated or untrusted code.
+
+For arbitrary `pnpm test`, `cargo test`, native toolchains, or full Linux, use `local-container`, `e2b`, `hetzner`, etc.
+
+## When to use
+
+Use when you want a lightweight, zero-OS, capability-scoped sandbox that participates in the normal `crabbox run --id` / warmup / capsule / timing surface.
+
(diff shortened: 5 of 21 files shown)


zozo123:
@clawsweeper re-review

Fixed symlink handling in copy (now preserves as symlinks, no deref of excluded targets; see commit and new regression test). Added live proof (actual module run output) to PR body. Latest CI run on the fix branch is green. Security boundary addressed; real behavior now demonstrated.

zozo123:
**Addressed clawsweeper review P1s for best PR shape (PR #211)**

- **wasmtime env P1**: bare `--env KEY` (names only in argv) + restricted `cmd.Env` (PATH/HOME + exactly allowed vars). Values never on command line (only in child env + wasmtime inherit). Fixes "no secrets as command-line arguments" + GitHub P1.
- **Symlinks P1**: always rewrite to clean relative target via `filepath.Rel` on resolved (after validation). Guest symlinks self-contained in /work (no abs escape). Dangling for excluded preserved (no leak). Best shape per review.
- **Docs**: fixed stale `/bin/false` example -> `false` builtin.
- **Tests/conciseness**: short skip, helper, trimmed e2e to core flows, moved units; security comments added.
- Deep subagent ultrathink review + tests (-race/-short green, vet/fmt clean) done.

@clawsweeper re-review

(Branch `feat/wasi` updated + pushed with "fix(wasi): address ... review P1s + best-shape polish" commit.)

zozo123:
**Addressed fresh clawsweeper review P1 (wasmtime child env) + P3 (aliases) for best PR shape (PR #211)**

- **wasmtime env P1**: `buildWasmtimeRunArgsAndEnv` now builds childEnv from minimal runtime base only (PATH, HOME, USER, LOGNAME, SHELL, TERM, LANG, LC_ALL, TMP*/TEMP) + overlay of explicit allowed `env` map from RunRequest. No full os.Environ() ambient host state leaked to the wasmtime subprocess. Updated godoc. Added regression in `TestWasiWasmtimeBuildArgsNeverPutsValuesInArgv`: `t.Setenv` unrelated host var + assert absent from childEnv (while still requiring PATH and allowed values present, no secret values in argv, correct /work:: mapping).

- **alias P3**: `ApplyWasiProviderFlags` guard now also triggers for aliases (`wasm`, `wazero`) since Apply can run while `cfg.Provider` is still the alias. Rejection error still reports as "wasi". Extended `TestWasiFlagsAndApply` with alias cases.

- All local gates green: `go test -race -short ./internal/providers/wasi ./internal/cli`, `./scripts/test-go-modules.sh`, `go build`, `go vet`, `gofmt`.

- New CI run 27009974225 (for this fix) in progress ~5m: Worker/Scripts/Docs success; Go (Vet done, Test + Test all Go modules in progress); Release Check in progress. Matches our local verification. Prior run (under review) completed success.

@clawsweeper re-review

(Branch `feat/wasi` updated + pushed with "fix(wasi): restrict wasmtime child env ... (P1 ...); reject ... aliases (P3)..." commit. Head now c3179202.)

zozo123:
@clawsweeper re-review

Pushed follow-up fix: WASI sync now rejects symlinked destination parents and symlink destination files before host-side writes when resyncing kept guest workdirs, with regression tests for both cases.

Local verification passed: go test -race ./internal/providers/wasi ./internal/cli && go vet ./... && go build -trimpath -o bin/crabbox ./cmd/crabbox

zozo123:
Proof update after WASI sync hardening:

```text
$ go test -race -v ./internal/providers/wasi -run 'TestWasiCopyManifestRejectsGuestSymlink|TestWasiRunWazeroModuleWithSyncedWorkdir'
=== RUN   TestWasiRunWazeroModuleWithSyncedWorkdir
--- PASS: TestWasiRunWazeroModuleWithSyncedWorkdir (9.40s)
=== RUN   TestWasiCopyManifestRejectsGuestSymlinkDestination
--- PASS: TestWasiCopyManifestRejectsGuestSymlinkDestination (0.00s)
=== RUN   TestWasiCopyManifestRejectsGuestSymlinkParent
--- PASS: TestWasiCopyManifestRejectsGuestSymlinkParent (0.00s)
PASS
ok   github.com/openclaw/crabbox/internal/providers/wasi  11.462s
```

This proves the synced WASI module path still works and that kept guest workdirs reject both symlinked destination files and symlinked parent directories before host-side writes.

zozo123:
@clawsweeper re-review

zozo123:
@clawsweeper re-review

Pushed follow-up fix for the nested symlink-parent case: WASI sync now creates destination parents one component at a time after Lstat checks, so `dir -> outside` plus `dir/sub/file.txt` cannot create host-side directories before rejection.

Verification passed: go test -race ./internal/providers/wasi ./internal/cli && go vet ./... && go build -trimpath -o bin/crabbox ./cmd/crabbox

zozo123:
@clawsweeper re-review

The remaining P1 (delete-sync reset following a symlinked guest workdir parent, `internal/providers/wasi/backend.go:388`) is now fixed in `7c22434`: `os.RemoveAll(target)` is replaced with a guest-root-scoped `removeAllNoFollow` that rejects any symlinked path component before removal, plus regression tests (`TestWasiRemoveAllNoFollowRejectsSymlinkedWorkdirParent`, `TestWasiRemoveAllNoFollowRemovesRealWorkdir`). The branch is rebased on latest main and CI is green (`go test -race ./internal/providers/wasi ./internal/cli`, vet, deadcode all pass).

zozo123:
@clawsweeper re-review

Rebased onto latest `main` (a47e886, "fix: preserve external provider resource identity (#244)") — the PR was in merge-conflict state after the provider-table reformat and dependency changes on main. Conflicts resolved in `README.md`, `docs/providers/README.md`, `go.mod`, `go.sum`, and `internal/cli/providers_builtin_test.go`; the WASI rows now use main's new table format. Also moved `github.com/tetratelabs/wazero` to the direct require block via `go mod tidy` (it is imported directly by `internal/providers/wasi`).

Verification on the rebased branch:
- `go build -trimpath -o bin/crabbox ./cmd/crabbox`, `go vet ./...`, `gofmt` clean
- `go test -race ./...` — all packages pass (one unrelated flake in `internal/providers/tart` under full-suite load; it passes standalone on both this branch and upstream main)
- `./scripts/check-docs.sh` — command surface, internal links, and docs site build all ok
- Binary smokes: `providers --json` lists `wasi` with `wasm`/`wazero` aliases; `doctor --provider wasi --json` ok; `run --provider wasi -- echo …` reports `sync_skipped=true`

zozo123:
@clawsweeper re-review

Addressed the P2 (guest-base sync pollution) in 0074638, rebased on latest `main` (616875a):

- `guestRootPath` now requires an absolute `wasi.guestBaseDir` (default `os.TempDir()` behavior unchanged), and `prepareGuest` rejects guest roots that overlap the synced repo tree in either direction — comparing symlink-resolved paths (longest existing prefix, so `/var` vs `/private/var` on macOS can't bypass it) — **before any host-side write**, so kept guest files can never enter later sync manifests.
- Regression tests: `TestWasiPrepareGuestRejectsRelativeGuestBase`, `TestWasiPrepareGuestRejectsGuestRootInsideRepo` (also asserts nothing was created in-repo), `TestWasiPrepareGuestRejectsRepoInsideGuestRoot`, `TestWasiGuestRootPathDefaultsToTempDir`.
- Documented the constraint in `docs/providers/wasi.md`. CHANGELOG.md untouched.

Live proof:

```text
$ CRABBOX_WASI_GUEST_BASE_DIR="guest-roots" bin/crabbox run --provider wasi -- ls .
wasi guest base dir must be an absolute path: "guest-roots"

$ CRABBOX_WASI_GUEST_BASE_DIR="$PWD/guest-roots" bin/crabbox run --provider wasi -- ls .
wasi guest root .../crabbox/guest-roots/crabbox-wasi-cbx_059aebabfb82 overlaps repo root .../crabbox; set wasi.guestBaseDir outside the repository

$ ls guest-roots
ls: guest-roots: No such file or directory   # rejected before any host-side write

$ bin/crabbox run --provider wasi -- echo still-works
still-works
wasi run summary sync_skipped=true command=0s total=1ms exit=0
```

On the failed Go check: the Deadcode gate failure (`internal/providers/railway/core.go:51: unreachable func: writeTimingJSON`) is pre-existing on current `main` (all three latest main CI runs fail the same way) and is fixed separately in https://github.com/openclaw/crabbox/pull/255. Once that lands, a rerun here should be green.

Verification on this branch: `go build`, `go vet`, `gofmt` clean, `go test -race ./internal/providers/wasi ./internal/cli` pass, `./scripts/check-docs.sh` ok.

zozo123:
@clawsweeper re-review

All required checks are green on the current head (aa51178): the branch is rebased on latest `main`, which resolves the previously-blocking Railway Deadcode failure (https://github.com/openclaw/crabbox/pull/255 closed as superseded), and the rebase-exposed tart deadcode orphan was removed in aa51178. The guest-base sync containment fix and its regression tests from the previous round are unchanged.

This satisfies the remaining contributor-actionable rank-up move ("rerun required checks after the unrelated main Deadcode failure is fixed"). The two remaining items — Socket triage for `wazero v1.12.0` and explicit acceptance of the WASI sandbox boundary — are maintainer decisions.

steipete:
Thanks for the substantial implementation and hardening work here. I don't currently see WASI as a fit for Crabbox's remote software testing and execution focus: there is no linked user request or concrete Crabbox workload that justifies adding and maintaining this provider and its runtime dependency, so I'm closing this PR.

I'm open to further discussion if there is concrete demand or an end-to-end Crabbox use case we're missing.

PR #212, opened 2026-06-05
feat: add Hyper-V provider for local Windows VMs

anagnorisis2peripeteia:
## Summary

A `hyperv` provider for running Crabbox leases on local Windows VMs via Microsoft Hyper-V and PowerShell Direct — the Windows-native counterpart to `tart` (macOS), with no external CLI or cloud account. The design goal is "bring a plain Windows VHDX with a known administrator password, and it just works": the provider supplies everything else. With `--hyperv-init-password` even that requirement drops away for auto-logon images: a **stock, password-less** Microsoft dev-environment VHDX works completely unmodified.

## How it works

`crabbox warmup --provider hyperv --hyperv-image <vhdx>`:

1. **Thin clone, not a copy.** Each lease is backed by a *differencing disk* over the template (`New-VHD -Differencing -ParentPath`) — near-instant and space-thin (a few MB up front), so the template stays read-only and shared instead of copying tens of GB per lease.
2. **(Opt-in) first-boot password for password-less templates.** Microsoft's downloadable dev-environment VHDXs auto-log-on with **no password**, and PowerShell Direct refuses empty credentials — so stock images used to be unusable. With `--hyperv-init-password` the provider mounts the *lease's own differencing disk* offline, loads its registry hive, and writes a `RunOnce` command that sets the guest password to `CRABBOX_HYPERV_GUEST_PASSWORD` at the template's auto-logon. The template VHDX is never modified. Guard rails: requires an explicit `CRABBOX_HYPERV_GUEST_PASSWORD` (refuses to stamp the default password onto a guest), rejects `"`/`%` in the password and the user name (both pass through cmd.exe), and the password reaches the host-side script via env, never argv.
3. **Create + place off the system drive.** `New-VM -Generation 2 -Path <vmDir>` puts the VM config and the multi-GB `.VMRS` runtime files on the same (data) drive as the VHD rather than defaulting to `C:\ProgramData`, and automatic checkpoints are disabled (`Set-VM -AutomaticCheckpointsEnabled $false`) so a lease never spawns a `.avhdx` that would strand a disk on release.
4. **Make the guest reachable.** Over PowerShell Direct (authenticating with the guest admin password, passed via env — never argv) the provider installs the OpenSSH server if absent (`Add-WindowsCapability`, `Start-Service sshd`, firewall) and installs git if absent (portable MinGit, pinned to an immutable release asset and SHA-256-verified before extraction) — mirroring how the Linux cloud-init path provisions a guest. Both are no-ops when already present, so a template that pre-bakes them just skips the download. MinGit is extracted to `C:\Program Files\MinGit`, **not** `C:\Program Files\Git`: MinGit's `etc\gitconfig` deliberately includes `C:/Program Files/Git/etc/gitconfig`, so extracting it there makes the include self-referential and every guest git command dies with "exceeded maximum include depth" (caught live by the second proof below; fixed in this PR).
5. **Inject the lease key.** The SSH public key is written to the user's `authorized_keys` and, for admin accounts, `administrators_authorized_keys` with the SYSTEM+Administrators-only ACL that Windows OpenSSH requires (otherwise sshd silently ignores it).
6. **Wait for readiness**, then run/sync/release normally. On release the VM is removed and only the lease's own differencing disk + per-VM config dir are deleted; the template is untouched. Cleanup is scoped to the `crabbox-` name prefix.

Net template requirement: **a Generation-2 Windows VHDX** — either with an administrator password known to Crabbox (`--hyperv-user` / `CRABBOX_HYPERV_GUEST_PASSWORD`), or, for auto-logon images, none at all with `--hyperv-init-password`. Non-Windows targets are rejected (from CLI, YAML, or env), and ISO images are rejected.

## Live proof 1 — passworded template (existing path), Windows 11 Pro host, Go 1.26.4

Captured on the latest head. Plain Windows 11 VHDX with only an admin password set; C: deliberately full so storage is redirected to D: (which also exercises the off-system-drive placement):

```
$ crabbox doctor --provider hyperv
ok  provider  provider=hyperv hyperv=Enabled inventory=ready api=powershell leases=0 image=… ssh_probe=unchecked

$ crabbox warmup --provider hyperv --hyperv-image D:\hyperv-templates\win11-ssh-ready.vhdx --hyperv-cpu 2 --hyperv-memory 3072
provisioned lease=cbx_e70a9aac5441 instance=crabbox-hvproofa-240c4232 state=ready
ready ssh=Administrator@172.22.155.0:22 network=public workroot=C:\crabbox

$ crabbox run --provider hyperv --id cbx_e70a9aac5441 -- cmd /c ver
sync candidate: 1 files, 27 B
sync complete in 12.056s
Microsoft Windows [Version 10.0.22621.3880]
run summary sync=12.056s command=6.513s total=21.454s sync_skipped=false exit=0

$ crabbox list --provider hyperv
crabbox-hvproofa-240c4232 ready hyperv 172.22.155.0 lease=cbx_e70a9aac5441 slug=hvproofa keep=true target=windows

$ crabbox stop --provider hyperv --id cbx_e70a9aac5441
released lease=cbx_e70a9aac5441 instance=crabbox-hvproofa-240c4232
# verify: crabbox- VMs remaining: 0   (VM, differencing disk, and config dir all removed)
```

## Live proof 2 — broader template coverage: STOCK password-less Microsoft image + `--hyperv-init-password`

The template here is Microsoft's downloadable `WinDev2407Eval.vhdx` **exactly as downloaded** (file mtime 2024-07-25 before and after the run — never written): auto-logon `User`, no password, no OpenSSH, no git. One flag makes it a working lease:

```
$ crabbox warmup --provider hyperv --hyperv-image C:\Users\…\WinDev2407Eval.vhdx ^
    --hyperv-user User --hyperv-init-password --hyperv-cpu 2 --hyperv-memory 3072
provisioned lease=cbx_48ec029c0094 instance=crabbox-hvproofb-5412dc36 state=ready
ready ssh=User@172.22.145.28:22 network=public workroot=C:\crabbox
warmup complete total=9m31.274s

$ crabbox run --provider hyperv --id cbx_48ec029c0094 -- git --version
sync complete in 14.581s
git version 2.54.0.windows.1
run summary sync=14.581s command=9.902s total=27.016s sync_skipped=false exit=0

$ crabbox stop --provider hyperv --id cbx_48ec029c0094
released lease=cbx_48ec029c0094 instance=crabbox-hvproofb-5412dc36
# verify: crabbox- VMs remaining: 0; template mtime unchanged (2024-07-25)
```

Reaching `state=ready` here exercises the full chain on a guest that started with **nothing** — first-boot password injection, OpenSSH install, pinned + checksum-verified MinGit install (to the corrected path; `git --version` exiting 0 in-guest is the regression proof for the MinGit fix), ACL-correct key injection, and the Windows readiness check over SSH.

## Configuration

CLI: `--hyperv-image` (required), `--hyperv-cpu`, `--hyperv-memory`, `--hyperv-switch`, `--hyperv-user`, `--hyperv-work-root`, `--hyperv-init-password`. YAML `hyperv:` and `CRABBOX_HYPERV_*` env equivalents, incl. `CRABBOX_HYPERV_GUEST_PASSWORD` and `CRABBOX_HYPERV_INIT_PASSWORD`.

## Not included (follow-up)

- A configurable storage root (today storage follows `%USERPROFILE%`).
- ISO install / autounattend (templates are pre-installed VHDX).
- WSL2 mode (native Windows only).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Added
+
+- Added `provider: hyperv` for local Windows VM SSH leases through Microsoft Hyper-V, including differencing-disk provisioning, OpenSSH and MinGit bootstrap, password-less dev-image initialization, retained lease reuse, and cleanup. Thanks @anagnorisis2peripeteia.
+
 ### Fixed
 
 - Fixed Namespace Devbox setup instructions to use the current browser workspace approval flow instead of obsolete token environment variables.
--- a/docs/providers/README.md
+++ b/docs/providers/README.md
@@ -33,7 +33,8 @@ SSH-lease providers further differ by how they reach the cloud:
   a Docker-compatible local runtime (Docker Desktop, OrbStack, Colima),
   `apple-container` uses Apple's native `container` runtime on Apple silicon
   macOS, `multipass` launches local Ubuntu VMs through Canonical Multipass,
-  and `tart` runs macOS VMs on Apple Silicon via Cirrus Labs tart.
+  `tart` runs macOS VMs on Apple Silicon via Cirrus Labs tart, and `hyperv`
+  creates local Windows VMs through Microsoft Hyper-V.
 - **Delegated sandbox** — managed sandbox/proof runners that execute remotely
   without an SSH lease (e.g. `e2b`, `modal`, `islo`, `cloudflare`,
   `azure-dynamic-sessions`, `docker-sandbox`).
@@ -67,6 +68,7 @@ the built-in adapter needs a separate local smoke contract.
 | [Apple Container](apple-container.md) — `apple-container` (`apple`, `applecontainer`) | Linux · local |
 | [Multipass](multipass.md) — `multipass` (`mp`, `canonical-multipass`) | Linux · local |
 | [Tart](tart.md) — `tart` (`local-tart`, `macos-vm`) | macOS · local |
+| [Hyper-V](hyperv.md) — `hyperv` | Windows · local |
 | [exe.dev](exe-dev.md) — `exe-dev` (`exe`, `exedev`) | Linux · direct |
 | [KubeVirt](kubevirt.md) — `kubevirt` (`kubernetes-vm`) | Linux · direct |
 | [External](external.md) — `external` (`exec-provider`) | Linux · direct |
--- a/docs/providers/hyperv.md
+++ b/docs/providers/hyperv.md
@@ -49,6 +49,37 @@ net user Administrator /active:yes
 Then point `--hyperv-image` at the VHDX and set `--hyperv-user Administrator`
 and `CRABBOX_HYPERV_GUEST_PASSWORD=<password>`. The provider handles OpenSSH.
 
+### Password-less templates (Windows dev-environment images)
+
+Microsoft's downloadable Windows dev-environment VHDXs auto-log-on as `User`
+with **no password**, and PowerShell Direct refuses empty credentials — so a
+stock image fails the bootstrap as-is. Pass `--hyperv-init-password` to use one
+unmodified:
+
+```sh
+set CRABBOX_HYPERV_GUEST_PASSWORD=<password>
+crabbox warmup --provider hyperv --hyperv-image C:\Images\WinDev2407Eval.vhdx ^
+  --hyperv-user User --hyperv-init-password
+```
+
+Before first boot the provider mounts the per-lease differencing disk, loads
+its offline registry hive, and writes a `RunOnce` command that sets the guest
+account's password to `CRABBOX_HYPERV_GUEST_PASSWORD` at the template's
+auto-logon. Only the lease disk is modified; the template VHDX stays untouched.
+
+Notes:
+
+- Requires an explicit `CRABBOX_HYPERV_GUEST_PASSWORD` (the provider refuses
+  to stamp its default password onto a guest).
+- The password cannot contain `"` or `%` (it passes through `cmd.exe` at
+  logon).
+- This only works for templates that auto-log-on an administrator account
+  (`RunOnce` fires at logon). Templates without auto-logon need a known
+  password baked in, as above.
+- The password is briefly visible inside the guest (the `RunOnce` registry
+  value, then the `net.exe` command line at first logon). The guest belongs to
+  the lease, and the differencing disk holding it is deleted on release.
+
 ## Configuration
 
 ### Flags
@@ -62,6 +93,7 @@ and `CRABBOX_HYPERV_GUEST_PASSWORD=<password>`. The provider handles OpenSSH.
 | `--hyperv-memory` | `8192` | Memory in MB |
 | `--hyperv-disk` | `50` | Reserved; leases inherit the template's virtual size (differencing disk) |
 | `--hyperv-switch` | `Default Switch` | Hyper-V virtual switch name |
+| `--hyperv-init-password` | `false` | Set the guest password at first boot via the lease disk (password-less auto-logon templates) |
 
 ### Config file
 
@@ -75,6 +107,7 @@ hyperv:
   disk: 50
   switch: Default Switch
   guestPassword: crabbox
+  initPassword: false
 ```
 
 ### Environment variables
@@ -89,6 +122,7 @@ hyperv:
 | `CRABBOX_HYPERV_DISK` | Disk in GB |
 | `CRABBOX_HYPERV_SWITCH` | Virtual switch name |
 | `CRABBOX_HYPERV_GUEST_PASSWORD` | Guest user password for SSH key injection |
+| `CRABBOX_HYPERV_INIT_PASSWORD` | Set the guest password at first boot (`true`/`false`) |
 
 ## Bootstrap contract
 
@@ -96,14 +130,16 @@ During `Acquire`, the provider:
 
 1. Creates a per-lease **differencing disk** backed by the template (near-instant
    and space-thin; the template stays read-only and shared — no multi-GB copy)
-2. Creates and starts the VM
-3. Installs and starts the Windows OpenSSH server in the guest via PowerShell
+2. With `--hyperv-init-password`, mounts the lease disk offline and writes a
+   first-boot `RunOnce` that sets the guest password (password-less templates)
+3. Creates and starts the VM
+4. Installs and starts the Windows OpenSSH server in the guest via PowerShell
    Direct if not already present (`Add-WindowsCapability`, `Start-Service sshd`,
    firewall rule), and installs git (MinGit) if absent — both required for SSH
    readiness and crabbox sync
-4. Injects the per-lease SSH public key via PowerShell Direct
+5. Injects the per-lease SSH public key via PowerShell Direct
    (`Invoke-Command -VMName`) using the configured guest password
-5. Waits for SSH readiness on the injected key
+6. Waits for SSH readiness on the injected key
 
 Both the OpenSSH-install and key-injection steps authenticate over PowerShell
 Direct using the guest administrator password and retry up to 5 times with
--- a/docs/providers/hyperv.md
+++ b/docs/providers/hyperv.md
@@ -1,7 +1,6 @@
 # Hyper-V
 
 Provider id: `hyperv`
-Aliases: `local-hyperv`, `hyper-v`, `windows-vm`
 Kind: SSH lease
 Targets: Windows (native)
 Family: `local-vm`
--- a/docs/providers/hyperv.md
+++ b/docs/providers/hyperv.md
@@ -21,14 +21,31 @@ reject configuration on non-Windows hosts.
 
 - Windows 10 Pro/Enterprise/Education or Windows Server with Hyper-V enabled
 - PowerShell 5.1 or later (ships with Windows)
-- A pre-configured Windows VHDX template with:
-  - OpenSSH Server installed and running
-  - A known user account (default: `crabbox`) with a known password
+- A Windows VHDX template (Generation 2 / UEFI) with:
+  - A local administrator account whose password is known to Crabbox
+    (`--hyperv-user` / `CRABBOX_HYPERV_GUEST_PASSWORD`)
   - Network configured for DHCP on the Hyper-V virtual switch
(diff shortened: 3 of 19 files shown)


anagnorisis2peripeteia:
@clawsweeper re-review

anagnorisis2peripeteia:
@clawsweeper re-review

anagnorisis2peripeteia:
@clawsweeper re-review

anagnorisis2peripeteia:
@clawsweeper re-review

anagnorisis2peripeteia:
@clawsweeper re-review

The provider now makes a plain Windows VHDX (just a known admin password) work end-to-end — it supplies the rest:
- **Differencing-disk leases** (thin clone, no multi-GB copy per lease)
- **Off-system-drive placement** (New-VM -Path for VM config + .VMRS; auto-checkpoints disabled to avoid disk leaks)
- **Guest provisioning over PowerShell Direct**: installs OpenSSH + git if absent (mirrors Linux cloud-init), key injected to administrators_authorized_keys with the SYSTEM+Administrators-only ACL Windows OpenSSH requires
- **Explicit non-Windows target rejection** (CLI/YAML/env)

Live proof in the body: doctor → warmup→state=ready (ssh=Administrator@…, workroot=C:\crabbox) → run (sync+command over SSH) → list → stop → cleanup (0 VMs), on a plain dev VHDX with C: full (storage redirected to D:).

anagnorisis2peripeteia:
@clawsweeper re-review

Addressed the P2: registered the documented `--hyperv-user` and `--hyperv-work-root` flags (guest password stays off the CLI — only `CRABBOX_HYPERV_GUEST_PASSWORD`). Added a test; verified `--hyperv-user` parses. The documented quickstart now works as written.

anagnorisis2peripeteia:
@clawsweeper re-review

Addressed the docs P3: the flags table now lists `--hyperv-user` and `--hyperv-work-root` alongside the others, matching flags.go.

anagnorisis2peripeteia:
@clawsweeper re-review

Took the "broader Hyper-V proof" option, and made it count: the second proof now runs on a **different template class** — Microsoft's stock `WinDev2407Eval.vhdx` exactly as downloaded (auto-logon, no password, no OpenSSH, no git).

What changed since the last review:

- **New: `--hyperv-init-password`** (`hyperv.initPassword` / `CRABBOX_HYPERV_INIT_PASSWORD`). Stock dev-environment images have no password and PowerShell Direct refuses empty credentials, so they previously couldn't bootstrap at all. The provider now (opt-in) mounts the lease's own differencing disk offline before first boot and writes a `RunOnce` that sets the guest password at the template's auto-logon. Template never modified (mtime-verified in the proof). Guard rails: explicit `CRABBOX_HYPERV_GUEST_PASSWORD` required, `"`/`%` rejected (cmd quoting), password via env never argv. Unit tests cover the injection ordering (before `New-VM`), the no-flag default (no offline mount), and both validation rejections.
- **Fix (caught by the broader proof): MinGit extraction path.** `ensureGit` extracted MinGit to `C:\Program Files\Git`, but MinGit's `etc\gitconfig` deliberately includes `C:/Program Files/Git/etc/gitconfig` — self-referential at that destination, so every guest git command failed with "exceeded maximum include depth" (visible as `exit=128` in the previous proof's run step). Now extracts to `C:\Program Files\MinGit`, where the include points at a missing file and is ignored. Regression-locked in the unit test; `git --version` → `git version 2.54.0.windows.1`, `exit=0` on a fresh stock guest in the new proof.

Both proofs in the PR body were recaptured on this head: proof 1 = passworded template lifecycle (doctor → warmup → ready → run exit=0 → list → stop → 0 VMs left), proof 2 = stock password-less template with `--hyperv-init-password`, same lifecycle including in-guest `git --version` exit=0, 0 VMs left, template byte-untouched.

anagnorisis2peripeteia:
@clawsweeper re-review

anagnorisis2peripeteia:
@clawsweeper re-review

anagnorisis2peripeteia:
@clawsweeper re-review

Both rank-up moves done, plus the P2:

- **Rebased onto current main** (`c8d469e`). The two real conflicts (provider registry import list and the providers README table, both against the new Incus provider) were resolved keeping both providers; no newer config behavior was dropped — Incus, external-provider, Tart, and local-container settings are untouched and `./internal/providers/all` passes with both registered.
- **[P2] `--hyperv-disk` removed entirely** (flag, YAML `disk`, `CRABBOX_HYPERV_DISK`, struct fields, defaults, test double, docs rows). Leases inherit the template's virtual size by design of the differencing disk — the docs now say to size the template. No sizing promise remains that the code doesn't keep.
- **Both live proofs recaptured on the exact resolved head** (PR body updated):
  - Proof 1 (passworded template): warmup → `ready ssh=Administrator@…` → run `cmd /c ver` → `exit=0` → stop → 0 VMs remaining.
  - Proof 2 (stock password-less `WinDev2407Eval.vhdx` + `--hyperv-init-password`): warmup → ready in 8m22s → `git --version` → `git version 2.54.0.windows.1`, `exit=0` → stop → 0 VMs remaining; template mtime unchanged (2024-07-25).

anagnorisis2peripeteia:
@clawsweeper re-review

Both rank-up moves done:

- **[P1] MinGit pinned + SHA-256-verified.** `ensureGit` no longer resolves `releases/latest` at all: it downloads the immutable asset `v2.54.0.windows.1/MinGit-2.54.0-64-bit.zip` and compares `Get-FileHash` against the official release-notes checksum **before** `Expand-Archive`; on mismatch it deletes the archive and throws (fail closed, surfaced as a provisioning error). The pinned checksum was independently confirmed by downloading the asset and hashing it. Three focused tests: immutable-pin assertions (fails if any `releases/latest`/`Invoke-RestMethod` reference returns), verification-ordering (hash + compare must precede extraction), and the fail-closed mismatch branch (archive removed + throw).
- **Both live proofs recaptured on the repaired head** (PR body updated). Proof 2 is the live validation of this exact fix: a stock password-less guest with no git ran the pinned download + checksum verification end-to-end and `git --version` → `git version 2.54.0.windows.1`, `exit=0`; both proofs released cleanly with 0 VMs remaining and the template untouched.

anagnorisis2peripeteia:
@clawsweeper re-review

Rank-up move done: the init-password **user name** now gets the same cmd.exe metacharacter validation as the password — `Acquire` rejects `"` and `%` in `--hyperv-user` when `--hyperv-init-password` is set, before any host or guest mutation, so an embedded quote can no longer alter the elevated first-boot `RunOnce` command. Regression tests cover injection-shaped user names (`us"er`, `u" & calc & "`, `%PATH%`) and extend the password cases (`%TEMP%`, `a"&calc&"b`); docs updated to state the constraint applies to both values.

Both live proofs were recaptured on this exact head (PR body updated): passworded template `cmd /c ver` → `exit=0`; stock password-less template + `--hyperv-init-password` → ready, `git --version` → `git version 2.54.0.windows.1`, `exit=0`; both released with 0 VMs remaining and the template untouched.

anagnorisis2peripeteia:
@clawsweeper re-review

Fixed the failing Go CI job: the coverage gate (core `internal/cli` files >= 90%) dipped to 89.7% because the new `hyperv` config wiring in `config.go` was uncovered. Added `config_hyperv_test.go` covering the full HyperV YAML and env wiring — image/user/workRoot/cpus/memory/switch/guestPassword, `initPassword` from both sources, the `*bool` tri-state (absent key preserves, explicit `false` wins), and env-false-overrides-config-true. Core coverage measured at 90.6% with the same filter the CI script uses.

Test-only change — no provider code touched, so the two live lifecycle proofs in the PR body (captured on the provider code this head ships) stand.

anagnorisis2peripeteia:
@clawsweeper re-review

anagnorisis2peripeteia:
@clawsweeper re-review

Rank-up move done: lease persistence is now **atomic**. `Acquire` writes the claim and its SSH endpoint in a single `ClaimLeaseForRepoProviderScopePondEndpoint` call (the same helper the Tart provider uses) instead of claim-then-endpoint — a persistence failure can no longer strand a claim for a VM the error path then removes. The separate endpoint-update path is gone from the provider entirely.

Focused regression tests: the success case asserts the written claim already carries the SSH endpoint and instance (one write, no second mutation), and the failure case forces the claim write to fail (state dir pointed at a plain file) and asserts **no** lease state survives.

Both live proofs recaptured on this exact head (PR body updated): passworded template `cmd /c ver` `exit=0`; stock password-less template + `--hyperv-init-password` → ready in 9m31s, `git --version` `exit=0`; both released, 0 VMs remaining, template untouched.

anagnorisis2peripeteia:
@clawsweeper re-review

anagnorisis2peripeteia:
@clawsweeper re-review

anagnorisis2peripeteia:
@clawsweeper re-review

anagnorisis2peripeteia:
@clawsweeper re-review

anagnorisis2peripeteia:
@clawsweeper re-review

anagnorisis2peripeteia:
@clawsweeper re-review

steipete:
Maintainer live validation on June 11, 2026:

- Provisioned a fresh nested Hyper-V Windows Server 2022 lease on an AWS Windows host: `cbx_458cefeed3ee` / `crabbox-hv-live-91dbde19`.
- Verified generated OpenSSH host-key ACLs allow only SYSTEM and Administrators, then completed SSH authentication with the per-lease key.
- Ran `crabbox run --provider hyperv --id cbx_458cefeed3ee --no-sync -- whoami`; output was `ec2amaz-t0tblif\administrator`, exit 0.
- Ran `crabbox stop`; verified the VM, VHDX, VM config, lease claim, and stored key were all removed.
- Reproduced and fixed Windows inbox OpenSSH hanging with Go-managed output pipes. Verified the final file-backed stream implementation on a fresh AWS Windows host, including ordered combined output, nil writers, local writer failures, 20 MiB forced spool rotation, immediate delete-pending temp files, and a real Crabbox SSH command exiting 0.

Local verification at `4c4c0d9159a70d7dea41215a2ea5c5ff4722b965`:

- `go test -race ./...`
- `go vet ./...`
- Windows cross-build and Windows-only test binary
- Autoreview: no accepted/actionable findings

PR #350, opened 2026-06-14
fix(webvnc): force VNC password authentication

steipete-oai:
## What changed

- add `webvnc local --security-type vnc`
- filter the initial RFB security offer to standard VNC password authentication
- keep all later RFB traffic end-to-end between noVNC and the loopback VNC source
- bound negotiation on both browser and server connections and stop cleanly on cancellation

## Why

Some VNC servers advertise account-based authentication before standard VNC password authentication. noVNC chooses the first supported method, so a valid independent VNC password can otherwise be sent through the wrong authentication path.

## Validation

- `go vet ./internal/cli`
- focused race tests for the local WebVNC bridge
- structured autoreview: clean

### Live RFB proof

A disposable TigerVNC 1.13.1 server was configured with `RA2ne,VncAuth`, so it advertised username/password account authentication before standard VNC password authentication. The built CLI then served `webvnc local --security-type vnc`; an RFB client connected through its authenticated WebSocket and completed initialization:

```text
upstream security offer: [6 2]
bridge security offer: [2]
VNC password authentication: success
framebuffer initialized: 640x480
```

The proof used only a disposable local server and dummy credentials.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 
 ### Fixed
 
+- Added an explicit `webvnc local --security-type vnc` mode that forces standard VNC password authentication when a server advertises account authentication first.
 - Fixed coordinator hibernation recovery to preserve unambiguous live bridges while rejecting duplicate or stale restored endpoints.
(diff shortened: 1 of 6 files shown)


steipete-oai:
@clawsweeper re-review

PR #391, opened 2026-06-15
fix: validate bridge ticket bindings before consume

### LINKED ISSUE #371: [security] Bridge tickets are consumed before endpoint binding validation
## Summary

Crabbox bridge agent endpoints consume single-use WebVNC, code, egress, and runtime-adapter tickets before verifying that the URL endpoint matches the lease or adapter recorded in the ticket. An attacker who obtains a valid bridge ticket can send it to a wrong lease or adapter agent URL, causing the coordinator to delete the ticket and then reject the upgrade. The legitimate bridge then cannot use the same ticket and must mint a new one.

The ticket values are 128-bit random capabilities, so this is not a guessing attack. The reportable weakness is that the ticket-consuming code performs the irreversible storage delete before the endpoint binding checks that decide whether the ticket is usable for the requested lease or adapter.

Scope status: `in_scope`. This fits Crabbox's bridge-ticket boundary because a valid bridge capability can be destroyed through a mismatched coordinator endpoint before the endpoint proves it is bound to the ticket's lease or adapter.

## Affected Components

- Checked commit: `d3d1891aafa7213e7b76380465a75da8d20fb64d`
- Component: Cloudflare Worker coordinator bridge ticket handling
- Affected files:
  - `worker/src/fleet.ts`
  - `internal/cli/webvnc.go`
  - `internal/cli/code.go`
  - `internal/cli/egress.go`

## Attack Path

Attacker role: anyone who can observe, steal, or receive a still-valid bridge ticket.

Prerequisites:

- The attacker has a valid `wvnc_...`, `code_...`, `egress_...`, or `adapter_...` ticket.
- The ticket has not expired. The affected ticket TTLs are short-lived; WebVNC, code, egress, and runtime-adapter tickets are created with 120 second lifetimes.
- The attacker can send a WebSocket upgrade request to the coordinator agent endpoint.

Steps for a lease-bound ticket:

1. A legitimate operator mints a WebVNC, code, or egress bridge ticket for lease `A`.
2. The attacker obtains the ticket before it is used.
3. The attacker sends a WebSocket upgrade to a different endpoint such as `/v1/leases/B/webvnc/agent`, `/v1/leases/B/code/agent`, or `/v1/leases/B/egress/host` with the stolen ticket.
4. The coordinator looks up the ticket and deletes it from storage.
5. The coordinator then loads the lease recorded in the ticket and rejects the request because the URL identifier does not match that lease.
6. The legitimate operator's subsequent upgrade to lease `A` fails because the ticket has already been consumed.

The same pattern applies to runtime-adapter tickets: the agent endpoint consumes the ticket before checking whether `ticket.adapterID` equals the adapter ID in the URL.

## Impact

A known valid bridge ticket can be burned by presenting it to a mismatched endpoint before the legitimate bridge uses it. The direct impact is loss of availability for that single bridge connection window:

- WebVNC backend agent setup can be forced to fail until a new ticket is minted.
- Code bridge setup can be forced to fail until a new ticket is minted.
- Egress host/client bridge setup can be forced to fail until a new ticket is minted.
- Runtime-adapter relay setup can be forced to fail until a new ticket is minted.

The attacker must already know a valid ticket. That prerequisite keeps the issue bounded, but the codebase has legitimate ticket exposure surfaces where an accidental leak is realistic: WebVNC and code bridges use a bearer header first and retry with `?ticket=...` when the upstream returns 401 or 403, while egress bridge URLs also carry `?ticket=...`.

### CVSS Assessment

| Metric | v3.1 | v4.0 |
|--------|------|------|
| **Score** | 3.7 / 10.0 | 6.3 / 10.0 |
| **Severity** | Low | Medium |
| **Vector** | CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:N/A:L | CVSS:4.0/AV:N/AC:H/AT:P/PR:N/UI:N/VC:N/VI:N/VA:L/SC:N/SI:N/SA:N |
| **Calculator** | [CVSS v3.1 Calculator](https://www.first.org/cvss/calculator/3.1#CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:N/A:L) | [CVSS v4.0 Calculator](https://www.first.org/cvss/calculator/4.0#CVSS:4.0/AV:N/AC:H/AT:P/PR:N/UI:N/VC:N/VI:N/VA:L/SC:N/SI:N/SA:N) |

## Recommended Remediation

Do not delete a ticket until every endpoint binding check needed for that ticket has passed.

For lease-bound tickets, move the lease lookup and `identifierMatchesLease` check before the storage delete. For egress, also ensure the role check does not consume a ticket for the wrong role. For runtime-adapter tickets, compare `ticket.adapterID` with the URL adapter ID before deleting.

One possible implementation shape is:

```ts
const ticket = await storage.get<TicketRecord>(key);
if (!ticket || expired(ticket)) return undefined;
if (!bindingMatches(ticket, requestIdentifier)) return undefined;
await storage.delete(key);
return ticket;
```

Preserve the current behavior that expired tickets are unusable, but avoid consuming unexpired tickets on mismatched endpoints.

## Validation

Validation method: source review of the isolated checkout.

### Ticket minting is scoped, but consumption is not bound before delete

`worker/src/fleet.ts` mints WebVNC and code tickets only after resolving a lease and requiring manage access:

- `createWebVNCTicket` resolves the lease, checks `leaseManageableByRequest`, stores a ticket containing `leaseID`, and returns it to the caller.
- `createCodeTicket` follows the same pattern.
- `createEgressTicket` also resolves the lease, checks manage access, stores `leaseID`, `role`, and `sessionID`, then returns the ticket.
- `createRuntimeAdapterTicket` stores `adapterID`, `owner`, and `org` in the ticket record.

Those creation checks are appropriate, but the consume helpers do not verify the endpoint binding before deleting:

- `consumeWebVNCTicket` reads the ticket record, executes `await this.state.storage.delete(key)`, then checks expiration and returns the ticket.
- `consumeCodeTicket` does the same.
- `consumeEgressTicket` does the same.
- `consumeRuntimeAdapterTicket` does the same.

### The endpoint binding checks happen after consumption

The agent handlers call the consume helper before validating the requested lease or adapter:

- `webVNCAgent` calls `consumeWebVNCTicket(request)`, then loads `ticket.leaseID`, then checks `identifierMatchesLease(identifier, lease)`.
- `codeAgent` calls `consumeCodeTicket(request)`, then loads `ticket.leaseID`, then checks `identifierMatchesLease(identifier, lease)`.
- `egressAgent` calls `consumeEgressTicket(request)`, then checks the role, then loads `ticket.leaseID`, then checks `identifierMatchesLease(identifier, lease)`.
- `runtimeAdapterAgent` calls `consumeRuntimeAdapterTicket(request)`, then checks `ticket.adapterID !== adapterID`.

In each case, a mismatched URL returns an error after the storage delete has already happened. There is no rollback path.

### Bridge agent upgrades intentionally rely on the ticket

`worker/src/coordinator-entry.ts` explicitly routes WebVNC, code, egress, and runtime-adapter WebSocket agent upgrades to the fleet without normal user authentication. The ticket is therefore the security control for those agent upgrade paths.

### Tickets can appear in URLs in normal client flows

The default code and WebVNC flows try to send tickets in the `Authorization: Bearer ...` header first, but they retry with a `ticket` query parameter when the websocket response is 401 or 403:

- `internal/cli/webvnc.go` retries with `webVNCAgentURLWithTicketAndCapabilities`, which sets `ticket` in the URL query.
- `internal/cli/code.go` retries with `webCodeAgentURLWithTicket`, which sets `ticket` in the URL query.

The egress bridge path always builds its agent URL with a `ticket` query parameter in `internal/cli/egress.go`. These URL surfaces do not create the bug by themselves, but they make accidental ticket disclosure through logs, screenshots, shell history, or proxy telemetry a realistic prerequisite.

Add focused tests in `worker/test/fleet.test.ts`:

- Mint a WebVNC ticket for lease `A`, attempt a websocket upgrade on lease `B` with that ticket, assert the request is rejected, then assert the same ticket can still be used on lease `A`.
- Repeat the same binding-preservation test for code and egress tickets.
- Mint a runtime-adapter ticket for adapter `A`, attempt an upgrade on adapter `B`, assert rejection, then assert the ticket remains usable for adapter `A`.
- Keep a separate test proving that a successful correct-endpoint upgrade still consumes the ticket exactly once.

Suggested verification:

```bash
npm test --prefix worker
```

## Disclosure Recommendation

- Recommended route: public GitHub issue.
- Reason: This is a bounded denial-of-service issue against short-lived bridge setup tokens. It does not disclose a privilege-escalation path, data exfiltration path, or a way to guess tickets.
- Sensitive details withheld from public issue: none.

## Counterevidence Considered

- Ticket guessing is infeasible. The ticket generators use 16 random bytes and hex-encode them with typed prefixes such as `wvnc_`, `code_`, `egress_`, and `adapter_`.
- The issue does not let the attacker connect to the victim bridge. The later lease or adapter check still rejects mismatched endpoints.
- The impact is availability-only and bounded by the need to obtain a valid short-lived ticket. This is why the v3.1 score is Low and the recommended remediation is targeted rather than architectural.

steipete:
## Summary

- validate WebVNC, Code, egress, and runtime-adapter endpoint bindings before deleting single-use tickets
- preserve valid tickets presented to the wrong lease, egress role, or adapter endpoint
- serialize ticket validation and deletion with a dedicated bridge-ticket mutex to retain single-use behavior under concurrency
- add regression coverage and changelog credit

Fixes https://github.com/openclaw/crabbox/issues/371

## Verification

- `npm test --prefix worker` (22 files, 609 tests)
- Codex autoreview: clean, no accepted/actionable findings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### Fixed
 
+- Preserved single-use bridge tickets when presented to the wrong lease, role, or runtime-adapter endpoint. Thanks @coygeek.
 - Required lease manage access before resetting another operator's WebVNC bridge. Thanks @coygeek.
 - Aligned the `apple-container` provider fallback image with the portable OS default while preserving explicit image choices. Thanks @coygeek.
 - Fixed `apple-container` inventory parsing for Apple container 1.0 object-form status and nested network addresses. Thanks @coygeek.
--- a/worker/src/fleet.ts
+++ b/worker/src/fleet.ts
@@ -306,6 +306,15 @@ interface EgressTicketRecord {
   expiresAt: string;
 }
 
+type LeaseBridgeTicketConsumption<T> =
+  | { status: "invalid" }
+  | { status: "not_found" }
+  | { status: "accepted"; ticket: T; lease: LeaseRecord };
+
+type RuntimeAdapterTicketConsumption =
+  | { status: "invalid" }
+  | { status: "accepted"; ticket: RuntimeAdapterTicketRecord };
+
 interface EgressSessionStatus {
   leaseID: string;
   sessionID: string;
@@ -642,6 +651,7 @@ export class FleetCoordinator {
   private readonly restoredLeaseBridgeSockets = new Set<WebSocket>();
   private bridgeRestoreReady: Promise<boolean> | undefined;
   private readyPoolBorrowQueue: Promise<void> = Promise.resolve();
+  private bridgeTicketQueue: Promise<void> = Promise.resolve();
   private awsIngressBarrier: Promise<void> = Promise.resolve();
   private readonly awsIngressAdditiveOperations = new Set<Promise<void>>();
   private providerMaintenanceQueue: Promise<void> = Promise.resolve();
@@ -5408,17 +5418,17 @@ export class FleetCoordinator {
         { status: 426 },
       );
     }
-    const ticket = await this.consumeWebVNCTicket(request);
-    if (!ticket) {
+    const consumed = await this.consumeWebVNCTicket(request, identifier);
+    if (consumed.status === "invalid") {
       return json(
         { error: "webvnc_ticket_required", message: "valid WebVNC bridge ticket required" },
         { status: 401 },
       );
     }
-    const lease = await this.getLease(ticket.leaseID);
-    if (!lease || !identifierMatchesLease(identifier, lease)) {
+    if (consumed.status === "not_found") {
       return notFound();
     }
+    const { lease } = consumed;
     const error = webVNCLeaseError(lease);
     if (error) {
       return json({ error: "webvnc_unavailable", message: error }, { status: 409 });
@@ -5511,17 +5521,17 @@ export class FleetCoordinator {
         { status: 426 },
       );
     }
-    const ticket = await this.consumeEgressTicket(request);
-    if (!ticket || ticket.role !== role) {
+    const consumed = await this.consumeEgressTicket(request, identifier, role);
+    if (consumed.status === "invalid") {
       return json(
         { error: "egress_ticket_required", message: "valid egress bridge ticket required" },
         { status: 401 },
       );
     }
-    const lease = await this.getLease(ticket.leaseID);
-    if (!lease || !identifierMatchesLease(identifier, lease)) {
+    if (consumed.status === "not_found") {
       return notFound();
     }
+    const { lease, ticket } = consumed;
     if (lease.state !== "active") {
       return json({ error: "egress_unavailable", message: "lease is not active" }, { status: 409 });
     }
@@ -5781,13 +5791,14 @@ export class FleetCoordinator {
         { status: 426 },
       );
     }
-    const ticket = await this.consumeRuntimeAdapterTicket(request);
-    if (!ticket || ticket.adapterID !== adapterID) {
+    const consumed = await this.consumeRuntimeAdapterTicket(request, adapterID);
+    if (consumed.status === "invalid") {
       return json(
         { error: "adapter_ticket_required", message: "valid runtime adapter ticket required" },
         { status: 401 },
       );
     }
+    const { ticket } = consumed;
     const identity = await this.state.storage.get<RuntimeAdapterIdentityRecord>(
       runtimeAdapterIdentityKey(adapterID),
     );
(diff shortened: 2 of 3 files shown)


steipete:
Landed as `699a79f887f6768e2da3d2800fea137720150e01`.

Verification:
- `npm test --prefix worker` (22 files, 609 tests)
- Codex autoreview: clean, no accepted/actionable findings
- GitHub CI: all checks green, including Worker, Go, Scripts, Docs, Apple VZ, Release Check, CodeQL, and Socket Security

Regression proof covers WebVNC, Code, egress role/lease, and runtime-adapter endpoint mismatches. Mismatched tickets remain available; accepted and expired tickets remain single-use.

PR #455, opened 2026-06-18
feat: add local benchmark timing ledger

### LINKED ISSUE #342: Add local benchmark timing records and provider performance reports
## Summary

Validated against current `main` after pulling [openclaw/crabbox](https://github.com/openclaw/crabbox) on 2026-06-13 (`a71c8a3`, `feat: add OVHcloud Public Cloud provider (#304)`). Crabbox already has stable per-run timing JSON, generated provider capability docs, live smoke scripts, and broker usage summaries. The missing product primitive is a **local benchmark ledger**: a user can see whether several providers can run a workload, but cannot easily answer which eligible provider is fastest for their repo, command, machine, network, and current provider conditions.

[#301](https://github.com/openclaw/crabbox/pull/301) closed the **capability chooser** problem by generating a drift-checked provider decision matrix in `docs/providers/README.md`. This issue covers the next layer: the **observed performance chooser**.

Important framing: this should not become a public or checked-in provider leaderboard. Provider performance is local and situational. Region, instance type, cache warmth, repo size, command shape, provider load, network path, and current quota/capacity all matter. The right v1 is an opt-in local record and report surface, not Crabbox declaring one provider globally faster than another.

## Current Evidence

Crabbox already has the ingredients for local performance comparison:

- `crabbox providers --json` (`internal/cli/providers.go`, `providerMatrix()`) returns the static capability surface: provider name, family, aliases, kind, targets, features, and coordinator support.
- `--timing-json` (`internal/cli/timing.go`) emits a stable machine-readable `TimingReport` with provider, lease/run identifiers, sync timings, command timings, total duration, exit code, machine type, repo path, workdir, stop metadata, artifacts, and failure classification fields when available.
- `crabbox doctor --all --providers a,b,c` returns the live readiness surface: whether providers are configured and usable right now.
- `crabbox usage --scope user|org|all --json` returns coordinator-gated usage and cost summaries, but it is aggregated guardrail data, not a per-run billing ledger. Direct-from-CLI providers are not covered by coordinator usage.
- `scripts/live-smoke.sh` and provider-specific live smoke scripts already drive uniform commands across providers and often pass `--timing-json`. They prove viability, but do not persist rows, repeat measurements, aggregate results, or compare providers.
- `docs/providers/provider-metadata.json` now covers 54 provider entries across categories such as `brokerable-cloud`, `delegated-sandbox`, `direct-cloud`, `gpu-cloud`, `local-runtime`, `local-vm`, `self-hosted-virtualization`, and `service-control`. Those fields give useful bucketing context and can prevent misleading cross-category rankings.
- Recent provider additions, including `ovh`, `nvidia-brev`, `vercel-sandbox`, and `cloudflare-dynamic-workers`, make local observed performance more useful than a hand-maintained provider short list.

What is missing today:

- No opt-in persistence for `TimingReport` rows.
- No local append-only timing store.
- No `crabbox bench` command group.
- No aggregation by provider, machine type, command fingerprint, and cold/warm run bucket.
- No JSON report surface that agents or scripts can consume.
- No safe recommendation layer that refuses to rank providers when local evidence is insufficient.

Implementation caveat from the current code: SSH-lease runs build the final timing payload inside `internal/cli/run.go`, but delegated-run providers receive `RunRequest.TimingJSON` and currently write their own timing JSON through provider-local helpers. A correct benchmark ledger cannot only hook the SSH-lease `finalTimingReport` path unless v1 intentionally excludes delegated-run providers. Prefer a shared recorder/envelope path that can accept either the core SSH timing report or a delegated `RunResult`/provider-emitted timing payload.

The user-visible consequence is simple: a maintainer who can lease Hetzner, AWS, E2B, and other providers cannot answer "which is fastest for this workload here?" without manually redirecting `--timing-json` and building a spreadsheet.

## Goal

Deliver a local-first benchmark workflow built on top of the existing timing model:

1. Persist timing records only when explicitly requested.
2. Add `crabbox bench run` to run the same workload across selected providers with repeats.
3. Add `crabbox bench report` to aggregate local observations into human-readable and JSON reports.

This is additive. It should not change the meaning of existing `TimingReport` fields, add provider-specific benchmark behavior, or publish user's local performance data.

## Non-Goals

- No public leaderboard service.
- No checked-in authoritative provider ranking.
- No provider-specific benchmark branches or provider-side performance semantics.
- No change to the existing `crabbox providers --json` schema.
- No change to the meaning of existing `TimingReport` fields.
- No exact billing ledger in v1.
- No regression alerts, weekly digests, or background telemetry in v1.
- No default cross-category ranking. SSH-lease providers, delegated sandboxes, local runtimes, local VMs, GPU clouds, and service-control providers have very different cold-start and sync models.

### New explicit `--timing-record` flag

Add a persistent-storage companion to `--timing-json`:

```text
--timing-record <path>     append one benchmark timing record to <path>
--timing-record=off        explicitly disable persistence
--timing-record=default    append to the default local store
```

Add the flag first to `crabbox run`. If the implementation path is clean, also support it on `warmup`, `prewarm`, and `actions hydrate`; otherwise make those follow-up work and keep v1 focused on `run`.

Default behavior remains **off**. Commands without `--timing-record` must not write repo paths, workdirs, command display text, labels, artifact paths, or timing rows to disk.

The default store should use Crabbox's state directory, not the config directory:

```text
<CrabboxStateDir()>/timings.jsonl
```

Today `CrabboxStateDir()` resolves to `$XDG_STATE_HOME/crabbox` when set, otherwise `<user-config-dir>/crabbox/state`. Timing observations are mutable runtime state, not user configuration, so this belongs with claims/checkpoints/state rather than in `~/.config/crabbox/config.yaml` territory.

The local store is append-only JSONL. The user owns it. The CLI writes it only when requested. `bench report` reads it. The file is human-inspectable with tools like `jq` and can be deleted without affecting provider resources.

### Agent workflow contract

Agents should use ordinary `crabbox run` for ordinary execution and `crabbox bench run` for benchmark intent.

Examples of ordinary execution intent:

```text
Spin up a remote box and run tests.
Run this command on AWS.
Use the warm lease and check the failure.
```

These should stay on the normal non-recording path unless the user explicitly asks to record:

```sh
crabbox run --provider <provider> -- <command>
```

Examples of benchmark intent:

```text
Benchmark this on a remote box.
Compare the default remote box against AWS and Hetzner.
Which provider is fastest for this command?
Record timings for this workload.
Run this a few times and tell me performance.
```

These should use the benchmark-aware entrypoint:

```sh
crabbox bench run --providers <provider> --repeats 3 -- <command>
```

`bench run` records observations by default. Users should not need to remember `--timing-record=default` when they choose the benchmark workflow. Conversely, plain `crabbox run` remains non-recording by default for privacy. If a user wants one ordinary run to enter the benchmark ledger, they can pass `--timing-record=default`.

Every command that writes benchmark rows should print the destination store path and the number of observations written. If `bench report` finds no matching local observations, it should say that directly instead of implying missing data means zero, slow, or failed performance.

### Before and after user questions

The workflow change should be visible in the questions users and agents can ask.

Before this feature, Crabbox can answer capability, readiness, and one-off execution questions:

| User question | Current answer shape |
| --- | --- |
| "Can this provider run my workload?" | Use provider docs, `crabbox providers`, `crabbox doctor`, and/or a live `crabbox run`. |
| "Spin up a remote box and run tests." | Run the command and show the one-off result. No benchmark history is created by default. |
| "How long did that run take?" | Answer from the just-finished terminal summary or `--timing-json` if it was captured. |
| "Why did the run fail?" | Use the existing failure digest, logs, events, doctor, and run history paths. |
| "Which provider is fastest for this repo?" | Not reliably answerable unless the user manually saved and aggregated timing records. |
| "Show p50/p95 for the last week." | Not available unless timing data was captured outside Crabbox. |

After this feature, Crabbox can answer local-observation questions when matching benchmark rows exist:

| User question | Expected answer shape |
| --- | --- |
| "Benchmark this on a remote box." | Use `crabbox bench run --providers <provider> --repeats N -- <command>`, record observations, then summarize local median/p95 data. |
| "Compare the default remote box against AWS and Hetzner for `pnpm test`." | Run or read benchmark observations for the same command fingerprint and compare providers within compatible buckets. |
| "Which provider has been fastest locally for this command?" | Answer from `bench report`, citing sample size, median total duration, p95 when meaningful, and cold/warm bucket. |
| "Is this slow because sync is slow or command execution is slow?" | Compare median sync duration and median command duration in the grouped benchmark report. |
| "Do we have enough data to pick a default provider?" | Report whether matching local observations are sufficient; ask for more repeats rather than guessing when data is weak. |
| "Show machine-readable benchmark data for the agent." | Use `crabbox bench report --json`. |
| "Which providers have no local benchmark data yet?" | Show eligible providers with no matching local observations. |

Questions that remain out of scope or require careful caveats:

| User question | Required response |
| --- | --- |
| "What is the fastest Crabbox provider globally?" | Refuse to answer as fact. The ledger is local, workload-specific, and time-sensitive. |
| "Which provider is always cheapest?" | Refuse exact claims. Cost is optional/estimated and coordinator usage is not per-run invoice data. |
| "Which provider should everyone use?" | Refuse universal advice. Provider choice depends on workload, target OS, region, credentials, features, capacity, budget, and local observations. |
| "Did AWS get slower globally this week?" | Only local observations can be reported. Global provider performance is not measured. |
| "What did yesterday's forgotten plain `crabbox run` score?" | Only answer if that run was recorded with `bench run` or `--timing-record=default`; plain runs are not retroactively in the ledger. |
| "Can I compare E2B and Hetzner and declare a winner?" | Allow local observed totals, but warn that delegated sandbox and SSH-lease providers belong to different categories with different cold-start/sync models. |

The agent-native loop should become:

```text
Before:
1. Inspect provider docs/capabilities.
2. Maybe run provider A once.
3. Maybe run provider B once.
4. Capture timing output manually.
5. Hand-compute a comparison.
6. Caveat heavily because the data is ad hoc.

After:
1. Run `crabbox bench report` for existing local observations.
2. If data is missing or weak, run `crabbox bench run --providers ... --repeats ... -- <command>`.
3. Run `crabbox bench report` again.
4. Answer from grouped local observations.
5. Say when sample size or category mismatch makes a recommendation weak.
```

The shortest rule:

```text
Use `crabbox run` to execute.
Use `crabbox bench run` to learn performance.
Use `crabbox bench report` to answer from local evidence.
```

### Persist a benchmark record envelope

Do not overload `TimingReport` with benchmark-only meaning. Keep `TimingReport` as the stable per-run timing payload and persist an envelope around it:

```json
{
  "schemaVersion": 1,
  "recordedAt": "2026-06-13T00:00:00Z",
  "source": "run",
  "benchmark": {
    "commandFingerprint": "sha256...",
    "commandDisplay": "pnpm test",
    "repoFingerprint": "repo-name@commit-or-dirty",
    "coldRun": true,
    "repeatIndex": 1
  },
  "timing": {
    "provider": "aws",
    "syncMs": 12000,
    "commandMs": 45000,
    "totalMs": 64000,
    "exitCode": 0
  },
  "cost": {
    "estimatedUsd": 0.012,
    "basis": "provider-rate-runtime",
    "confidence": "estimated"
  }
}
```

The exact field names can change during implementation, but the boundaries should stay clear:

- `timing` is the existing `TimingReport`.
- `benchmark` is local comparison context.
- `cost` is optional and must say whether it is unavailable, estimated, or backed by a specific future per-lease lookup.

If the team prefers a flatter JSONL shape for `jq`, use namespaced fields such as `benchCommandFingerprint`, `benchColdRun`, and `costEstimatedUsd`. Avoid generic top-level `cold_run` or `estimated_cost_usd` on `TimingReport` unless Crabbox wants those fields to become universal timing semantics.

### New `crabbox bench run`

```sh
crabbox bench run --providers aws,hetzner,e2b --repeats 3 -- pnpm test:changed
crabbox bench run --providers aws,hetzner --repeats 2 --cold --warm -- pnpm check
```

`bench run` is a thin core CLI orchestrator:

- Validate providers with the existing provider registry and capability surface.
- Optionally call `doctor --all --providers ...` or share its readiness code before spending money.
- Run the same command for each selected provider.
- Record each run using `--timing-record=default` semantics.
- Distinguish cold and warm buckets explicitly.
- Tear down one-shot leases according to normal `run` lifecycle defaults.
- Exit non-zero if any selected provider fails unless a future flag explicitly allows partial results.

The command should call the same core path as `crabbox run`, not duplicate provider-specific logic. It may still need a small provider-neutral interface change so delegated-run backends can return or hand off the same timing payload that they currently print for `--timing-json`; otherwise reports would silently cover SSH-lease providers more completely than delegated providers.

### New `crabbox bench report`

```sh
crabbox bench report
crabbox bench report --since 7d --providers aws,hetzner --json
crabbox bench report --command-fingerprint <sha256> --bucket-by category
```

`bench report` reads the local JSONL store and groups successful observations by a conservative key:

```text
provider + category + kind + machineType + commandFingerprint + coldRun
```

For each group, print:

- `n`
- median total duration
- p95 total duration when `n` is large enough
- min/max or interquartile range when helpful
- median sync duration
- median command duration
- failure count or skipped failed rows
- optional cost estimate only when a defensible basis exists

The human report should be careful with language:

- "local observations", not "global provider performance"
- "insufficient data" when `n` is too small
- no cross-category winner by default
- no comparison between failed and successful groups as if they are equivalent

The `--json` shape should be stable enough for agents and scripts:

```json
{
  "schemaVersion": 1,
  "storePath": ".../timings.jsonl",
  "filters": {
    "since": "7d",
    "providers": ["aws", "hetzner"]
  },
  "groups": [
    {
      "provider": "aws",
      "category": "brokerable-cloud",
      "kind": "ssh-lease",
      "machineType": "c7a.4xlarge",
      "commandFingerprint": "sha256...",
      "coldRun": true,
      "n": 5,
      "medianTotalMs": 64000,
      "p95TotalMs": 83000,
      "medianSyncMs": 12000,
      "medianCommandMs": 45000,
      "failureCount": 0,
      "cost": {
        "medianEstimatedUsd": 0.012,
        "confidence": "estimated"
      }
    }
  ]
}
```

### Defer or constrain `crabbox bench recommend`

`recommend` is useful, but risky as v1. It can easily imply more certainty than the local data supports.

If implemented in this issue, keep it conservative:

```sh
crabbox bench recommend -- pnpm test:changed
```

Rules:

- Only rank providers with enough local observations for the matching command fingerprint and capability bucket.
- Refuse to choose a winner when data is insufficient.
- Explain every ranking with observed local metrics.
- Include live readiness from `doctor --all` as a filter or warning, not as a performance substitute.
- Never compare across categories by default.
- Mark cost as unavailable unless the cost basis is explicit.

It is acceptable, and probably better, to make `recommend` a follow-up issue after `bench report` is proven useful.

## Documentation Shape

For v1, document the benchmark workflow rather than publishing a provider leaderboard.

Add:

- `docs/commands/bench.md` for command usage and flags.
- `docs/observability.md` updates for `--timing-record` and the local timing store.
- `docs/cli.md` command tree updates.

Optional follow-up: add `docs/providers/PERFORMANCE.md` as a generated **sample report format**, not an authoritative ranking. If this is added later, generate it from synthetic or clearly sample-only fixture rows and make every table label say "sample" or "example". Do not put real maintainer-local measurements in the repo.

The provider matrix generator from #301 remains a good pattern for drift checking, but performance data is different from provider metadata:

- provider metadata is curated and static;
- performance is observed, local, and time-sensitive.

Keep those sources separate.

## Source-of-Truth Model

Two stores for v1:

1. **Local benchmark store**
   - Path: `<CrabboxStateDir()>/timings.jsonl` by default.
   - Owner: local user.
   - Writer: Crabbox CLI, only when `--timing-record` or `bench run` asks for it.
   - Reader: `bench report` and future local tools.
   - Format: append-only JSONL with a schema version.

2. **Static provider metadata**
   - Path: `docs/providers/provider-metadata.json`.
   - Owner: repository.
   - Purpose: category/kind/docs bucketing for fair grouping and explanatory labels.
   - Not a performance store.

Do not extend `provider-metadata.json` with timing numbers. That would conflate "what a provider is" with "what happened on one user's machine."

## Acceptance Criteria

- `crabbox run --timing-record <path|default|off>` is accepted and documented.
- `--timing-record` is opt-in. Default `crabbox run` behavior writes no timing store.
- `--timing-record=default` appends one valid JSONL benchmark record to `<CrabboxStateDir()>/timings.jsonl`.
- `--timing-record <path>` appends one valid JSONL benchmark record to that path, creating parent directories only when appropriate and failing clearly on permission/path errors.
- The persisted record contains the existing `TimingReport` payload without changing existing `TimingReport` field meanings.
- Persisted rows include a schema version, recorded timestamp, source, command fingerprint, cold/warm bucket when known, and enough command/repo context to group comparable runs without requiring global telemetry.
- Sensitive handling is documented: repo paths and command display may be local/private, so persistence is explicit and user-owned.
- SSH-lease and delegated-run providers either both support timing records or the docs and CLI errors clearly state which provider kinds are supported in v1. Silent partial coverage is not acceptable for a provider comparison tool.
- `crabbox bench run --providers a,b,c --repeats 2 -- pnpm test` runs the same workload across selected providers and records every attempted observation.
- `bench run` does not add provider-specific branches in core beyond existing provider registry/config routing.
- `crabbox bench report` reads the local store and prints grouped local observations with `n`, median, p95 when meaningful, sync/command breakdown, and failure counts.
- `crabbox bench report --json` emits a stable schema with report metadata, filters, and groups.
- Reports bucket by provider/category/kind/machine type/command fingerprint/cold-warm rather than declaring a global winner.
- Cost is omitted or explicitly marked unavailable unless a defensible estimate exists.
- Documentation is updated in `docs/commands/bench.md`, `docs/observability.md`, and `docs/cli.md`.
- Existing checks still pass: `scripts/check-command-docs.mjs`, `scripts/check-provider-matrix.mjs`, `scripts/check-docs-links.mjs`, `go test ./...`.

### Slice 1: Timing persistence

Add the benchmark record type and append-only JSONL writer.

Likely code areas:

- `internal/cli/timing.go` for shared timing record helpers.
- `internal/cli/run.go` for `--timing-record`.
- `internal/cli/provider_backend.go` if delegated-run backends need to return timing data instead of only printing it.
- Provider-local `core.go` timing helpers for delegated providers if the shared timing writer needs to move closer to `internal/cli`.
- Tests beside existing timing/run tests.
- Docs for `run` and observability.

Exit criteria:

- `crabbox run --timing-record <tempfile> -- <command>` appends exactly one valid JSONL record.
- Default `crabbox run` does not create or touch the timing store.
- Existing `--timing-json` output remains compatible.

### Slice 2: `bench report`

Add read/aggregate/report functionality before adding orchestration. This makes the data model testable without cloud credentials.

Likely code areas:

- `internal/cli/bench.go`
- unit tests with fixture JSONL rows
- `app.go` command registration
- `docs/commands/bench.md`

Exit criteria:

- Fixture rows produce deterministic human and JSON reports.
- Failed rows are either shown separately or counted clearly.
- Small sample sizes are labeled as insufficient instead of over-interpreted.

### Slice 3: `bench run`

Add the fan-out runner after persistence and reporting are stable.

Likely code areas:

- `internal/cli/bench.go`
- command invocation/shared run path plumbing
- tests with fake providers or command harnesses where possible

Exit criteria:

- `bench run --providers a,b --repeats 2 -- <cmd>` records rows for each attempted provider.
- Provider failure handling is clear and exits non-zero by default.
- Normal lease cleanup semantics are preserved.

### Follow-up: `bench recommend`

Add only after `bench report` has settled.

Rules:

- require enough matching local observations;
- explain rankings with observed metrics;
- show readiness/capability blockers separately;
- refuse to rank when data is weak.

### Follow-up: generated sample performance docs

If maintainers still want a generated `docs/providers/PERFORMANCE.md`, generate it from synthetic fixture rows and frame it as a sample report format. Do not present it as a live provider leaderboard.

## Risk Controls

- **Privacy:** persistence is opt-in because timing records can include local repo paths, command display, workdirs, labels, and artifact paths.
- **Misleading rankings:** default reports bucket by category/kind and require sufficient local observations before comparative language.
- **Cost ambiguity:** cost is optional and explicitly labeled by basis/confidence.
- **Provider drift:** provider metadata can be joined for labels, but provider timing remains local observed data.
- **Scope creep:** v1 ends at persistence plus run/report. Recommendation, public sharing, regression alerts, and generated leaderboard docs are follow-ups.

## Why This Is the Right Boundary

Crabbox is a local CLI that already owns sync, command execution, timing summaries, and provider-neutral orchestration. Provider adapters should not learn benchmark behavior. The benchmark layer belongs in core because it compares providers using common already-emitted timing data.

The product should optimize for repeatable local truth:

- "Here is what happened on my machine for this command."
- "Here is enough data to compare similar providers honestly."
- "Here is JSON an agent can consume."

It should avoid global claims:

- "Provider X is fastest."
- "This checked-in fixture is a leaderboard."
- "Aggregated monthly usage equals per-run cost."

That boundary gives users something immediately useful while keeping Crabbox's architecture, privacy posture, and documentation honest.

TurboTheTurtle:
Closes https://github.com/openclaw/crabbox/issues/342.

## Summary

- add an opt-in local benchmark timing ledger around existing `TimingReport` payloads
- add `run --timing-record`, `bench run`, `bench record`, and evidence-aware `bench report --json`
- keep observations local, report successful sample and failure counts, and mark insufficient evidence instead of ranking providers globally
- preserve delegated providers' complete timing metadata and the existing single final `--timing-json` payload contract

## Verification

- `go test -race ./internal/cli`
- `go vet ./...`
- `scripts/check-docs.sh`
- focused benchmark/delegated timing regression tests
- autoreview: clean after preserving provider-authored delegated sync timing and phases

## Real behavior proof

Built the CLI from head `1a62922c` and exercised a real Docker-backed `local-container` workflow:

- direct `run --timing-record`: one successful observation, automatic lease cleanup
- `bench run --provider local-container --repeats 2 -- true`: two observations, zero failures, automatic cleanup after both runs
- `bench report --provider local-container --min-samples 2 --json`: three matched successful observations, `providerCategory=local-runtime`, `insufficientEvidence=false`, `evidence=sufficient_local_samples`

The delegated regression coverage also proves that recording captures provider-authored sync duration, sync phases, and machine metadata; `--timing-json` remains one final JSON object, and timing JSON is not printed when only recording is requested.


TurboTheTurtle:
@clawsweeper re-review

Requesting ClawSweeper review for this new PR.

TurboTheTurtle:
@clawsweeper re-review

TurboTheTurtle:
@clawsweeper re-review

steipete:
After-fix proof for https://github.com/openclaw/crabbox/pull/455 at `1a62922c`.

Fixes:

- delegated timing capture preserves provider-authored sync timing and keeps `--timing-json` to one final payload: https://github.com/openclaw/crabbox/blob/1a62922cf2684467986acef6fbf7c8d334dc7b0e/internal/cli/run.go
- benchmark rows and reports now populate drift-checked provider categories from canonical provider metadata: https://github.com/openclaw/crabbox/blob/1a62922cf2684467986acef6fbf7c8d334dc7b0e/internal/cli/bench.go

Redacted real Docker-backed output:

```text
$ crabbox run --provider local-container --timing-record <store> -- true
run summary sync=1.375s command=64ms total=1.454s sync_skipped=false exit=0
lease cleanup stopped=true policy=auto
benchmark timing record appended path=<store> observations=1

$ crabbox bench run --store <store> --provider local-container --repeats 2 -- true
benchmark timing record appended path=<store> observations=1
benchmark timing record appended path=<store> observations=1
benchmark run completed path=<store> observations=2 failures=0

$ crabbox bench report --store <store> --provider local-container --min-samples 2 --json
{"schemaVersion":1,"observationCount":3,"matchedCount":3,"groups":[{"provider":"local-container","providerFamily":"container","providerKind":"ssh-lease","providerCategory":"local-runtime","n":3,"failureCount":0,"insufficientEvidence":false,"evidence":"sufficient_local_samples"}]}
```

Additional proof: `go test -race ./internal/cli`, `go vet ./...`, `scripts/check-docs.sh`, provider-metadata drift check, and structured autoreview all pass.

PR #329, opened 2026-06-13
[codex] fix Windows rsync fallback for local containers

brokemac79:
﻿## What changed

- Only use WSL rsync on Windows when the selected WSL distro has native `rsync` and `ssh` available.
- Reject WSL probe results that resolve to Windows interop shims, including `.exe`, `/mnt/c`, and `/mnt/host/c` command paths.
- Fall back to native MSYS2/Cygwin rsync with the existing path-conversion safeguards when WSL is present but unsuitable.
- Copy the SSH key and known_hosts data into WSL through stdin instead of relying on WSL mount-path access to the Windows key path.
- Add `/mnt/host`-aware Windows-to-WSL path conversion coverage and focused shim-rejection coverage.

## Why

On Windows hosts where the default WSL distro is `docker-desktop`, `wsl.exe` can exist while the distro lacks native `rsync` and `ssh`. Crabbox then tried `wsl rsync`, which could fall through to Windows PATH shims while still using WSL `/tmp` key paths. That caused synced local-container runs to fail because the SSH identity path was not accessible from the process that actually ran rsync.

## Validation

- `docker run --rm -v "C:\oc-work\crabbox-pr-windows-rsync:/src" -w /src golang:1.26 sh -lc "/usr/local/go/bin/go test ./internal/cli -run 'Test(RsyncLocalPath|WindowsToWSL|WindowsWSL|NormalizeRsyncOptions)' && /usr/local/go/bin/go test ./internal/cli -run 'TestWindows'"`
- Built `crabbox.exe` from this branch as `0.30.1-dev`.
- Synced local-container smoke passed from Windows with `provider=local-container`, lease `cbx_e28706413742`, sync enabled, and `rsync` phase completing in 1.522s:
  `crabbox run --provider local-container --no-hydrate --timing-json --label windows-rsync-pr-smoke-rankup --shell -- "printf 'sync-ok\n'; test -f go.mod; uname -a"`

## Note

A broader `go test ./internal/cli` run from the Windows-mounted Docker volume previously failed in unrelated checkpoint/pond ACL golden tests with newline/golden-output differences; the focused rsync/path and Windows tests above pass.

PR #487, opened 2026-06-20
fix(sync): use native rsync when WSL2 cannot reach the rsync target

anagnorisis2peripeteia:
## Problem

On a Windows host, `windowsRsyncCommand` runs rsync inside WSL whenever the default distro has native `rsync` + `ssh`. But WSL2 has its own NAT network namespace, so a target reachable **only from the Windows host** — e.g. a Hyper-V guest on the Default Switch (host-internal NAT) — is unroutable from WSL. The WSL rsync's `-e ssh` transport then blocks on connect, and because `Sync.Timeout` is unset by default the sync **hangs indefinitely, transferring nothing** into the guest workroot.

The selection only checks whether WSL *has* rsync+ssh (`windowsWSLHasNativeRsyncSSH`), never whether WSL can *reach* the target — so on a Win + Hyper-V + WSL2 host every sync to a Default-Switch guest hangs.

## Fix

- `windowsRsyncCommand`: before committing to the WSL rsync path, probe whether WSL can open a TCP connection to the target's SSH endpoint (`windowsWSLCanReachTarget`, `timeout`-bounded). If it can't, fall back to native rsync — the host's own network + `ssh.exe` reach the target. Provider-neutral; unchanged for targets WSL can already reach. Extends the native-fallback hardening from #329 from tool *presence* to *reachability*.
- `normalizeRsyncOptions`: default `Timeout` to 30m when unset, so a stalled sync surfaces rsync's existing actionable timeout error instead of hanging forever. A generous backstop, not a tight deadline.

## Tests

- `TestWSLReachabilityProbeScriptDefaultsToSSHPort` — table-driven; explicit / empty / whitespace ports.
- `TestWindowsWSLCanReachTargetEmptyHostKeepsWSLPath` — empty host short-circuits to reachable without invoking `wsl` (default WSL path preserved).
- `go test ./internal/cli/` passes; `gofmt` clean.

## Not covered by CI

Live end-to-end Hyper-V sync through the fallback — there's no Hyper-V runner in CI and the provider has no e2e. Validating manually on a real Windows + Hyper-V + WSL2 host; I'll post the result as a comment.

## Risk / mitigation

- One `timeout`-bounded (≤8s) `wsl` exec per Windows rsync call to probe reachability. Empty-host short-circuits without exec.
- The default timeout changes behaviour for syncs that previously ran unbounded; set generously (30m) so real source syncs are unaffected.

### DIFF
--- a/internal/cli/ssh.go
+++ b/internal/cli/ssh.go
@@ -626,18 +626,9 @@ type rsyncOptions struct {
 	HeartbeatInterval time.Duration
 }
 
-// defaultRsyncTimeout bounds a sync that would otherwise block forever (for
-// example an SSH transport that can never reach the target). It is a generous
-// backstop, not a tight deadline: real source syncs finish well inside it, and
-// hitting it surfaces rsync's actionable timeout error instead of hanging.
-const defaultRsyncTimeout = 30 * time.Minute
-
 func normalizeRsyncOptions(opts rsyncOptions) rsyncOptions {
 	if opts.NoTimes {
 		opts.Checksum = true
-	}
-	if opts.Timeout <= 0 {
-		opts.Timeout = defaultRsyncTimeout
 	}
 	return opts
 }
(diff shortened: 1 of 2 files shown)


anagnorisis2peripeteia:
Pushed 11a81811 addressing the review:

- **Shell-injection (P1):** `host`/`port` are now validated against a hostname/IP/port-only charset (`isSafeProbeToken`) before they reach the probe shell; anything with a metacharacter is never handed to a shell and falls back to native rsync. Added hostile-string regression tests (`TestIsSafeProbeToken`, `TestWindowsWSLCanReachTargetRejectsHostileHost`).
- **Timeout default (P2):** dropped — `defaultConfig` already sets a 15m sync timeout, so whether an explicit/internal zero stays unbounded is a maintainer call, not this PR's.
- **bash/timeout assumption (P2):** a WSL distro missing either makes the probe error → native-rsync fallback (the safe path), so it degrades gracefully rather than changing correctness.

`gofmt` clean; `go test ./internal/cli` green in CI. Real Windows + Hyper-V + WSL2 behaviour proof (WSL unreachable → native fallback → workroot transfer) to follow as a comment.

@clawsweeper re-review

anagnorisis2peripeteia:
Closing this.

On further investigation, crabbox already syncs Windows-native targets via the archive path (`syncWindowsNative` in `sync_windows_target.go`) — so `windowsRsyncCommand`, the function this PR patches, is never on the Windows-guest sync path. That archive path predates the binary I was testing against, so Windows-guest sync has never used WSL rsync.

The sync failure that motivated this turned out to be a local test-harness issue on my side (a double-provision contention bug in my own runner script), not a crabbox bug. Since this change hardens a path the motivating scenario never exercises and can't be reproduced as described, I'd rather close it than leave an unverifiable change open. Apologies for the churn.

PR #510, opened 2026-06-21
fix: reject noncanonical signed user tokens

### LINKED ISSUE #499: [security] Signed user token malleability bypasses Code portal logout revocation
## Summary

Crabbox signed browser-login tokens are minted as `cbxu_<payload>.<signature>`, where the HMAC signature covers the base64url payload. At commit `184c2258dbcad99c5e6fc1b0a20b033b45b22443`, `worker/src/auth.ts` authenticates the token body with `raw.split(".", 2)`. JavaScript truncates the split result to the first two segments, so `cbxu_<payload>.<signature>.<suffix>` authenticates as the same signed GitHub user even though Crabbox never minted that exact token string.

This crosses a supported Crabbox boundary because the Code portal logout flow stores revocation state by hashing the raw portal session token string. A token holder can authenticate with a suffixed spelling of the same signed token after the canonical portal session has been logged out and revoked, producing a different revocation key while preserving the same signed owner, org, login, and expiry.

## Affected Components

- `worker/src/auth.ts`
  - `issueUserToken` mints exactly `cbxu_<encodedPayload>.<sig>`.
  - `verifyUserToken` accepts the first two dot-delimited fields and ignores any trailing fields.
  - `decodeUserTokenPayload` also reads only the first payload field.
- `worker/src/coordinator-entry.ts`
  - `requestWithPortalCookie` converts the `crabbox_session` cookie into an `Authorization: Bearer` credential before normal authentication.
- `worker/src/fleet.ts`
  - `redirectCodeViewer` hashes the raw authenticated bearer token into `portalSessionHash`.
  - `portalLogout` hashes the raw logout cookie token and writes a `code-viewer-session-revocation:<hash>` record.
  - `consumeCodeViewerTicket`, `codeViewerSession`, and `codeViewerWebSocket` enforce the revocation record keyed by that raw-token hash.
- `worker/test/fleet.test.ts`
  - The existing Code portal logout test shows the intended control: after logout, the canonical portal token cannot bootstrap a stale Code viewer session.

## Attack Path

1. A signed GitHub user token is issued by the normal browser login flow and stored in the `crabbox_session` portal cookie.
2. The canonical token is used to open an isolated Code viewer. The coordinator records the viewer session against `sha256Hex(<canonical token>)`.
3. The user logs out through `/portal/logout`. Crabbox clears the browser cookie, records a revocation for `sha256Hex(<canonical token>)`, and closes active Code viewers for that portal session.
4. A party that can still present the same signed token value, for example from a copied portal cookie or another bearer-token client, appends an arbitrary dot-delimited suffix and sends it as `crabbox_session=cbxu_<payload>.<signature>.anything`.
5. `authenticateRequest` accepts the suffixed token because `verifyUserToken` validates only the first two token segments.
6. `redirectCodeViewer` hashes the suffixed raw token string. That hash differs from the canonical token hash stored during logout, so the Code portal revocation check does not match and a fresh Code viewer ticket can be issued for the same signed identity.

The prerequisite is possession of an otherwise valid signed user credential; this is not a hostile multi-tenant isolation claim. The reportable boundary is narrower: Crabbox's own signed-token parser permits non-canonical credential spellings that bypass the Code portal logout revocation control documented and tested for browser-login sessions.

## Impact

Portal logout is intended to end isolated Code viewer access tied to the signed GitHub browser session. Because the authenticator accepts multiple raw strings for the same signed token, the logout revocation list can be sidestepped by changing the token spelling without changing the signed payload or signature.

For a lease with Code enabled, a successful bypass can restore browser Code access as the signed lease owner until the underlying user token expires or another server-side control blocks the lease. That can expose repository contents, terminal/editor state, and Code-server actions available through the viewer. The issue does not grant admin by itself, does not forge a different owner or org, and does not help without a valid signed user token.

### CVSS Assessment

| Metric | v3.1 | v4.0 |
|--------|------|------|
| **Score** | 8.3 / 10.0 | 7.6 / 10.0 |
| **Severity** | High | High |
| **Vector** | CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:U/C:H/I:H/A:L | CVSS:4.0/AV:N/AC:L/AT:P/PR:L/UI:N/VC:H/VI:H/VA:L/SC:N/SI:N/SA:N |
| **Calculator** | [CVSS v3.1 Calculator](https://www.first.org/cvss/calculator/3.1#CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:U/C:H/I:H/A:L) | [CVSS v4.0 Calculator](https://www.first.org/cvss/calculator/4.0#CVSS:4.0/AV:N/AC:L/AT:P/PR:L/UI:N/VC:H/VI:H/VA:L/SC:N/SI:N/SA:N) |

## Recommended Remediation

- Parse signed user tokens without a truncating split.
- Require exactly two non-empty token segments after the `cbxu_` prefix.
- Keep the existing signature, expiry, type, owner, org, login, and `admin`-claim checks.
- Add a regression test in `worker/test/http.test.ts` proving `token + ".ignored"` is rejected by `authenticateRequest`.
- Add a focused Code portal regression test showing that a suffixed spelling of a logged-out portal token cannot obtain a new Code viewer ticket.

One minimal parser change is:

```ts
const parts = raw.split(".");
if (parts.length !== 2) return undefined;
const [encodedPayload, signature] = parts;
```

## Validation

Source validation was performed against the isolated checkout at commit `184c2258dbcad99c5e6fc1b0a20b033b45b22443`.

- `crabbox/SECURITY.md` defines authentication, ownership, sharing controls, cross-owner access, and generated-credential exposure as supported Crabbox boundaries, while excluding hostile shared-tenant assumptions and leaked credentials outside Crabbox's control. This finding is scoped to the signed browser-login and Code portal logout controls, not to hostile tenants sharing a broker.
- `crabbox/docs/security.md` documents signed `cbxu_` user tokens as HMAC-SHA256 signed payloads issued by GitHub browser login and used to scope owner/org identity.
- `worker/src/auth.ts` mints signed user tokens with exactly one payload segment and one signature segment in `issueUserToken`.
- `worker/src/auth.ts` verifies user tokens with `raw.split(".", 2)`, so an extra third segment is ignored during signature verification.
- `worker/src/coordinator-entry.ts` turns a `crabbox_session` portal cookie into a bearer token before calling `authenticateRequest`.
- `worker/src/fleet.ts` computes Code portal revocation state from `sha256Hex(token)`, where `token` is the raw bearer or cookie value. The canonical token and a suffixed token therefore produce different revocation keys even though both authenticate to the same signed identity.
- `worker/test/fleet.test.ts` already contains an end-to-end test proving the intended Code portal behavior: after `/portal/logout`, the canonical portal session cannot reuse a pending Code viewer bootstrap or open a new Code portal session.

Counterevidence considered: the bug does not forge a signature, change the signed owner/org/login payload, mint admin rights, or authenticate without a valid signed user token. It is still reportable because Crabbox itself uses raw signed-token strings as Code portal session revocation keys, and the authenticator accepts non-canonical spellings that evade that revocation check.

steipete:
## Summary

- parse signed user tokens through one canonical two-segment boundary
- reject suffixed or otherwise non-canonical spellings before identity or expiry decoding
- prove altered portal cookies cannot obtain a fresh isolated Code viewer ticket after logout

Fixes https://github.com/openclaw/crabbox/issues/499

## Verification

- `npm test --prefix worker -- http.test.ts fleet.test.ts` (306 tests)
- `npm test --prefix worker` (646 tests)
- Worker format, lint, typecheck, and dry-run build
- `bash scripts/check-docs.sh`
- `go test -race ./...`
- `go vet ./...`
- structured autoreview: clean

## Live proof

The patched Worker ran under local Wrangler with a freshly HMAC-signed user token. Actual HTTP results:

```text
canonical_status=200
suffixed_status=401
[wrangler:info] GET /v1/whoami 200 OK
[wrangler:info] GET /v1/whoami 401 Unauthorized
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 
 ### Fixed
 
+- Rejected non-canonical signed browser-session tokens so suffix changes cannot bypass Code portal logout revocation. Thanks @coygeek.
 - Required a matching local claim before Cloudflare container reuse, status, or stop operations can reach the runner. Thanks @coygeek.
 - Redacted configured Freestyle API keys from lifecycle, command, and file-operation error diagnostics. Thanks @coygeek.
 - Redacted configured OpenComputer API keys from control-plane and upload error diagnostics. Thanks @coygeek.
--- a/docs/features/portal.md
+++ b/docs/features/portal.md
@@ -83,6 +83,7 @@ successful GitHub login through the same OAuth flow as `crabbox login`. The
 Worker converts the cookie to a `Bearer` token internally; an unauthenticated
 GET request to a portal page is redirected to `/portal/login` with a
 `returnTo` parameter. The session carries owner/org claims, and the Worker
+rejects altered or suffixed spellings of its signed session tokens. The session
 scopes every page to that identity.
 
 ```text
--- a/worker/src/auth.ts
+++ b/worker/src/auth.ts
@@ -203,14 +203,11 @@ async function verifyUserToken(
   token: string,
   env: Pick<Env, "CRABBOX_SHARED_TOKEN" | "CRABBOX_SESSION_SECRET">,
 ): Promise<UserTokenPayload | undefined> {
-  if (!token.startsWith(tokenPrefix)) {
+  const parts = userTokenParts(token);
+  if (!parts) {
     return undefined;
   }
-  const raw = token.slice(tokenPrefix.length);
-  const [encodedPayload, signature] = raw.split(".", 2);
-  if (!encodedPayload || !signature) {
-    return undefined;
-  }
+  const { encodedPayload, signature } = parts;
   const expected = await sign(encodedPayload, sessionSecret(env));
   if (!constantTimeEqual(signature, expected)) {
     return undefined;
@@ -231,19 +228,28 @@ async function verifyUserToken(
 }
 
 function decodeUserTokenPayload(token: string): Partial<UserTokenPayload> {
-  if (!token.startsWith(tokenPrefix)) {
-    return {};
-  }
-  const raw = token.slice(tokenPrefix.length);
-  const [encodedPayload] = raw.split(".", 1);
-  if (!encodedPayload) {
+  const parts = userTokenParts(token);
+  if (!parts) {
     return {};
   }
   try {
-    return JSON.parse(decoder.decode(base64URLDecode(encodedPayload))) as Partial<UserTokenPayload>;
+    return JSON.parse(
+      decoder.decode(base64URLDecode(parts.encodedPayload)),
+    ) as Partial<UserTokenPayload>;
   } catch {
     return {};
   }
+}
+
+function userTokenParts(token: string): { encodedPayload: string; signature: string } | undefined {
+  if (!token.startsWith(tokenPrefix)) {
+    return undefined;
+  }
+  const parts = token.slice(tokenPrefix.length).split(".");
+  if (parts.length !== 2 || !parts[0] || !parts[1]) {
+    return undefined;
+  }
+  return { encodedPayload: parts[0], signature: parts[1] };
 }
 
 function sessionSecret(env: Pick<Env, "CRABBOX_SHARED_TOKEN" | "CRABBOX_SESSION_SECRET">): string {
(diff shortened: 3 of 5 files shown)


steipete:
@clawsweeper re-review

The valid generic-product fixture finding is fixed. The changelog entry remains intentionally: this is a maintainer-authored user-visible security fix, and repository policy requires changelog coverage.

PR #689, opened 2026-06-25
fix(windows): make tightvnc startup deterministic

fcoury-oai:
## Why

Azure Windows desktop leases can show the Windows Firewall prompt for `TightVNC Server` the first time the per-user VNC process starts. That modal blocks the visible desktop and can sit in front of automation that is trying to inspect or drive the target application.

The VNC bootstrap should make the desktop transport ready without depending on a human click or on best-effort registry timing.

| Example |
|---|
| image |

## What Changed

- Add a `Crabbox-TightVNC-Loopback` Windows Firewall allow rule for `tvnserver.exe` on TCP port `5900` before the user-session VNC process is scheduled or started.
- Make the per-user TightVNC startup fail loudly if the HKCU password values cannot be copied from the machine-wide TightVNC configuration.
- Wait for both `Password` and `ControlPassword` registry values instead of assuming they are immediately available after install.
- Use ScheduledTasks cmdlets for the interactive `CrabboxUserVNC` task so the task definition is explicit about interactive logon and highest run level.
- Mirror the same bootstrap behavior in the worker-side Windows bootstrap generator.

## How to Test

1. Build the local CLI from this branch:

   ```bash
   ```

2. Acquire a Windows desktop lease using the Azure provider.
3. Connect to the lease through `crabbox vnc`.
4. Confirm that the desktop is reachable through VNC without a Windows Security firewall prompt for `TightVNC Server`.
5. Confirm that the VNC server still requires VNC authentication and only listens on the expected loopback-forwarded path.

Targeted tests:

- `go test ./internal/cli -run TestAWSUserDataWindowsProfile -count=1`
- `npm test -- bootstrap.test.ts` from `worker/`

## Documentation

No developers.openai.com/codex documentation update is needed.

### DIFF
--- a/internal/cli/bootstrap.go
+++ b/internal/cli/bootstrap.go
@@ -1,3 +1,4 @@
+// Builds provider bootstrap scripts that prepare disposable Crabbox machines for SSH and desktop access.
 package cli
 
 import (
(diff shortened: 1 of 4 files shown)


steipete:
Superseded by https://github.com/openclaw/crabbox/commit/184979d683b79f6ad60411013ea44fca5db3400d.

Current `main` removes the per-user TightVNC startup path entirely and runs the authenticated TightVNC service instead. That addresses the reported first-login firewall modal and registry timing race without adding TCP/5900 firewall ingress.

Verified on current `main`:

- `go test ./internal/cli -run 'TestAWSUserDataWindowsProfile|Test.*Windows.*VNC|Test.*WebVNC' -count=1`
- `npm test --prefix worker -- bootstrap.test.ts` (17 tests)
- Azure Windows desktop acquisition was attempted with `Standard_D4s_v5`, but the coordinator returned no VM/host before its 30-minute timeout; lease `cbx_7538f819aace` was released, so no live desktop claim is being made.

Thanks @fcoury-oai for isolating the modal and startup race. The report was useful; the safer service-mode fix landed independently while this PR was open.

PR #743, opened 2026-07-01
fix: validate forwarded environment names

### LINKED ISSUE #695: [security] Invalid allowlisted environment variable names can inject commands into SSH-backed runs
## Summary

Crabbox documents environment forwarding as name-based and states that names outside `[A-Za-z_][A-Za-z0-9_]*` are dropped during forwarding. The current POSIX SSH-backed run path does not enforce that rule for inline environment values selected from the local process environment. `allowedEnv` accepts any `os.Environ()` key that matches the configured allowlist, and `writeRemoteCommandPrefix` writes that key directly into the remote shell prefix as `KEY='value'` while only quoting the value.

When an operator uses a wildcard or broad allowlist such as `PROJECT_*`, a local environment entry with a name like `PROJECT_$(touch /tmp/cbx-env-pwn)#` is accepted and rendered into the SSH command prefix. Bash performs command substitution while parsing that malformed assignment, so the injected command runs on the leased box before the intended command starts or before bash reports the malformed assignment.

This crosses Crabbox's supported boundary because the maintainer-authored policy treats command injection outside an explicitly configured command, helper, runtime, or project-automation surface as in scope. The issue does not depend on hostile tenants sharing a broker or on a lease owner intentionally running their own command; it is an unexpected command created by Crabbox's inline env-name rendering.

## Affected Components

- Checked commit: `0ec69d6427646c7bddffbac2c33576f211c203b1`
- Component: POSIX SSH-backed `crabbox run` command construction
- `internal/cli/repo.go:175-187`: `allowedEnv` splits `os.Environ()` entries and accepts raw keys based only on `envAllowed`.
- `internal/cli/run_env_profile.go:51-57`: `allowedEnvWithoutProfileKeys` returns the same unchecked local-environment keys as inline forwarded env.
- `internal/cli/run.go:449-454` and `internal/cli/run.go:1421-1440`: `selectRunEnv` output is merged into `runEnv`, then passed to the POSIX remote command builders.
- `internal/cli/ssh.go:1139-1165`: POSIX command and shell command paths call `writeRemoteCommandPrefix` before invoking `bash -lc`.
- `internal/cli/run_script.go:154-171`: POSIX `--script` runs also call `writeRemoteCommandPrefix`.
- `internal/cli/ssh.go:1241-1262`: `writeRemoteCommandPrefix` writes each env key directly with `b.WriteString(k)` and quotes only the value.

Counterevidence checked:

- `internal/cli/run_env_profile.go:92-119`, `:370-391`, and `:485-501` already validate env names for parsed env-profile files and generated env files.
- `internal/cli/profiles.go:60-68` validates configured profile and preset env keys before they are expanded.
- Those controls do not sanitize keys coming from the process environment through `allowedEnv`.

## Attack Path

Attacker role: a local environment influence that can place an unusual variable name into the operator's Crabbox process environment, without controlling the explicit command the operator asks Crabbox to run.

Prerequisites:

- The operator uses a POSIX SSH-backed provider path.
- The resolved environment allowlist contains a wildcard or broad prefix that matches the hostile name, for example `PROJECT_*`.
- The operator's Crabbox process environment contains a variable whose name includes shell metacharacters and matches that allowlist, for example `PROJECT_$(touch /tmp/cbx-env-pwn)#=1`.

Steps:

1. The hostile environment entry is present in the local Crabbox process environment.
2. The operator runs a normal command such as `crabbox run -- make test` with an allowlist that matches the hostile variable name.
3. `allowedEnv` accepts the raw key because `envAllowed` only checks exact or prefix allowlist matching.
4. `run.go` merges the selected inline env into `runEnv` and builds a POSIX remote command.
5. `writeRemoteCommandPrefix` emits a prefix shaped like:

   ```text
   cd '<workdir>' && PROJECT_$(touch /tmp/cbx-env-pwn)#='1' bash -lc '<script>'
   ```

6. Bash performs the `$(touch /tmp/cbx-env-pwn)` substitution while parsing the malformed assignment, so the injected command executes on the leased machine as the Crabbox SSH user.

## Impact

An attacker who can influence the local process environment can cause Crabbox to run an unintended command on the leased POSIX machine when a matching wildcard or broad env allowlist is active. The command runs with the same remote OS privileges as the intended Crabbox command, so it can read or modify data available to the lease user and alter the validation environment before the operator's requested command executes.

The impact is bounded by trusted-operator assumptions and the selected lease user's authority, so this is not a hostile multi-tenant isolation break. It remains reportable because the current behavior violates the documented env-forwarding contract and creates command execution outside the operator's explicit Crabbox command, helper, runtime, or project-automation surface.

### CVSS Assessment

| Metric | v3.1 | v4.0 |
|--------|------|------|
| **Score** | 6.0 / 10.0 | 5.6 / 10.0 |
| **Vector** | CVSS:3.1/AV:L/AC:L/PR:H/UI:N/S:C/C:H/I:N/A:N | CVSS:4.0/AV:L/AC:L/AT:P/PR:H/UI:N/VC:H/VI:N/VA:N/SC:N/SI:N/SA:N |
| **Calculator** | [CVSS v3.1 Calculator](https://www.first.org/cvss/calculator/3.1#CVSS:3.1/AV:L/AC:L/PR:H/UI:N/S:C/C:H/I:N/A:N) | [CVSS v4.0 Calculator](https://www.first.org/cvss/calculator/4.0#CVSS:4.0/AV:L/AC:L/AT:P/PR:H/UI:N/VC:H/VI:N/VA:N/SC:N/SI:N/SA:N) |

The score is medium because exploitation requires a local environment precondition plus a matching allowlist, and the resulting command runs with the same remote user authority as the intended lease command. Scope changes in v3.1 because the local CLI constructs command text that executes on the remote leased machine.

## Recommended Remediation

Validate inline env keys before writing them into a shell or PowerShell command prefix.

- In `writeRemoteCommandPrefix`, skip or reject any env key where `!validEnvName(k)` before `b.WriteString(k)`.
- Apply the same rule at selection time in `allowedEnv`, `allowedEnvFromProfiles`, `allowedEnvWithoutProfileKeys`, or a shared sanitization helper so printed forwarding summaries and command construction agree.
- Add equivalent protection to `writeWindowsRemotePrefix`, which also writes env keys into command text.
- Add a regression test for an inline env map containing a key such as `PROJECT_$(touch /tmp/cbx-env-pwn)#` and assert that the generated remote command does not contain the raw key or command substitution.

## Validation

Validation method: source review plus a local shell-parser smoke test.

Evidence:

- `crabbox/docs/features/env-forwarding.md` documents that invalid env names are dropped during forwarding.
- `internal/cli/repo.go:175-187` accepts `os.Environ()` keys without checking `validEnvName`.
- `internal/cli/run_env_profile.go:51-57` preserves those local inline keys for forwarding.
- `internal/cli/run.go:1430-1440` passes the inline env map into the POSIX remote command builders.
- `internal/cli/ssh.go:1241-1262` writes each env key directly into shell syntax and quotes only the value.
- `internal/cli/run_env_profile.go:485-501` contains the needed `validEnvName` helper, and other env-file/profile paths already use it.
- A bounded local parser check with `bash -lc "CI\$(touch /tmp/crabbox-env-name-pwn-proof)#='v' true"` produced `substitution_executed=yes`, confirming that command substitution runs before bash rejects the malformed assignment.

Counterevidence and limits:

- Exact allowlist entries such as the built-in `CI` and `NODE_OPTIONS` do not match names like `CI$(...)`; a wildcard or broad allowlist is required.
- Configured profile env keys are already validated by `validateProfileConfig`, so the confirmed source path is local process environment selection, not profile-env config injection.
- This was validated statically and with a local shell parser smoke rather than by provisioning a live lease, because the vulnerable behavior is fully visible in generated remote command construction and shell parsing.

steipete:
## Summary

- drop malformed environment names while collecting allowlisted ambient and profile values
- revalidate names before rendering both POSIX and Windows remote command prefixes
- add regressions for the reported command-substitution name and the Windows path

Closes https://github.com/openclaw/crabbox/issues/695

## Verification

- `go test ./internal/cli -run 'Test(RemoteCommandDropsInvalidEnvNames|WindowsNativeRemoteCommandDropsInvalidEnvNames|AllowedEnvDropsInvalidNames)$' -count=1`
- autoreview clean

## Security/config implications

Invalid names outside `[A-Za-z_][A-Za-z0-9_]*` are now dropped consistently. Valid exact and prefix allowlists are unchanged; no secrets or config migrations are involved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,7 @@
 
 ### Fixed
 
+- Dropped invalid allowlisted environment names before rendering remote POSIX or Windows commands, preventing shell metacharacters in ambient names from creating unintended commands. Thanks @coygeek.
 - Pinned Windows Chocolatey image bootstrap to a checksum-verified versioned package before privileged installation. Thanks @TurboTheTurtle.
 - Redacted Daytona API and upload credentials from provider error diagnostics, including reflected authorization headers and token-bearing JSON fields. Thanks @TurboTheTurtle.
 - Recognized current Windows 11 Sandbox host processes during run monitoring and cleanup, preventing false early exits and orphaned sandboxes on 24H2 and newer builds. Thanks @paulcam206.
--- a/internal/cli/repo.go
+++ b/internal/cli/repo.go
@@ -179,6 +179,9 @@ func allowedEnv(allow []string) map[string]string {
 		if !ok {
 			continue
 		}
+		if !validEnvName(k) {
+			continue
+		}
 		if envAllowed(k, allow) {
 			out[k] = v
 		}
--- a/internal/cli/run_env_profile.go
+++ b/internal/cli/run_env_profile.go
@@ -31,7 +31,7 @@ func loadEnvProfiles(paths []string) (map[string]string, error) {
 func allowedEnvFromProfiles(allow []string, profileEnv map[string]string) map[string]string {
 	out := allowedEnv(allow)
 	for key, value := range profileEnv {
-		if envAllowed(key, allow) {
+		if validEnvName(key) && envAllowed(key, allow) {
 			out[key] = value
 		}
 	}
@@ -41,7 +41,7 @@ func allowedEnvFromProfiles(allow []string, profileEnv map[string]string) map[st
 func allowedProfileEnv(allow []string, profileEnv map[string]string) map[string]string {
 	out := map[string]string{}
 	for key, value := range profileEnv {
-		if envAllowed(key, allow) {
+		if validEnvName(key) && envAllowed(key, allow) {
 			out[key] = value
 		}
 	}
--- a/internal/cli/ssh.go
+++ b/internal/cli/ssh.go
@@ -1254,6 +1254,9 @@ func writeRemoteCommandPrefix(b *strings.Builder, workdir string, env map[string
 		b.WriteString("; fi && ")
 	}
 	for k, v := range env {
+		if !validEnvName(k) {
+			continue
+		}
 		b.WriteString(k)
 		b.WriteByte('=')
 		b.WriteString(shellQuote(v))
--- a/internal/cli/ssh_test.go
+++ b/internal/cli/ssh_test.go
@@ -43,6 +43,20 @@ func TestRemoteCommandQuotesWorkdirEnvAndArgs(t *testing.T) {
 	}
 }
 
+func TestRemoteCommandDropsInvalidEnvNames(t *testing.T) {
+	invalid := `PROJECT_$(touch /tmp/cbx-env-pwn)#`
+	got := remoteCommand("/work/repo", map[string]string{
+		"CI":    "1",
+		invalid: "ignored",
+	}, []string{"true"})
+	if !strings.Contains(got, "CI='1'") {
+		t.Fatalf("remoteCommand() missing valid environment variable in %q", got)
+	}
+	if strings.Contains(got, invalid) || strings.Contains(got, "$(touch") {
+		t.Fatalf("remoteCommand() rendered invalid environment name in %q", got)
+	}
+}
+
 func TestRemoteShellCommandRunsScript(t *testing.T) {
 	got := remoteShellCommand("/work/crabbox/cbx_1/repo", map[string]string{"CI": "1"}, "pnpm install && pnpm test")
 	for _, want := range []string{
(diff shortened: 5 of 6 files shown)


steipete:
Landed in https://github.com/openclaw/crabbox/commit/d4df5141.

Verification:

- `go test ./internal/cli -run 'Test(RemoteCommandDropsInvalidEnvNames|WindowsNativeRemoteCommandDropsInvalidEnvNames|AllowedEnvDropsInvalidNames)$' -count=1`
- autoreview clean with no actionable findings
- exact-head CI and CodeQL: https://github.com/openclaw/crabbox/actions/runs/28493943527 and https://github.com/openclaw/crabbox/actions/runs/28493942693

The fix drops malformed allowlisted names at selection time and revalidates them at both POSIX and Windows command-rendering boundaries.

PR #772, opened 2026-07-02
fix: recover ASCII Box snapshot-guarded release

steipete:
## Summary

- preserve the normal ASCII Box `stop` then `delete` release path
- recover only the service's exact recent-snapshot guard by shortening TTL to one second, polling the managed state for up to two minutes, and retrying deletion
- prepare Box CLI configuration once for the release sequence and retain redacted stage diagnostics on recovery failure
- add success, unrelated-error, extend-failure, cancellation, and status-precedence regression coverage
- document the bounded recovery behavior and add an Unreleased changelog entry

## Verification

Current exact head: `d470f3c5b996c13c3b842f7e4e30c173f170d309`.

- `go test -race ./...`
- `go vet ./...`
- `git diff --check`
- focused post-main-sync `go test -race ./internal/providers/asciibox`
- autoreview — clean, no accepted/actionable findings; patch correct at 0.83 confidence
- the only main-sync conflict was the Unreleased changelog list; both user-facing entries were retained

## Production canary

Functional commit `6c31c95014c5d19a70206912e8dfc3f31ea471ef`, an ancestor of the current exact head, was built and run with the current official Box CLI against the production ASCII Box endpoint:

- initial provider inventory: zero Boxes
- disposable sandbox remote command: expected marker emitted
- immediate stop/delete reproduced the service snapshot guard
- Crabbox shortened TTL, observed the managed transition, and retried deletion automatically
- run exit: 0
- manual cleanup path: not used
- final provider inventory: zero Boxes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 
 ### Fixed
 
+- Recovered ASCII Box release when the service temporarily requires a recent snapshot by shortening the sandbox TTL, waiting for the managed stop transition, and retrying deletion.
 - Prevented ASCII Box API credentials from reaching unsafe explicit base URLs by requiring HTTPS except for loopback development endpoints, rejecting ambiguous URL components, and supporting config discovery in the current Box CLI. Thanks @coygeek.
 - Pinned the Google Linux package signing fingerprint, preserved its source-scoped APT keyring across Chrome installation, and failed closed to Chromium when verification fails. Thanks @coygeek.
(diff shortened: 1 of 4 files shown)


steipete:
Maintainer decision: retain the Unreleased changelog entry. Repository policy requires changelog coverage for user-facing fixes and says maintainers or AI, rather than contributor authors, add it. This is a maintainer-authored recovery PR, so the release-owned-file heuristic is non-actionable here. Functional review and exact-head production proof are otherwise accepted.

PR #850, opened 2026-07-04
fix(railway): reject unowned service stops

### LINKED ISSUE #833: [security] Railway stop can stop unclaimed provider resources
# [security] Railway stop can stop unclaimed provider resources

## Summary

The Railway provider exposes a destructive `crabbox stop --provider railway --id <service-id>` path that accepts a Railway service id and stops the latest deployment for that service without proving that the service belongs to a Crabbox lease claim. This is in scope under Crabbox's trusted-operator policy because the supported boundary includes destructive provider actions against resources Crabbox cannot strongly identify as its own; it does not depend on Crabbox being a hostile multi-tenant sandbox.

Railway services are intentionally created out of band, and the provider rejects normal `Run` and `Warmup` lease creation. That design makes an exact ownership/adoption check more important before `Stop` performs the provider mutation. Today the stop path can mutate any deployment visible to the configured Railway token and selected project/environment when the caller supplies its service id.

## Affected Components

- `internal/providers/railway/backend.go:73-80`: `Warmup` is rejected because Railway services must be created outside Crabbox, so this provider does not create a normal Crabbox lease claim before later lifecycle actions.
- `internal/providers/railway/backend.go:383-403`: `Stop` requires only `req.ID`, project id, environment id, and a Railway API client; it reads the latest deployment for the caller-supplied service id and then calls `StopDeployment` without resolving or authorizing a local claim.
- `internal/providers/railway/client.go:395-412`: `StopDeployment` sends Railway's `deploymentStop(id)` GraphQL mutation for the deployment id returned by `LatestDeployment`; the sink has no Crabbox ownership parameter.
- `internal/cli/run.go:3064-3068`: the CLI stop path dispatches `DelegatedRunBackend` providers directly to `Stop`, so Railway does not pass through the `SSHLeaseBackend.Resolve` and `ReleaseLease` ownership flow used by SSH lease providers.
- `internal/providers/agentsandbox/backend.go:539-565`: a delegated-run sibling provider resolves a local claim, locks the lease operation, re-resolves the claim, authorizes provider scope, and deletes only the owned provider claim before releasing the resource.
- `SECURITY.md:34-49`: the maintainer policy lists destructive provider actions against resources Crabbox cannot strongly identify as its own as in scope.
- `docs/security.md:431-436`: the operations guide says provider tags can explain cleanup candidates, but destructive action requires an exact retained coordinator lease binding; tag-only and legacy candidates remain report-only.

## Attack Path

Attacker role: a credentialed Crabbox operator or automation path with access to a configured Railway provider token, project id, and environment id. The path can also happen accidentally through an id mix-up; the report does not rely on unsupported hostile arbitrary tenants sharing one broker.

Prerequisites:

1. The Railway provider is configured with `RAILWAY_API_TOKEN`, `--railway-project`, and `--railway-environment` or their config/env equivalents.
2. The caller knows a Railway service id in that project/environment. Crabbox's Railway `List` path enumerates visible service ids, and operators may also know them from Railway itself.
3. The target service has at least one deployment.

Steps:

1. Run `crabbox stop --provider railway --id <target-service-id>`.
2. The CLI sees that Railway implements `DelegatedRunBackend` and calls `delegated.Stop(ctx, StopRequest{ID: <target-service-id>})` directly.
3. `railwayBackend.Stop` checks only that the service id, project id, environment id, and Railway API client are present.
4. `railwayBackend.Stop` calls `LatestDeployment(ctx, projectID, environmentID, req.ID)`.
5. If Railway returns a deployment id, `railwayBackend.Stop` calls `StopDeployment(ctx, deployment.ID)`.
6. `StopDeployment` sends the `deploymentStop(id)` GraphQL mutation.

The missing control is the absence of a local claim lookup, provider-scope authorization, exact resource binding, or explicit adoption/reclaim step between the caller-controlled service id and the destructive Railway mutation.

## Impact

A caller with the configured Railway provider token can stop a Railway deployment that Crabbox has not proven belongs to the caller's Crabbox lease, run, or local claim. In a shared trusted-team setup, this can disrupt another operator's service or an externally managed Railway service reachable with the same token and project/environment. This crosses the provider-resource mutation boundary described in the maintainer security policy: Crabbox is performing a destructive action against a resource it cannot strongly identify as its own.

Counterevidence considered: Railway's API token, project id, and environment id still bound the blast radius to resources visible through that Railway credential, and Crabbox is not promising hostile multi-tenant isolation behind one shared token. Those facts limit severity, but they do not remove the specific in-scope boundary because the destructive stop action lacks the exact claim/resource ownership proof that sibling providers apply.

### CVSS Assessment

| Metric | v3.1 | v4.0 |
|--------|------|------|
| **Score** | 6.5 / 10.0 | 7.1 / 10.0 |
| **Severity** | Medium | High |
| **Vector** | `CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:U/C:N/I:N/A:H` | `CVSS:4.0/AV:N/AC:L/AT:N/PR:L/UI:N/VC:N/VI:N/VA:H/SC:N/SI:N/SA:N` |
| **Calculator** | [CVSS v3.1 Calculator](https://www.first.org/cvss/calculator/3.1#CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:U/C:N/I:N/A:H) | [CVSS v4.0 Calculator](https://www.first.org/cvss/calculator/4.0#CVSS:4.0/AV:N/AC:L/AT:N/PR:L/UI:N/VC:N/VI:N/VA:H/SC:N/SI:N/SA:N) |

The vectors model a low-complexity, authenticated provider-operation path with no user interaction and high availability impact to the affected Railway deployment. Confidentiality and direct data integrity are not scored because the demonstrated sink stops the latest deployment rather than reading secrets or modifying application data; the configured Railway credential and project/environment scope also limit the blast radius.

## Recommended Remediation

Require exact ownership proof before Railway can stop a deployment. A minimal fix would make `railwayBackend.Stop` reject unclaimed service ids by default, then resolve and authorize a local claim or equivalent provider-resource binding before calling `LatestDeployment` and `StopDeployment`.

Because Railway services are created outside Crabbox, there are a few compatible implementation options:

1. Add an explicit adoption/reclaim flow for Railway service ids that records a local claim bound to provider, project id, environment id, and service id. `Stop` should accept only an exact matching claim.
2. If Crabbox supports Railway redeploy as the lifecycle entry point, record a claim when `redeployService` first operates on a service and require that claim for later `Stop`.
3. If claim-backed Railway stop is not intended, make `Stop` fail closed with a clear message and leave out-of-band stopping to Railway's own UI/API rather than performing an unowned destructive mutation through Crabbox.

Regression coverage should include:

- A Railway provider test proving `Stop` rejects an unclaimed service id before any `LatestDeployment` or `StopDeployment` call.
- A Railway provider test proving `Stop` succeeds only when the local claim exactly matches provider name, Railway project id, environment id, and service id.
- A CLI stop test, or focused provider-backend test, proving delegated-run Railway stop no longer bypasses ownership validation.

## Validation

Validation method: static source review against the latest verification checkout `62831c8972de420505096693ee290ac70b79f062`. Live Railway reproduction was not attempted because it would require real provider credentials and a real service; the static source/control/sink tuple is complete and sufficient to validate the missing ownership check.

Evidence:

- `SECURITY.md:34-49` makes destructive provider actions against resources Crabbox cannot strongly identify as its own an in-scope supported boundary.
- `docs/security.md:431-436` documents the stronger cleanup rule that destructive provider mutations require exact retained lease bindings, not only provider-visible tags or identifiers.
- `internal/providers/railway/backend.go:73-80` confirms Railway does not create services through Crabbox warmup and therefore has no normal warmup-created local claim.
- `internal/providers/railway/backend.go:383-403` confirms `Stop` uses the caller-supplied service id to find the latest deployment and then stops that deployment without claim resolution.
- `internal/providers/railway/client.go:395-412` confirms the provider sink is Railway's `deploymentStop(id)` mutation.
- `internal/cli/run.go:3064-3068` confirms delegated-run providers bypass the SSH lease resolve/release path and rely on their own `Stop` implementation for validation.
- `internal/providers/agentsandbox/backend.go:539-565` shows the expected delegated-run ownership pattern: resolve a local claim, authorize provider scope, and delete only the owned claim/resource.
- Prior public Crabbox security reports treat unsafe provider resource mutation as reportable when a caller-controlled provider id reaches a destructive provider call without exact ownership proof; the Railway stop path has the same source/control/sink shape.
- I did not find a prior public issue covering Railway stop ownership validation; this path is also distinct from trusted repository configuration choosing a local helper command because the destructive provider call is implemented by Crabbox's Railway provider.

Counterevidence and limits:

- `internal/providers/railway/backend.go:83-95` rejects arbitrary `Run` commands, so this report is not about arbitrary code execution on Railway.
- `internal/providers/railway/backend.go:427-459` rejects many normal lease/run options, including `--keep`, `--reclaim`, sync, shell mode, and env summary. Those checks reduce unrelated risk but do not validate ownership before `Stop`.
- `internal/providers/railway/bridge.go:65-90` also treats the service id as the Railway lease id for bridge URL lookup, but that path is read-like and not the destructive sink in this report.
- The Railway API token and project/environment config limit the affected resources to the provider account scope visible to that credential.

Verdict: accepted. The source establishes a reachable destructive provider mutation from a caller-controlled service id to Railway's stop mutation with no Crabbox claim or exact resource ownership check.

steipete:
## Summary

- make Railway `stop` fail closed before authentication, inventory lookup, or mutation because out-of-band services have no verifiable Crabbox ownership claim
- preserve Railway `status` and `list` as read-only inspection paths; direct operators to Railway for destructive lifecycle actions
- replace the prior mutation test with a regression test proving an unclaimed service ID causes zero Railway API calls
- update the Railway guide, generated provider matrix, metadata, and changelog to match the supported boundary

This follows the lifecycle policy in `VISION.md`: destructive operations require exact provider/resource/claim ownership, while legacy or external adoption must be explicit and conflict-safe. Railway currently has no adoption contract, so a raw service ID cannot authorize a stop.

Closes https://github.com/openclaw/crabbox/issues/833

## Verification

Exact candidate: `c718a781f01b85d6fdd921cb2c140958e798a06a`

- `go test -race ./internal/providers/railway`
- `go vet ./internal/providers/railway`
- `go test -race ./...`
- `scripts/check-docs.sh`
- `node scripts/generate-provider-matrix.mjs`
- `git diff --check`
- autoreview clean, confidence 0.86

## Live proof

Built the real CLI with `go build -trimpath -o bin/crabbox ./cmd/crabbox`, started a listening local GraphQL spy, then invoked `crabbox stop --provider railway --id svc-unclaimed` with complete project/environment/API configuration and a test-only token. The CLI returned the ownership rejection and the spy recorded exactly zero HTTP requests. This proves the changed end-to-end behavior without risking a real Railway service.

## Compatibility

This intentionally removes Crabbox's unsafe raw-ID Railway stop behavior. Existing services remain inspectable through `status` and `list`; stop them through Railway until Crabbox has an explicit adoption and exact-claim contract.


vincentkoc:
Closing as superseded by https://github.com/openclaw/crabbox/pull/851, which landed in https://github.com/openclaw/crabbox/commit/b9434f7636b847c5595c343e4e2ae8cc772c01b1.

Reason: this PR took the older fail-closed-only shape for Railway stop. The landed fix keeps the same ownership boundary while supporting exact local claims and explicit `stop --reclaim` one-deployment adoption, so carrying this branch forward would remove intended supported behavior and reintroduce release-owned changelog churn.

Issue https://github.com/openclaw/crabbox/issues/833 is already closed by the landed PR.

PR #866, opened 2026-07-04
fix(gcp): revalidate release ownership before delete

### LINKED ISSUE #861: [security] GCP release deletes instances by CloudID without ownership revalidation
# [security] GCP release deletes instances by CloudID without ownership revalidation

## Summary

The GCP direct-provider release path deletes the instance named by the stored lease `CloudID` without first re-fetching the live instance and verifying that it is still the exact Crabbox-owned resource for the lease being released.

This is in scope for Crabbox's documented trusted-operator model because it is not about hostile tenants sharing a broker or an operator intentionally deleting their own VM. The supported boundary is Crabbox performing a destructive provider action only after it can strongly identify the target as its own. `SECURITY.md` explicitly lists destructive provider actions against resources Crabbox cannot strongly identify as its own as an in-scope report family, and `docs/security.md` says provider names and weak metadata must not become destructive ownership proof.

Current GCP resolve/list/cleanup paths already use a canonical ownership predicate, but `ReleaseLease` bypasses that predicate and sends the stored instance name directly to the GCP delete API. If stale local state, stale lease state, or provider-side label/name drift causes the release target to point at a non-Crabbox instance in the configured project and zone, a normal release can delete that instance instead of failing closed.

## Affected Components

- `internal/providers/gcp/backend.go:193-210`: `ReleaseLease` optionally switches zones from the stored lease labels, then calls `client.DeleteServer(ctx, req.Lease.Server.CloudID)` and removes the lease claim without checking the live instance's canonical Crabbox labels or matching lease ID.
- `internal/providers/gcp/backend.go:125-162`: `Resolve` uses `isCrabboxGCPLease` before returning a direct instance by name, showing that GCP already has an ownership predicate for read/resolve decisions.
- `internal/providers/gcp/backend.go:238-288`: cleanup starts from the canonical filtered inventory, skips non-canonical servers, and only then deletes expired servers, so the safer ownership pattern exists on another destructive GCP path.
- `internal/cli/gcp.go:389-417`: `IsCanonicalGCPServer` requires a canonical lease ID, non-empty slug, generated instance name, and the expected `crabbox=true`, `created_by=crabbox`, and `provider=gcp` labels, while `DeleteServer` itself deletes only by project, zone, and instance name.
- `internal/cli/provider_labels.go:10-24`: newly created direct leases receive the label set that `IsCanonicalGCPServer` later expects, including `crabbox`, `created_by`, `lease`, `slug`, and `provider`.
- `internal/cli/run.go:2731-2738`: the normal direct-provider release path calls the provider backend's `ReleaseLease`, making the GCP backend the last place to refuse an unsafe destructive target before deletion.

## Attack Path

Required actor and state:

1. A legitimate Crabbox operator or release flow has authority to release a GCP lease.
2. The stored `LeaseTarget.Server.CloudID` names an instance in the configured GCP project and zone, or in the zone selected from the lease's stored `zone` label.
3. That live instance is not currently the canonical Crabbox-owned instance for the lease being released. This can happen through stale local state, stale lease state, claim corruption, or provider-side label/name drift after the lease target was originally resolved.

Trigger:

1. The operator runs a normal release action such as `crabbox stop --id <lease-id-or-slug>`, or another direct release flow reaches `releaseBackendLease`.
2. `releaseBackendLease` calls the GCP backend's `ReleaseLease` with the stored `LeaseTarget`.
3. `gcpLeaseBackend.ReleaseLease` creates a GCP client, optionally trusts the stored `zone` label to select the zone, and calls `DeleteServer` with the stored `CloudID`.
4. `GCPClient.DeleteServer` issues `instances.Delete` for that project, zone, and instance name. It does not fetch the instance first and does not verify that the live instance has canonical Crabbox ownership labels or the expected lease ID.

The supported boundary crossed is Crabbox's destructive provider-resource safety boundary. The path does not require treating a leased machine as a sandbox or treating trusted operators as mutually adversarial tenants.

## Impact

A stale or drifted GCP lease target can cause Crabbox to delete a GCP Compute Engine instance that it cannot prove belongs to the lease being released. The direct consequence is complete availability loss for the targeted VM. In a shared project, the blast radius can include manually managed or production-like instances if their name is carried into stale Crabbox release state.

The risk is bounded by the need for valid Crabbox release authority and an uncommon stale/drifted target condition. It remains security-relevant because Crabbox documents exact ownership as the requirement for destructive provider actions, and the current GCP release path does not perform that check at the point of deletion.

### CVSS Assessment

| Metric | v3.1 | v4.0 |
|--------|------|------|
| **Score** | 5.3 / 10.0 | 4.1 / 10.0 |
| **Vector** | CVSS:3.1/AV:L/AC:H/PR:L/UI:R/S:C/C:N/I:N/A:H | CVSS:4.0/AV:L/AC:H/AT:P/PR:L/UI:A/VC:N/VI:N/VA:H/SC:N/SI:N/SA:N |
| **Calculator** | [CVSS v3.1 Calculator](https://www.first.org/cvss/calculator/3.1#CVSS:3.1/AV:L/AC:H/PR:L/UI:R/S:C/C:N/I:N/A:H) | [CVSS v4.0 Calculator](https://www.first.org/cvss/calculator/4.0#CVSS:4.0/AV:L/AC:H/AT:P/PR:L/UI:A/VC:N/VI:N/VA:H/SC:N/SI:N/SA:N) |

The scoring reflects a local/trusted release path with valid Crabbox release privilege, user action, and a high-complexity stale or drifted target precondition. Availability impact is high for the deleted GCP instance, while confidentiality and integrity are not directly affected by this path.

## Recommended Remediation

Fail closed in `gcpLeaseBackend.ReleaseLease` before calling `DeleteServer` unless the release target can be re-proven as the exact Crabbox-owned instance for the lease:

1. Re-fetch the live GCP instance by the stored `CloudID` in the selected zone.
2. Require `isCrabboxGCPLease(liveServer)` to pass.
3. Require the live server's `lease` label to match `req.Lease.LeaseID`.
4. Reject missing or mismatched live `CloudID`, generated instance name, `slug`, `provider`, or project/zone identity before deletion.
5. Consider adding an exact local claim binding check, consistent with providers that now require unchanged local claims before destructive actions.
6. Only after those checks pass, call `DeleteServer`, remove the lease claim, and report release success.

Add regression coverage with a fake GCP client proving that `ReleaseLease` refuses to call `DeleteServer` when the stored target points at a non-canonical or wrong-lease instance. A positive test should prove that canonical labels plus a matching lease ID still release normally.

## Validation

Validated against current `main` at commit `f12f3b6a88f6eb269806afb2656bdb9d945af5c0`.

Source evidence:

- `internal/providers/gcp/backend.go:193-210` deletes the stored `CloudID` directly and removes the lease claim afterward.
- `internal/providers/gcp/backend.go:125-162` and `internal/cli/gcp.go:389-402` show that the intended GCP ownership predicate exists and is already used for GCP resolve/list behavior.
- `internal/providers/gcp/backend.go:238-288` shows GCP cleanup deletes only after starting from canonical filtered inventory.
- `internal/cli/gcp.go:404-417` confirms the underlying delete method performs no ownership validation; it issues the GCP delete request for the configured project, zone, and instance name.
- `internal/providers/aws/backend.go:213-225`, `internal/providers/azure/backend.go:178-187`, and `internal/providers/hetzner/backend.go:203-228` provide same-repository contrast where release paths refuse or gate deletion when canonical provider ownership does not match.

Counterevidence considered: normal freshly resolved GCP leases are expected to be canonical because GCP resolve/list paths filter through `IsCanonicalGCPServer`, and cleanup uses that filtered inventory. That does not cover release-time stale targets or provider-side drift because `ReleaseLease` does not re-fetch or revalidate the live GCP instance immediately before calling the destructive API. A focused package test run was attempted, but the local sandbox could not create Go module-cache lock files under the existing module cache; this did not change the static source conclusion.

vincentkoc:
## Summary
- re-fetch the live GCP instance before direct release deletion
- refuse missing, non-canonical, mismatched CloudID, or wrong-lease live targets before calling DeleteServer
- preserve refused local claims, while removing stale claims when the live GCP instance is already gone

Fixes https://github.com/openclaw/crabbox/issues/861

## Verification
- git diff --check
- go test -count=1 ./internal/providers/gcp
- go test -count=1 ./internal/cli -run 'Test.*GCP|TestIsCanonicalGCPServer'\n- go test ./...\n\n## Credentials\n- no live GCP credentials used; regression covered with fake GCP client ownership tests

### DIFF
--- a/internal/providers/gcp/backend.go
+++ b/internal/providers/gcp/backend.go
@@ -195,6 +195,10 @@ func (b *gcpLeaseBackend) ReleaseLease(ctx context.Context, req ReleaseLeaseRequ
 	if err != nil {
 		return err
 	}
+	cloudID := strings.TrimSpace(req.Lease.Server.CloudID)
+	if cloudID == "" {
+		return exit(4, "refusing to delete gcp lease=%s: missing cloud id", req.Lease.LeaseID)
+	}
 	if zone := req.Lease.Server.Labels["zone"]; zone != "" {
 		cfg := b.Cfg
 		cfg.GCPZone = zone
@@ -203,7 +207,24 @@ func (b *gcpLeaseBackend) ReleaseLease(ctx context.Context, req ReleaseLeaseRequ
 			return err
 		}
 	}
-	if err := client.DeleteServer(ctx, req.Lease.Server.CloudID); err != nil {
+	live, err := client.GetServer(ctx, cloudID)
+	if err != nil {
+		if core.IsGCPNotFound(err) {
+			removeLeaseClaim(req.Lease.LeaseID)
+			return nil
+		}
+		return err
+	}
+	if strings.TrimSpace(live.CloudID) != cloudID {
+		return exit(4, "refusing to delete gcp lease=%s: live instance cloud id %q does not match stored cloud id %q", req.Lease.LeaseID, live.CloudID, cloudID)
+	}
+	if !isCrabboxGCPLease(live) {
+		return exit(4, "refusing to delete gcp instance %q for lease=%s: live instance is not canonical Crabbox-owned", cloudID, req.Lease.LeaseID)
+	}
+	if liveLeaseID := strings.TrimSpace(live.Labels["lease"]); liveLeaseID != req.Lease.LeaseID {
+		return exit(4, "refusing to delete gcp instance %q for lease=%s: live instance belongs to lease=%s", cloudID, req.Lease.LeaseID, blank(liveLeaseID, "-"))
+	}
+	if err := client.DeleteServer(ctx, cloudID); err != nil {
 		return err
 	}
 	removeLeaseClaim(req.Lease.LeaseID)
--- a/internal/providers/gcp/backend_doctor_test.go
+++ b/internal/providers/gcp/backend_doctor_test.go
@@ -17,6 +17,8 @@ type fakeGCPDoctorClient struct {
 	deleted   []string
 	mutated   bool
 	servers   []Server
+	get       map[string]Server
+	getErr    error
 	created   Server
 	createCfg Config
 	createErr error
@@ -51,8 +53,16 @@ func (c *fakeGCPDoctorClient) WaitForServerIP(context.Context, string) (Server, 
 	return c.created, nil
 }
 
-func (c *fakeGCPDoctorClient) GetServer(context.Context, string) (Server, error) {
-	return Server{}, nil
+func (c *fakeGCPDoctorClient) GetServer(_ context.Context, name string) (Server, error) {
+	if c.getErr != nil {
+		return Server{}, c.getErr
+	}
+	if c.get != nil {
+		if server, ok := c.get[name]; ok {
+			return server, nil
+		}
+	}
+	return Server{}, errors.New("gcp server not found: " + name)
 }
 
 func (c *fakeGCPDoctorClient) DeleteServer(_ context.Context, name string) error {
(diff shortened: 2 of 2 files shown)


vincentkoc:
Maintainer acceptance: this PR intentionally changes direct GCP release to fail closed for stale, noncanonical, or wrong-lease live instances and preserve the local claim rather than deleting by stored CloudID. That compatibility tradeoff is accepted for the provider-resource safety boundary; any self-service recovery path should be separate and must not weaken release ownership checks.

PR #894, opened 2026-07-05
Fix Azure lease readiness cleanup

### LINKED ISSUE #862: Azure leases churn: coordinator recreates VM every ~16s, never reaches ready, orphans accumulate
## Summary

Azure lease provisioning is effectively unusable on the deployed coordinator. Two related failures:

1. **Readiness churn (root cause).** The coordinator creates the Azure VM successfully, but the lease never reaches `ready`. It then tears the VM down and recreates it every ~16s in an infinite loop until the CLI times out (default 10m). No lease is ever returned; some attempts surface as `POST /v1/leases: http 500: error code: 1101` (Worker exception), others as a silent 10m timeout.
2. **Orphan sweep does not reap.** Each churn round can leave a stranded VM behind whose lease record the coordinator no longer knows about (`GET /v1/leases/<slug>` → 404). These accumulate and slowly exhaust the region's per-family vCPU quota, which then blocks even a healthy attempt.

Net effect for users: `--provider azure` hangs for 10 minutes and falls back / fails. AWS and Blacksmith Testbox are unaffected.

## Evidence (2026-07-04, `crabbox-leases` RG, eastus2, OpenClaw Foundation Subscription)

Coordinator create loop for one lease (`az monitor activity-log`, VM writes for a single slug):

```
17:42:24 Started/Succeeded  crabbox-coral-hermit-d16b1c71
17:42:40 Started/Succeeded  crabbox-coral-hermit-d16b1c71
17:42:58 Started/Succeeded  crabbox-coral-hermit-d16b1c71
17:43:14 Started/Succeeded  crabbox-coral-hermit-d16b1c71
17:43:30 Started/Succeeded  crabbox-coral-hermit-d16b1c71
17:43:47 Started/Succeeded  crabbox-coral-hermit-d16b1c71
```

Interleaved `Delete Network Interface` / `Delete Public Ip Address` return `Bad Request (400)` — the teardown side of the same loop. Meanwhile the CLI only ever prints `waiting for coordinator lease ... elapsed=Nm ... timeout=10m0s`.

Stranded orphans found (all `keep:false`, no live lease record): `crabbox-violet-shrimp` (D32ads_v6, created 2026-07-01), `crabbox-amber-shrimp` (D32ds_v6, 2026-07-01), `crabbox-amber-crayfish` (D32ds_v6, 2026-07-04), `crabbox-crimson-prawn` (D32ds_v6, 2026-07-04, coordinator returned 404 for its slug).

Quota pin caused by the orphans: eastus2 `standard Dadv6 Family vCPUs` was `48/65`. The default lease type `Standard_D32ads_v6` needs 32 vCPUs (48+32=80 > 65), so the default azure lane could not provision at all until the orphans were cleared.

## Ruled out

- **Auth / service principal secret** — app `crabbox-worker-azure-20260527` client secret is valid until 2027-05-27. VMs create successfully, so token acquisition works.
- **NSG / SSH reachability** — `crabbox-nsg-eastus2` allows TCP 22 from `0.0.0.0/0` at priority 100.
- **Region / family quota** — after reaping orphans, `Dadv6` is `16/65` and `Ddsv6` is `0/350`; churn still reproduces, so quota is a secondary amplifier, not the trigger.

The remaining suspect is the coordinator's post-create readiness gate (SSH auth / bootstrap / custom-data on the `Canonical:ubuntu-26_04-lts:server:latest` image). The deployed coordinator Worker version is `e8219f79` (2026-05-31); azure worked until ~2026-07-03, which suggests an image-`latest` drift or a readiness-probe regression rather than a code change.

## Asks

1. Fix the readiness churn: on repeated readiness failure, stop recreating and surface the underlying SSH/bootstrap error to the CLI instead of looping to timeout.
2. Fix the orphan sweep so `keep:false` VMs with no live lease record are reaped (these are the ones that pinned the quota).
3. Consider pinning the Azure Linux image to an explicit published version instead of `:latest`.

## Mitigation already applied

- `openclaw/openclaw` `.crabbox.yaml` default moved from `azure` to `blacksmith-testbox` so agents are unblocked.
- Reaped the four stranded VMs above (+ their disks/NICs/public IPs), leaving only the `keep:true` intentional hosts. eastus2 family quota is now clear.

TurboTheTurtle:
Closes #862.

Restores Azure's default worker image to Ubuntu 24.04 LTS while keeping the 26.04 image family available as an explicit selector. Azure orphan sweep deletion remains claim-backed; stale tag-only resources stay report-only.

### DIFF
--- a/worker/src/azure.ts
+++ b/worker/src/azure.ts
@@ -28,10 +28,12 @@ const MIN_LRO_POLL_INTERVAL_MS = 15_000;
 const DEFAULT_AZURE_NETWORK_LRO_TIMEOUT_MS = 60_000;
 const DEFAULT_AZURE_VM_CREATE_TIMEOUT_MS = 180_000;
 const DEFAULT_AZURE_SPOT_FALLBACK_MS = 120_000;
-const DEFAULT_AZURE_LINUX_IMAGE = "Canonical:ubuntu-26_04-lts:server:latest";
-const DEFAULT_AZURE_LINUX_ARM64_IMAGE = "Canonical:ubuntu-26_04-lts:server-arm64:latest";
 const AZURE_NOBLE_LINUX_IMAGE = "Canonical:ubuntu-24_04-lts:server:latest";
 const AZURE_NOBLE_LINUX_ARM64_IMAGE = "Canonical:ubuntu-24_04-lts:server-arm64:latest";
+const DEFAULT_AZURE_LINUX_IMAGE = AZURE_NOBLE_LINUX_IMAGE;
+const DEFAULT_AZURE_LINUX_ARM64_IMAGE = AZURE_NOBLE_LINUX_ARM64_IMAGE;
+const AZURE_ORACULAR_LINUX_IMAGE = "Canonical:ubuntu-26_04-lts:server:latest";
+const AZURE_ORACULAR_LINUX_ARM64_IMAGE = "Canonical:ubuntu-26_04-lts:server-arm64:latest";
 const LEGACY_AZURE_JAMMY_IMAGE = "Canonical:0001-com-ubuntu-server-jammy:22_04-lts-gen2:latest";
 const LEGACY_AZURE_NOBLE_GEN2_IMAGE =
   "Canonical:0001-com-ubuntu-server-noble:24_04-lts-gen2:latest";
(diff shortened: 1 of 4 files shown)


steipete:
Thanks @TurboTheTurtle for investigating this.

I am closing this candidate because there is no safe, currently justified subset to land:

- The linked https://github.com/openclaw/crabbox/issues/862 no longer reproduces on current `main`. Its deployed Azure proof reached ready, executed a real SSH command, and destroyed the lease with zero provider residue.
- The current Azure 26.04 default also completed a separate exact create/use/destroy live proof, so the proposed 24.04 rollback is not needed by the available evidence.
- Deleting a claimless VM from mutable provider tags alone conflicts with Crabbox's verified-ownership boundary. A copied or stale `lease` tag cannot authorize destructive cleanup after the coordinator claim is gone; those resources must remain report-only.

The unit-test changes accurately expose the proposed behavior, but the behavior itself is unsafe. If a fresh Azure readiness failure becomes reproducible, please open a new issue with the region, SKU, redacted provider error, and lease diagnostics so we can repair the actual failing path without weakening ownership checks.

PR #897, opened 2026-07-05
fix(worker): revoke incomplete legacy Code viewers

### LINKED ISSUE #895: [security] Legacy Code viewer sockets survive share revocation when restored without a complete principal
# [security] Legacy Code viewer sockets survive share revocation when restored without a complete principal

## Summary

Crabbox treats coordinator ownership and sharing controls as product security boundaries: `SECURITY.md` says these controls protect against unauthorized access and cross-owner operations, and lists access to another owner's resources contrary to documented authorization as in scope. The Code viewer bridge revocation path does not fully enforce that boundary for legacy or restored Code viewer WebSocket attachments that do not carry a complete `{owner, org, admin}` principal.

When a lease owner removes a user's share access, `putLeaseShareAndRevokeBridges` calls `revokeUnauthorizedLeaseBridges`. The WebVNC path closes viewers that no longer satisfy `leaseViewerAuthorized`, but the Code viewer path skips any attachment whose principal is incomplete before it reaches that authorization check. Current Code viewer sockets are created with a complete principal, so the issue is limited to older or hibernated/restored attachments that the current deserializer still accepts. For those sockets, a user who had prior `use` access can keep the Code viewer WebSocket open after the owner revokes that share.

## Affected Components

- `SECURITY.md:22-24`: Maintainer policy identifies coordinator authentication, ownership, and sharing controls as supported product boundaries.
- `SECURITY.md:36-40`: Maintainer policy lists access to another owner's resources contrary to documented authorization as an in-scope security issue.
- `worker/src/fleet.ts:5137-5145`: Share updates persist the new lease share and invoke `revokeUnauthorizedLeaseBridges` when access shrinks.
- `worker/src/fleet.ts:5160-5171`: The Code viewer revocation loop skips attachments when `completeBridgePrincipal(attachment)` is false, so incomplete-principal viewers are not passed to `leaseViewerAuthorized` and are not closed.
- `worker/src/fleet.ts:5201-5208`: `leaseViewerAuthorized` itself would fail closed for incomplete principals, which shows the skip in the caller is the broken control.
- `worker/src/fleet.ts:14443-14465`: Restored `code-viewer` attachments are accepted even when they come from old serialized state with no auth binding, and `org`/`admin` remain optional after deserialization.
- `worker/src/fleet.ts:14498-14521`: `validOptionalBridgePrincipal` permits the absent-principal state, while `completeBridgePrincipal` requires all of `owner`, `org`, and `admin`.
- `worker/src/fleet.ts:8153-8163`: New Code viewer WebSockets are created with `owner`, `org`, and `admin`, limiting this issue to legacy or restored sockets rather than ordinary fresh connections on current code.

## Attack Path

Attacker role: a non-admin user who previously had `use` access to a lease's Code viewer.

Prerequisites:

1. The lease owner previously shared Code access with the attacker.
2. The attacker's Code viewer WebSocket is a legacy or restored attachment that lacks the complete `{owner, org, admin}` principal. The current deserializer accepts this state for backward compatibility.
3. The attacker keeps the existing Code viewer WebSocket open while the lease owner removes their share.

Trigger:

1. The lease owner updates or deletes the share entry that granted the attacker access.
2. `putLeaseShareAndRevokeBridges` detects the share access shrank and calls `revokeUnauthorizedLeaseBridges`.
3. The Code viewer revocation loop reads the attacker's restored socket attachment.
4. Because the attachment lacks a complete principal, the loop reaches the `continue` branch before calling `leaseViewerAuthorized`.
5. The socket remains in `codeViewers` and the Code bridge continues forwarding traffic for that viewer.

This is not a hostile multi-tenant isolation claim and it does not depend on arbitrary command execution on the attacker's own lease. The crossed boundary is narrower: a lease owner's explicit share revocation is not enforced for a class of long-lived or restored Code viewer sockets that Crabbox still accepts.

## Impact

A user whose `use` share was revoked can retain active Code viewer access for the lifetime of the legacy/restored WebSocket, until another event closes the socket, the portal session is revoked, the Code bridge disconnects, or the lease expires. During that window the user can continue interacting with the Code server surface they were supposed to lose, exposing or changing workspace state available through that bridge.

The impact is bounded because fresh Code viewer WebSockets on current code carry complete principal data, and because the attacker must already have had an active shared Code viewer session. The issue is still reportable because share revocation is a supported coordinator authorization boundary, and Crabbox's own hardening guidance says long-lived Code sessions are revalidated while data flows.

### CVSS Assessment

| Metric | v3.1 | v4.0 |
|--------|------|------|
| **Score** | 4.2 / 10.0 | 2.3 / 10.0 |
| **Severity** | Medium | Low |
| **Vector** | `CVSS:3.1/AV:N/AC:H/PR:L/UI:N/S:U/C:L/I:L/A:N` | `CVSS:4.0/AV:N/AC:H/AT:P/PR:L/UI:N/VC:L/VI:L/VA:N/SC:N/SI:N/SA:N` |
| **Calculator** | [CVSS v3.1 Calculator](https://www.first.org/cvss/calculator/3.1#CVSS:3.1/AV:N/AC:H/PR:L/UI:N/S:U/C:L/I:L/A:N) | [CVSS v4.0 Calculator](https://www.first.org/cvss/calculator/4.0#CVSS:4.0/AV:N/AC:H/AT:P/PR:L/UI:N/VC:L/VI:L/VA:N/SC:N/SI:N/SA:N) |

The scores reflect a network-reachable coordinator path with low privileges and continued access to Code viewer confidentiality and integrity after authorization is revoked. Attack complexity and attack requirements are elevated because the vulnerable state requires a pre-existing legacy or restored incomplete-principal Code viewer attachment; new Code viewer sockets in current code carry complete principal fields.

Treat Code viewer attachments without a complete, verifiable principal as unauthorized during share-revocation and restored-socket reconciliation. The revocation path should close incomplete-principal Code viewers for the affected lease instead of skipping them, matching the fail-closed behavior already used by `leaseViewerAuthorized`.

Preserve backward-compatible deserialization only as needed to identify and clean up old sockets. Add regression coverage for an active `code-viewer` attachment with no `owner/org/admin` that remains open before a share update, then verify that a share removal closes it with the same revocation behavior as a complete-principal revoked viewer. Also keep coverage showing retained users, the lease owner, and current valid admin grants remain connected.

## Validation

Validation method: source review plus focused existing target test.

Evidence:

- `worker/src/fleet.ts:5160-5171` contains the vulnerable branch: incomplete-principal Code viewer attachments are skipped before authorization is checked or the socket is closed.
- `worker/src/fleet.ts:5201-5208` shows the intended authorization helper fails closed for incomplete principals, so the caller-side skip bypasses the stricter helper behavior.
- `worker/src/fleet.ts:14443-14465` and `worker/src/fleet.ts:14498-14521` show restored Code viewer attachments without complete principal data are still accepted as structurally valid.
- `worker/src/fleet.ts:8153-8163` shows current fresh Code viewer WebSockets include `owner`, `org`, and `admin`, which limits the reachability to legacy or restored sockets.
- `worker/test/fleet.test.ts:9987-10001` constructs a `code-legacy` viewer with `auth: "github"` and no complete principal, then `worker/test/fleet.test.ts:10091-10101` asserts the revoked complete-principal Code viewer closes while `code-legacy` remains connected after the share removal.
- Focused command run against the isolated target checkout: `npm test --prefix worker -- fleet.test.ts -t "revokes only unauthorized live viewers after an API share removal"` passed with 1 test passed and 368 skipped, confirming the current suite encodes this behavior.

Counterevidence and scope limits:

- Fresh Code viewer WebSockets on the current code path include complete principal fields, so this is not a general bypass for newly created sockets.
- Portal-session logout still closes matching GitHub Code viewers by portal session hash. The issue is specifically share-revocation revalidation for incomplete-principal restored Code viewer attachments.

steipete:
Closes https://github.com/openclaw/crabbox/issues/895

## Summary

- fail closed restored Code viewer sockets that lack a complete owner/org/admin principal
- apply the same rule when lease sharing access shrinks, while retaining authorized owner, shared-user, and current-admin viewers
- preserve the more specific `portal session ended` reason when both the principal and portal-session checks fail
- document the compatibility boundary and add the Unreleased changelog entry

## Verification

- `npm test --prefix worker` (26 files, 796 tests)
- `go test -race ./...`
- `go vet ./...`
- `.agents/skills/autoreview/scripts/autoreview --mode local --stream-engine-output` (clean, confidence 0.86)

## Live proof

Deployed the candidate to a disposable Cloudflare Worker plus a disposable service-bound Code-origin front. Using independently signed owner and shared-user sessions, a real isolated Code WebSocket attached to the Durable Object, received the backend `ws_open`, and was then revoked through the owner share API. The viewer and backend notification both closed with code `1008` and reason `lease access revoked`; replay with the old Code session returned `404`.

Both disposable Workers were deleted after proof, both public URLs returned `404`, and the registered external proof lease was released. No production route, credential, or state changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ### Fixed
 
+- Closed restored legacy Code viewer sessions that lack a complete organization-bound principal during restore and after lease share revocation. Thanks @coygeek.
 - Required a canonical public origin for GitHub OAuth and bound callbacks to the initiating origin before exchanging codes or issuing sessions. Thanks @coygeek.
 - Redacted configured credentials, authorization headers, signed URLs, URL userinfo, and secret-bearing JSON fields before coordinator provider diagnostics are stored or returned. Thanks @coygeek.
 - Redacted broker URL userinfo, queries, and fragments from `login`, `whoami`, and `doctor` text and JSON output. Thanks @coygeek.
--- a/docs/security.md
+++ b/docs/security.md
@@ -162,7 +162,9 @@ Changing any configured admin source revokes older active, restored, ticketed,
 or durable-session grants on the next authenticated request or scheduled
 reconciliation; active senders and recipients are also revalidated while data
 flows. Legacy admin attachments without a verifiable source and version fail
-closed after upgrade.
+closed after upgrade. Restored Code viewer sessions without a complete
+organization-bound principal also fail closed during restore and whenever lease
+sharing access shrinks.
 Shared-operator requests do **not** trust caller-supplied `X-Crabbox-Owner` /
 `X-Crabbox-Org` headers — pin that automation's identity with
 `CRABBOX_SHARED_OWNER` (and `CRABBOX_DEFAULT_ORG`), or prefer per-user signed
--- a/worker/src/fleet.ts
+++ b/worker/src/fleet.ts
@@ -1200,12 +1200,10 @@ export class FleetCoordinator {
         continue;
       }
       if (
-        (attachment.kind === "webvnc-viewer" ||
-          (attachment.kind === "code-viewer" && completeBridgePrincipal(attachment))) &&
+        (attachment.kind === "webvnc-viewer" || attachment.kind === "code-viewer") &&
         !this.leaseViewerAuthorized(lease, attachment)
       ) {
         revokedViewers.set(socket, "lease access revoked");
-        continue;
       }
       if (
         (attachment.kind === "egress-host" || attachment.kind === "egress-client") &&
@@ -5162,7 +5160,6 @@ export class FleetCoordinator {
       if (
         attachment?.kind !== "code-viewer" ||
         attachment.leaseID !== lease.id ||
-        !completeBridgePrincipal(attachment) ||
         this.leaseViewerAuthorized(lease, withCurrentAdminGrant(attachment, adminGrants))
       ) {
         continue;
--- a/worker/test/fleet.test.ts
+++ b/worker/test/fleet.test.ts
@@ -725,6 +725,13 @@ describe("runtime adapter relay", () => {
       org: "example-org",
       admin: false,
     });
+    const legacyCodeViewer = new FakeWebSocket({
+      kind: "code-viewer",
+      leaseID: lease.id,
+      id: "code-legacy",
+      auth: "github",
+      portalSessionHash: "a".repeat(64),
+    });
     const revokedWebAgent = new FakeWebSocket({
       kind: "webvnc-agent",
       leaseID: lease.id,
@@ -777,6 +784,7 @@ describe("runtime adapter relay", () => {
             codeAgent,
             revokedCodeViewer,
             ownerCodeViewer,
+            legacyCodeViewer,
             revokedWebAgent,
             revokedWebViewer,
             retainedWebAgent,
@@ -792,6 +800,8 @@ describe("runtime adapter relay", () => {
     expect(revokedCodeViewer.closeCode).toBe(1008);
     expect(revokedCodeViewer.closeReason).toBe("lease access revoked");
     expect(ownerCodeViewer.closeCode).toBeUndefined();
+    expect(legacyCodeViewer.closeCode).toBe(1008);
+    expect(legacyCodeViewer.closeReason).toBe("lease access revoked");
     expect(revokedWebViewer.closeCode).toBe(1008);
     expect(revokedWebViewer.closeReason).toBe("lease access revoked");
     expect(revokedWebAgent.closeCode).toBe(1011);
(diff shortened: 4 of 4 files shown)


steipete:
Maintainer decision: approved as an intentional fail-closed compatibility boundary.

An incomplete `{owner, org, admin}` principal cannot be re-authorized against the current lease share. Preserving such a restored session would keep an authorization bypass alive after explicit revocation, so legacy Code viewer continuity does not outweigh the coordinator sharing boundary. The deployed proof confirms current complete-principal sessions retain normal behavior and a revoked shared user closes deterministically.

The `CHANGELOG.md` entry is retained. Repository policy requires user-facing fixes to be recorded in the Unreleased changelog, and this maintainer-authored fix credits the reporter as required.

PR #913, opened 2026-07-05
fix(worker): bound run-history storage to stop Fleet DO OOM (Azure lease churn)

steipete:
## What Problem This Solves

The Crabbox coordinator's Fleet Durable Object OOMs (Cloudflare 128 MB isolate limit) and resets mid-request. Live tail of the deployed coordinator, provoked with an Azure lease, shows it on both `POST /v1/leases` and `GET /v1/runs`:

```
Error: Durable Object's isolate exceeded its memory limit and was reset.
```

Root cause: the run-history feature accumulates `run:*` records forever — run IDs are random (`newRunID`), there is no run retention (the alarm prunes only terminal *workspaces*), and `runRecords()` plus the maintenance alarm load the entire key-space with unbounded `storage.list`. DO isolates are reused across requests, so a `/v1/runs` load bloats the heap and the next request (a lease create) OOMs, loses the lease record, and the CLI retries — producing the ~16 s Azure VM create/teardown churn and the stranded `keep:false` orphans reported in #862.

## Why This Change Was Made

Stop the OOM so lease provisioning (Azure especially) stops churning, and keep run history bounded going forward.

- 7-day run retention, drained in **bounded, cursor-paginated batches** in the alarm so the maintenance path itself never loads the whole key-space (which would OOM and wedge the DO). Pruning deletes each expired run's metadata, log, log chunks, and events. In-flight runs (`state: "running"`) are never deleted; retention is measured from completion time.
- A read ceiling (`maxRunScanRecords`) so no read path (`listRuns`, portal history) loads the whole `run:*` key-space; retention keeps the true total under the ceiling. Follow-up: a time-ordered run index for exact newest-N ordering when volume exceeds the ceiling (noted in code).
- `CoordinatorStorage.list` is widened to expose the native `DurableObjectStorage` `limit`/`startAfter` options the bounded scan and pagination rely on (the abstraction is the real DO storage; only the type was narrowed).

## User Impact

Azure leases provision instead of churning; `GET /v1/runs` and portal lease history stop 500ing. No config or API surface change. Existing legacy `run:*` records drain out over the retention window.

## Evidence

- `tsgo`/tsc typecheck clean.
- `worker`: 382 tests pass (`vitest run test/fleet.test.ts test/coordinator-runtime.test.ts`), including a new test asserting the alarm prunes expired terminal runs (with log/chunk/event side data) while keeping recent and in-flight runs.
- Diagnosis with live coordinator tail on the OpenClaw Cloudflare account: openclaw/crabbox#862.
- Post-merge: deploy to the coordinator and verify a real Azure lease provisions end-to-end (no OOM in tail).

Refs #862

### DIFF
--- a/worker/src/coordinator-runtime.ts
+++ b/worker/src/coordinator-runtime.ts
@@ -2,7 +2,14 @@ export interface CoordinatorStorage {
   get<T>(key: string): Promise<T | undefined>;
   put<T>(key: string, value: T): Promise<void>;
   delete(key: string): Promise<unknown>;
-  list<T>(options?: { prefix?: string }): Promise<Map<string, T>>;
+  // limit/startAfter map straight to DurableObjectStorage.list; they let callers
+  // page a large key-space (e.g. run history) instead of loading it whole and
+  // OOMing the 128MB isolate.
+  list<T>(options?: {
+    prefix?: string;
+    limit?: number;
+    startAfter?: string;
+  }): Promise<Map<string, T>>;
 }
 
 export type CoordinatorRequestQueue = "direct" | "lifecycle";
--- a/worker/src/fleet.ts
+++ b/worker/src/fleet.ts
@@ -214,6 +214,20 @@ const workspaceMinimumTTLSeconds =
   (workspaceProvisionClaimMs + workspaceProvisionRecoveryGraceMs) / 1000;
 const workspaceMaxRecordsPerOwner = 100;
 const workspaceTerminalRetentionMs = 24 * 60 * 60_000;
+// Run history is otherwise unbounded: run IDs are random (not time-ordered) and
+// runs were never deleted, so `run:*` grows until the DO isolate OOMs on any
+// list-all (which then resets the DO mid-lease-create and churns provisioning).
+// Retain a bounded window; drain the backlog in the alarm; cap history reads.
+const runRetentionMs = 7 * 24 * 60 * 60_000;
+const runPruneBatchSize = 100;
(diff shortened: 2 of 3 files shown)


steipete:
Superseded by https://github.com/openclaw/crabbox/pull/909, landed as https://github.com/openclaw/crabbox/commit/979e6bf97ffea2a0ce462df3992786ea3934b177.

The landed fix covers the same Fleet Durable Object OOM without an arbitrary run-read ceiling: run history and lease detail use paged scans while retaining the true newest requested results, terminal runs and their events/log chunks drain through resumable bounded pruning with configurable retention, bridge restoration uses exact lease reads, and Azure orphan maintenance keeps only inventory-relevant ownership records. It was deployed and verified with repeated production history reads plus a create/use/release/zero-residue Azure canary.

Thanks for independently diagnosing and addressing the same root cause.

steipete:
Closing as superseded. While building this, #909 ("bound coordinator lease scans" + "page coordinator run history" + "prune terminal run history") landed on `main` and implements the same fix more completely — `pruneTerminalRuns` in the alarm, cursor-paginated bounded scans (`storageRecordScanBatchSize`), the widened `CoordinatorStorage.list`, and it also bounds coordinator lease scans (which this PR did not). It correctly prunes only terminal runs, matching the concern this PR's review raised.

Root cause and live evidence are documented in #862. The deployed coordinator had been stale since 2026-06-26 and did not yet include #909, which is why it was still OOMing. Resolved by deploying canonical `main` (which contains #909) to the coordinator and verifying the OOM is gone (details on #862). No code from this PR is needed.

PR #939, opened 2026-07-05
fix: disable git seed for credentialed HTTPS remotes

### LINKED ISSUE #928: [security] Git seed sends credential-bearing remote URLs to lease runners
# [security] Git seed sends credential-bearing remote URLs to lease runners

## Summary

Crabbox's SSH sync path can silently copy credentials embedded in the local Git `origin` remote URL into a leased runner. `findRepo` stores `git remote get-url origin` directly in `Repo.RemoteURL`; the default git-seed gate checks only whether the repository is seedable; and both Linux and native Windows seed commands pass the normalized URL to `git clone`. The normalizer only rewrites GitHub SSH shorthand and otherwise preserves HTTPS URL userinfo.

This crosses Crabbox's documented credential boundary rather than relying on hostile multi-tenant assumptions. The security policy treats Crabbox silently sending a trusted credential to a lower-trust destination as in scope, while the operations guide says secret values should not be passed as command-line arguments and environment forwarding is allowlist-based.

## Affected Components

- `internal/cli/repo.go:26-40`: `findRepo` reads `git remote get-url origin` and stores the value in `Repo.RemoteURL` without rejecting or stripping URL userinfo.
- `internal/cli/repo.go:144-145`: `syncGitSeedEnabled` enables git seed whenever `sync.gitSeed` is true, no include whitelist is configured, and the repository passes the seedability check.
- `internal/cli/repo.go:225-230`: `remoteGitSeedCandidate` requires only a root, remote URL, head commit, and containing remote ref; it does not treat credential-bearing HTTPS remotes as unsafe for seed.
- `internal/cli/run.go:1198-1203`: the normal `crabbox run` SSH sync flow sends `repo.RemoteURL` into `remoteGitSeed` before the rsync overlay.
- `internal/cli/actions.go:525-545`: local Actions workspace hydration uses the same `remoteGitSeed(workdir, repo.RemoteURL, repo.Head)` path.
- `internal/cli/ssh.go:1320-1342`: `remoteGitSeed` calls `normalizeGitRemoteURL` and embeds the returned URL in a remote `git clone` command; `normalizeGitRemoteURL` only rewrites `git@github.com:` URLs and otherwise returns the input unchanged.
- `internal/cli/sync_windows_target.go:12-19`: native Windows target sync invokes git seed when the same seedability gate passes.
- `internal/cli/sync_windows_target.go:81-92`: `windowsGitSeed` uses the same normalizer and places the resulting URL in a remote PowerShell `git clone` command.
- `SECURITY.md:34-43`: the maintainer-authored policy lists silent transmission of a trusted credential to a lower-trust destination as an in-scope boundary crossing.
- `docs/security.md:190-211`: the maintainer-authored operations guide keeps remote environment forwarding allowlist-based and says secret values should not be passed as command-line flags.

## Attack Path

An operator or CI job runs `crabbox run` or Crabbox local Actions hydration from a checkout whose `origin` remote is an HTTPS URL containing userinfo, such as a repository token. This can happen when automation or a local migration stores credentials directly in the remote URL.

Crabbox reads that remote URL from Git, decides the checkout is eligible for default git seed when `HEAD` is reachable from a remote ref, and sends the remote URL to the leased runner as part of the seed command. The Linux path passes it to `git clone --filter=blob:none --no-checkout`; the native Windows path does the same through PowerShell. Because `normalizeGitRemoteURL` does not remove HTTPS userinfo, the lease runner receives the credential-bearing URL.

The credential can then be exposed to the lease environment. It is present in the remote `git clone` command arguments while the seed operation runs, and a successful clone can preserve the credential-bearing `origin` URL in the seeded worktree's Git configuration. Later project commands, delegated tooling, debugging sessions, or captured failure artifacts on that runner may be able to read it.

## Impact

If the embedded credential has repository read or write scope, code running on the leased machine can recover and reuse it. That can expose private source, allow writes to the repository, or grant access to other repositories covered by the same credential. The practical blast radius depends on token scope and on what code runs in the lease after sync.

The issue has a meaningful precondition: the local Git remote must contain credential-bearing URL userinfo. It is still reportable because Crabbox's documented git-seed sync path moves that credential to a remote execution environment without the explicit credential-forwarding control, source binding, warning, or redaction that the security policy and operations guide reserve for secret material.

### CVSS Assessment

| Metric | v3.1 | v4.0 |
|--------|------|------|
| **Score** | 5.9 / 10.0 | 5.9 / 10.0 |
| **Vector** | CVSS:3.1/AV:N/AC:H/PR:N/UI:R/S:U/C:H/I:L/A:N | CVSS:4.0/AV:N/AC:H/AT:P/PR:N/UI:A/VC:H/VI:L/VA:N/SC:N/SI:N/SA:N |
| **Calculator** | [CVSS v3.1 Calculator](https://www.first.org/cvss/calculator/3.1#CVSS:3.1/AV:N/AC:H/PR:N/UI:R/S:U/C:H/I:L/A:N) | [CVSS v4.0 Calculator](https://www.first.org/cvss/calculator/4.0#CVSS:4.0/AV:N/AC:H/AT:P/PR:N/UI:A/VC:H/VI:L/VA:N/SC:N/SI:N/SA:N) |

The score is Medium because exploitation requires user action and the non-default condition of a credential-bearing Git remote, but successful exploitation exposes credential material to the leased environment. Confidentiality is High for the token itself, integrity is Low when the token has write scope, and availability impact is not required.

Treat Git remote URL userinfo as credential material before seed command construction. The safest default is to disable or reject git seed when the local remote contains userinfo unless Crabbox has an explicit, source-bound credential delivery mechanism for that destination. If private-repository seed support must be preserved, route credentials through an auditable, runner-scoped mechanism that avoids process arguments and avoids persisting the secret in the remote worktree.

Apply the control consistently across Linux SSH sync, native Windows sync, and local Actions hydration. Add operator-facing diagnostics that identify the unsafe credential-bearing remote shape without printing the credential. Add regression coverage for HTTPS remotes with userinfo, including checks that generated seed commands and seeded remote configuration cannot contain the secret.

## Validation

Source review of the current checkout confirmed the credential flow:

- `internal/cli/repo.go:26-40` reads and stores the raw origin remote URL.
- `internal/cli/repo.go:144-145` and `internal/cli/repo.go:225-230` enable git seed without checking URL userinfo.
- `internal/cli/run.go:1198-1203` and `internal/cli/actions.go:525-545` pass the raw remote URL into the Linux seed path.
- `internal/cli/ssh.go:1320-1342` preserves credential-bearing HTTPS URLs and inserts them into the remote clone command.
- `internal/cli/sync_windows_target.go:12-19` and `internal/cli/sync_windows_target.go:81-92` show the same behavior for native Windows targets.

Counterevidence checked: `shellQuote` and `psQuote` protect command syntax, but they do not remove credentials from process arguments or Git remote configuration. The existing credential provenance checks protect many Crabbox configuration and provider credential destinations, but this path uses a Git remote URL gathered from `git remote get-url origin`.

Focused package verification passed:

- `GOCACHE=/tmp/crabbox-gocache go test ./internal/cli -run 'TestRemoteGitSeedRemovesFailedCheckout|TestSyncGitSeedDisabledByIncludeWhitelist'`

TurboTheTurtle:
Closes #928.

Prevents git seed from cloning HTTPS remotes that include URL userinfo, so credential-bearing remotes fall back to the normal sync path instead of being sent to lease runners. Keeps SSH remotes with ordinary usernames eligible for git seed.

### DIFF
--- a/internal/cli/repo.go
+++ b/internal/cli/repo.go
@@ -226,7 +226,22 @@ func remoteGitSeedCandidate(repo Repo) bool {
 	if repo.Root == "" || repo.RemoteURL == "" || repo.Head == "" {
 		return false
 	}
+	if !remoteGitSeedURLSafe(repo.RemoteURL) {
+		return false
+	}
 	return gitOutput(repo.Root, "for-each-ref", "--contains", repo.Head, "--format=%(refname)", "refs/remotes") != ""
+}
+
+func remoteGitSeedURLSafe(remoteURL string) bool {
+	remoteURL = strings.TrimSpace(remoteURL)
+	if !strings.Contains(remoteURL, "://") {
+		return true
+	}
+	u, err := url.Parse(remoteURL)
+	if err != nil || u.User == nil {
+		return true
+	}
+	return u.Scheme != "http" && u.Scheme != "https"
 }
 
 func defaultBaseRef(root string) string {
--- a/internal/cli/repo_test.go
+++ b/internal/cli/repo_test.go
@@ -420,6 +420,19 @@ func TestRemoteGitSeedCandidateRequiresRemoteTrackingRef(t *testing.T) {
 	if !remoteGitSeedCandidate(repo) {
 		t.Fatal("head in a remote-tracking ref should be a seed candidate")
 	}
+
+	repo.RemoteURL = "https://token@github.com/openclaw/crabbox.git"
+	if remoteGitSeedCandidate(repo) {
+		t.Fatal("credential-bearing HTTPS remote should not be a seed candidate")
+	}
+	repo.RemoteURL = "https://user:token@github.com/openclaw/crabbox.git"
+	if remoteGitSeedCandidate(repo) {
+		t.Fatal("HTTPS remote with password userinfo should not be a seed candidate")
+	}
+	repo.RemoteURL = "ssh://git@github.com/openclaw/crabbox.git"
+	if !remoteGitSeedCandidate(repo) {
+		t.Fatal("SSH remotes with a username should remain seed candidates")
+	}
 }
 
 func TestCheckSyncPreflightFailsLargeCandidate(t *testing.T) {


steipete:
Thanks @TurboTheTurtle — this raced with the already-prepared maintainer fix and is now superseded by https://github.com/openclaw/crabbox/pull/941, landed as https://github.com/openclaw/crabbox/commit/50cfda5da8902a55f5a0cd9ba1699f3210f6a9fd.

The landed version keeps your core fail-safe direction and also covers Linux run, native Windows, and local Actions callers; adds redacted skip warnings and command-builder defense in depth; documents the behavior; and includes exact-head runtime clone/no-leak proof. Closing only to avoid overlapping implementations. Thank you for the quick security fix.

PR #945, opened 2026-07-06
fix(lambda): require claim-bound cleanup

### LINKED ISSUE #927: [security] Lambda cleanup trusts tag-only instances without a local claim
# [security] Lambda cleanup trusts tag-only instances without a local claim

## Summary

Lambda direct-provider cleanup currently treats provider-visible tags as enough proof that a Lambda Cloud instance is Crabbox-owned. If a Lambda instance in the configured provider account has the generic Crabbox label tuple but the local machine has no unchanged lease claim for that instance, `crabbox cleanup` can still terminate it and can also delete a tagged Lambda SSH key.

This crosses Crabbox's documented provider-resource boundary. `SECURITY.md` lists destructive provider actions against resources Crabbox cannot strongly identify as its own as in scope, while `docs/security.md` says provider names and `crabbox-` prefixes are discovery hints rather than destructive ownership proof. The issue is not that cleanup exists, but that this provider path allows tag-only inventory to become delete authority.

## Affected Components

- `internal/providers/lambda/tags.go:27-39`: `isOwnedInstance` accepts any provider inventory item whose tags satisfy `validateLambdaLabels`; the required fields are generic labels such as `crabbox=true`, `created_by=crabbox`, `provider=lambda`, a non-empty `lease`, a non-empty `slug`, and `target=linux`.
- `internal/providers/lambda/backend.go:603-625`: Lambda inventory is converted into cleanup-visible servers when `ownedServersFromInstances` sees `isOwnedInstance(item)`, so tag-only provider inventory can enter the cleanup flow even before any local claim is found.
- `internal/cli/pool.go:535-568`: the shared cleanup predicate selects expired `leased`, `ready`, and `active` records, stale `running` or `provisioning` records, and terminal states based on labels and timestamps.
- `internal/providers/shared/direct.go:28-45`: direct-provider cleanup calls the provider delete callback for every selected server when the request is not a dry run.
- `internal/providers/lambda/backend.go:503-570`: `deleteServer` reads a local claim, but only enforces claim-provider and claim-cloud-ID checks when `claimExists` is true; if no claim exists, the function can still terminate the live instance and delete the tagged SSH key.
- `internal/providers/lambda/backend_test.go:512-525`: the current regression test expects cleanup to terminate a provider-only expired Lambda instance, confirming this is the implemented behavior rather than a purely theoretical path.
- `SECURITY.md:34-49`: the maintainer security policy includes destructive provider actions against resources Crabbox cannot strongly identify as its own as an in-scope boundary.
- `docs/security.md:286-306`: the operational security guidance says comparable direct-provider cleanup paths should require exact claims and should skip weakly labeled, unclaimed, and stale-claim servers instead of treating provider inventory as ownership proof.

## Attack Path

Attacker role or triggering condition:

A lower-trust actor or workflow with access to create or modify Lambda Cloud instance tags in the configured provider account, a provider inventory response containing stale Crabbox-like tags, or a resource left behind after local state loss.

Prerequisites:

- The victim is using the Lambda direct provider.
- A Lambda Cloud instance appears in provider inventory with tags that satisfy `validateLambdaLabels`.
- The cleanup labels indicate an expired or stale cleanup state, or another direct stop/release flow resolves that tagged id.
- The current local Crabbox installation does not have an unchanged lease claim binding the lease id to that exact Lambda Cloud instance.

Steps:

1. Lambda provider inventory returns an instance whose tags pass `isOwnedInstance`.
2. `ownedServersFromInstances` converts that inventory item into a `core.Server` and returns it to the cleanup flow.
3. `ShouldCleanupServer` selects the server from its labels and timestamps.
4. `CleanupServers` invokes the Lambda delete callback.
5. `deleteServer` observes that no local claim exists, but does not fail closed on that condition.
6. The function validates the live instance against the same provider-tag-derived server data, calls `TerminateInstances`, and deletes the tagged SSH key when the labels mark it as owned.

Expected result:

Crabbox can terminate a Lambda Cloud instance, and can remove a Lambda SSH key, even though it cannot bind the resource to an unchanged local Crabbox lease claim or an equivalent unforgeable ownership proof.

## Impact

This turns provider-visible metadata into authority for destructive Lambda operations. In a shared cloud account, after local state loss, or when stale provider-side tags survive beyond the original local claim, a Crabbox cleanup or stop operation can delete resources that the current Crabbox installation cannot strongly identify as its own.

The primary impact is integrity and availability loss for affected Lambda instances. A secondary impact is availability and operational disruption from deleting a tagged Lambda SSH key. This fits the maintainer-defined security boundary for unsafe destructive provider actions and is distinct from ordinary missing orphan cleanup or cost-control hardening.

### CVSS Assessment

| Metric | v3.1 | v4.0 |
|--------|------|------|
| **Score** | 6.6 / 10.0 | 4.4 / 10.0 |
| **Vector** | `CVSS:3.1/AV:L/AC:L/PR:L/UI:R/S:U/C:N/I:H/A:H` | `CVSS:4.0/AV:L/AC:L/AT:P/PR:L/UI:A/VC:N/VI:H/VA:H/SC:N/SI:N/SA:N` |
| **Calculator** | [CVSS v3.1 Calculator](https://www.first.org/cvss/calculator/3.1#CVSS:3.1/AV:L/AC:L/PR:L/UI:R/S:U/C:N/I:H/A:H) | [CVSS v4.0 Calculator](https://www.first.org/cvss/calculator/4.0#CVSS:4.0/AV:L/AC:L/AT:P/PR:L/UI:A/VC:N/VI:H/VA:H/SC:N/SI:N/SA:N) |

The score is Medium because exploitation requires Lambda provider-account or stale-inventory preconditions plus an operator cleanup or stop action, but once those preconditions hold the vulnerable path can cause high integrity and availability impact to provider resources. Confidence is high for the source path because current code and tests show provider-only tags are sufficient for Lambda cleanup; the remaining uncertainty is the exact live Lambda Cloud tag-mutability model in a given deployment.

Require an unchanged local Lambda lease claim, or an equivalent unforgeable per-lease ownership proof, before any Lambda direct-provider operation terminates an instance or deletes a provider SSH key. Treat provider tags as discovery and reporting hints unless they are tied back to a local claim or a deliberate reclaim flow.

Preserve rollback and failed-create recovery, but bind those paths to a locally recorded recovery claim and exact resource identity. Ambiguous provider-only inventory should be skipped or reported for operator review rather than deleted automatically. Add regression coverage for tag-only Lambda inventory, missing-claim cleanup, SSH-key deletion labels, and the approved local-claim or reclaim paths.

## Validation

Validation was performed against commit `bd9adff6ec950f3584c57883f82ce35e77157a4d`.

Evidence:

- `internal/providers/lambda/tags.go:27-39` shows provider tags alone can satisfy `isOwnedInstance`.
- `internal/providers/lambda/backend.go:603-625` shows those instances are returned as owned cleanup candidates.
- `internal/cli/pool.go:535-568` and `internal/providers/shared/direct.go:28-45` show selected expired, stale, or terminal records are passed to the provider delete callback.
- `internal/providers/lambda/backend.go:511-570` shows `deleteServer` only applies local-claim mismatch checks when a claim exists, then can terminate the instance and delete the tagged SSH key.
- `internal/providers/lambda/backend_test.go:512-525` confirms the current expected behavior is to delete a provider-only expired Lambda instance.
- `SECURITY.md:34-49` and `docs/security.md:286-306` confirm the reportability boundary: destructive provider mutations need stronger ownership proof than tags or resource prefixes.

Counterevidence considered:

- `internal/providers/lambda/backend.go:515-522` rejects mismatched existing local claims, which protects stale local-claim cases but not the no-claim path.
- `internal/providers/lambda/backend.go:527-538` validates the live provider instance before termination, but that validation uses the same tag-derived ownership model and does not establish an unchanged local claim.
- The issue does not depend on hostile repository configuration, arbitrary commands on a leased machine, or mutually adversarial tenants sharing a broker; it depends on Crabbox performing a destructive provider mutation without strong ownership proof.

Focused verification command:

`GOCACHE=/tmp/crabbox-go-cache GOTMPDIR=/tmp go test ./internal/providers/lambda -run TestCleanupDeletesProviderOnlyExpiredLambdaInstance -count=1`

The command passed and confirms the current test suite still expects provider-only expired Lambda inventory to be terminated.

steipete:
Fixes https://github.com/openclaw/crabbox/issues/927

## Summary

- require an exact local Lambda claim before inventory, raw-ID resolution, stop, or cleanup can act on an instance
- carry a cloned claim-generation snapshot through every lifecycle mutation and fail closed when ownership or cleanup eligibility changes
- bind ambiguous-create recovery to exactly one live instance carrying the recorded per-lease SSH key, then persist that instance binding before termination so partial cleanup is retryable
- remove tag-only compatibility behavior and document provider tags, names, and IDs as discovery hints only

## Verification

- `go vet ./...`
- `go test -race -p=1 ./...`
- `npm test --prefix worker`
- AutoReview: clean after accepted ownership-transfer, lifecycle-generation, recovery-conflict, durable-binding, and Tailscale metadata findings

The first parallel full-race run hit the existing Apple helper readiness timing test once; the isolated test then passed three race-enabled reruns, and the serialized full race suite passed.

## Live proof gate

Authenticated Lambda create/use/destroy proof could not run because no launch-capable API credential is currently available. The owner explicitly waived this item-specific live gate on July 6, 2026; local lifecycle regression proof and exact-head hosted CI remain required before merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 - Confined Islo API redirects to the configured scheme, hostname, and effective port before replaying authorization or request bodies, while keeping rejected Location secrets out of diagnostics. Thanks @TurboTheTurtle.
 - Redacted colon-delimited and line-folded bearer credentials from CLI and coordinator diagnostics. Thanks @TurboTheTurtle.
 - Required coordinator AWS, Azure, and GCP release and provisioning-failure cleanup to re-read the stored cloud resource and verify exact provider, resource, lease, owner, and slug ownership before deletion; Azure now persists the exact subscription/resource-group scope for deferred retries and fails legacy unscoped cleanup closed for manual resolution. Thanks @coygeek.
-- Required Lambda inventory, stop, and cleanup to use an unchanged instance-bound local claim, with claim-bound SSH-key deletion and explicit per-lease-key recovery for ambiguous creates. Thanks @coygeek.
+- Required Lambda inventory, stop, and cleanup to use an unchanged instance-bound local claim, with claim-bound SSH-key deletion and durable unique-instance recovery for ambiguous creates. Thanks @coygeek.
 
 ## 0.36.0 - 2026-07-05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Confined Islo API redirects to the configured scheme, hostname, and effective port before replaying authorization or request bodies, while keeping rejected Location secrets out of diagnostics. Thanks @TurboTheTurtle.
 - Redacted colon-delimited and line-folded bearer credentials from CLI and coordinator diagnostics. Thanks @TurboTheTurtle.
 - Required coordinator AWS, Azure, and GCP release and provisioning-failure cleanup to re-read the stored cloud resource and verify exact provider, resource, lease, owner, and slug ownership before deletion; Azure now persists the exact subscription/resource-group scope for deferred retries and fails legacy unscoped cleanup closed for manual resolution. Thanks @coygeek.
+- Required Lambda inventory, stop, and cleanup to use an unchanged instance-bound local claim, with claim-bound SSH-key deletion and explicit per-lease-key recovery for ambiguous creates. Thanks @coygeek.
 
 ## 0.36.0 - 2026-07-05
 
--- a/docs/providers/lambda.md
+++ b/docs/providers/lambda.md
@@ -156,7 +156,8 @@ Lambda has no safe tag-update path in this phase, and the launch path does not
 depend on provider-side tag persistence. Local Crabbox claims carry fresh touch
 and idle-timeout state. Explicit failed-create recovery remains claim-bound:
 ambiguous launches must match the unique per-lease Lambda SSH key before Crabbox
-can terminate them.
+persists the recovered instance binding before termination, so partial cleanup
+can safely resume after the instance disappears.
 
 Direct mode has no coordinator alarm. Use:
 
--- a/docs/providers/lambda.md
+++ b/docs/providers/lambda.md
@@ -121,8 +121,8 @@ classed doctor output rather than hidden behind generic errors.
 5. Claim the lease locally with Lambda instance and SSH-key metadata.
 6. Run normal Crabbox sync/run/ssh workflows over SSH.
 7. Terminate the Lambda instance and owned Lambda SSH key on `stop`; `cleanup`
-   deletes resources backed by a complete local Crabbox Lambda claim, and can
-   also reclaim complete Crabbox-tagged Lambda instances.
+   deletes only resources backed by an unchanged, exact local Crabbox Lambda
+   claim.
 
 If SSH-key creation, instance launch, or post-launch readiness becomes
 indeterminate, Crabbox records a local recovery claim with enough metadata to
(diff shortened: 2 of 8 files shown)


steipete:
Landed as https://github.com/openclaw/crabbox/commit/d2660efa48dfa8943cba81381e9c769f69aa70d3.

Verification:

- `go vet ./...`
- `go test -race -p=1 ./...`
- Worker format, lint, typecheck, Vitest, and build gates
- docs-site build
- AutoReview clean on the complete branch after the accepted ownership-transfer, lifecycle-generation, recovery-conflict, durable-binding, and Tailscale metadata fixes
- exact-head hosted CI and CodeQL: all checks green at `ce88987de521a39cd2744d00a16a9914ddcd1f41`

Live Lambda create/use/destroy proof was unavailable because the authenticated account had no launch-capable API credential. The owner explicitly waived that item-specific live gate before merge. No Lambda resources were created.

PR #984, opened 2026-07-06
fix(security): bind External config argv

### LINKED ISSUE #950: [security] External lifecycle templates can place inherited config credentials on argv
# [security] External lifecycle templates can place inherited config credentials on argv

## Summary

Crabbox's External provider declarative lifecycle expands `{{config.*}}` placeholders into lifecycle `argv` and `steps`, but it does not carry sensitivity or source-provenance metadata for those config-derived values. That creates a reportable mixed-source configuration case: a trusted user config can provide `external.config` values, while a lower-trust repository config can replace only the External lifecycle command and reference the inherited config value in process arguments.

This is not a report about repository-only project automation. Crabbox's security policy treats repository-configured commands, helper arguments, and repo-provided secrets as expected behavior. The supported boundary crossed here is narrower: Crabbox documents that inherited user-config or environment credentials should not be sent to a repository-defined destination unless the operator explicitly overrides that destination, but the External lifecycle template engine currently lets a repo-controlled lifecycle command select argv as the destination for an inherited `external.config` value.

## Affected Components

- `docs/features/configuration.md:14-26`: Documents the source-boundary model: repository config may select providers, helper commands, runtimes, images, and runtime arguments, while user config and environment variables remain the expected sources for operator credentials; inherited credentials should not be sent to repository-defined destinations without an explicit operator override.
- `docs/features/configuration.md:30-36`: Documents the merge order and fallthrough rule: user config is applied before repo-local config, and each layer only overrides fields that are explicitly set.
- `internal/cli/config.go:437-447`: Defines `ExternalConfig` as one structure containing `Config`, `Lifecycle`, and `Connection`, which means the effective External provider state can combine fields from multiple config layers.
- `internal/cli/config.go:5716-5733`: Applies External config field by field. A repo-local config can replace `external.lifecycle` or `external.connection` while leaving a previously loaded `external.config` map in place when the repo file omits that field.
- `internal/providers/external/lifecycle.go:37-58`: Invokes External lifecycle operations by expanding each configured `argv` or `steps` command through `expandLifecycleArgv` before launching the command environment.
- `internal/providers/external/lifecycle.go:551-554`: Adds all effective `external.config` entries into lifecycle template values, then initializes the sensitivity map as empty.
- `internal/providers/external/lifecycle.go:604-625`: Recursively flattens `external.config` scalar values into `config.*` template keys without marking any of those keys sensitive.
- `internal/providers/external/lifecycle.go:627-638`: Blocks lifecycle argv only when the expanded item is marked sensitive and `allowEnvArgv` is not set.
- `internal/providers/external/lifecycle.go:664-692`: Marks `{{env.*}}` expansions sensitive, but non-env placeholders such as `{{config.token}}` only inherit sensitivity from `templateCtx.sensitive`, which has no entries for config-derived values.
- `docs/providers/external.md:157-177`: Documents the intended argv safety behavior for environment-derived secrets: `{{env.*}}` values are rejected in lifecycle argv by default and should be passed through operation `env:` instead.

## Attack Path

Attacker role: a repository-controlled configuration source, such as a branch or pull request that an operator reviews and runs with Crabbox.

Prerequisites:

1. The operator has trusted user config, or another trusted config file outside the active repository, that defines an External provider `external.config` value used by a site-local adapter. Examples include a token, namespace secret, project credential, or another sensitive adapter setting.
2. The repository supplies or modifies only External lifecycle command fields, leaving `external.config` unset so the trusted value falls through from user config.
3. The operator runs Crabbox in that repository.

Steps:

1. Crabbox loads defaults, then trusted user config, then repo-local config.
2. The trusted layer contributes a value such as `external.config.token`.
3. The repo-local layer contributes an External lifecycle operation whose `argv` or `steps` item references `{{config.token}}`.
4. `applyFileConfigWithTrust` replaces the lifecycle operation but does not clear inherited `external.config` when the repo file omits that field.
5. `lifecycleContext` flattens the effective config map into `config.*` template values without sensitivity metadata.
6. `expandLifecycleArgv` expands `{{config.token}}` into a command argument and does not trigger the `allowEnvArgv` guard because the expansion is not marked sensitive.
7. Crabbox launches the lower-trust lifecycle command with the trusted config value in process arguments, exposing the value to the selected helper process and normal process-argument observation paths.

This crosses Crabbox's supported trusted-credential boundary because a value from a trusted operator configuration layer can be sent to a destination selected by a different, lower-trust repository configuration source. A repo-only External config that stores its own value and passes it to its own command remains expected project automation and is not the reportable case.

## Impact

A repository can cause inherited External provider config values from trusted operator config to be copied into process arguments for repo-defined lifecycle commands. For sensitive adapter settings, this exposes the value to the selected helper command and to process argument surfaces that Crabbox's own External provider documentation tells operators not to use for secrets.

The impact is credential disclosure rather than command injection. The affected boundary is not hostile multi-tenant isolation, another trusted operator intentionally attacking the same broker, or another process under the same trusted OS account by itself. The issue is Crabbox's documented source-bound credential rule: a lower-trust repository layer can select argv as the destination for a trusted inherited config value without the explicit opt-in required for environment-derived values.

### CVSS Assessment

| Metric | v3.1 | v4.0 |
|--------|------|------|
| **Score** | 6.3 / 10.0 | 6.7 / 10.0 |
| **Vector** | `CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:C/C:H/I:N/A:N` | `CVSS:4.0/AV:L/AC:L/AT:N/PR:N/UI:A/VC:H/VI:N/VA:N/SC:L/SI:N/SA:N` |
| **Calculator** | [CVSS v3.1 Calculator](https://www.first.org/cvss/calculator/3.1#CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:C/C:H/I:N/A:N) | [CVSS v4.0 Calculator](https://www.first.org/cvss/calculator/4.0#CVSS:4.0/AV:L/AC:L/AT:N/PR:N/UI:A/VC:H/VI:N/VA:N/SC:L/SI:N/SA:N) |

The score reflects local execution with active operator interaction: an operator must run Crabbox in a repository that supplies the lifecycle command. Confidentiality impact is high for the exposed trusted config value, while integrity and availability are not directly affected. Severity is kept at Medium because repository configuration is executable project automation in Crabbox's trusted-operator model, and repo-only config secrets remain outside the vulnerable mixed-source case.

Track source and sensitivity for `external.config` values used in lifecycle templates. Config-derived template values that originate from trusted user config or another trusted non-repository layer should be treated like environment-derived values when they flow into lifecycle argv, steps, or resource-name expansion. Preserve supported workflows for non-secret config values, but require credentials to flow through operation `env:`, an adapter credential store, or another explicit non-argv channel.

Add regression coverage for the mixed-source case: trusted user config contributes `external.config.token`, repo-local config contributes lifecycle argv or steps referencing `{{config.token}}`, and Crabbox rejects the command unless an explicit non-secret argv opt-in is present. Also cover the negative cases: repo-local config-only values remain ordinary project automation, and existing `{{env.*}}` argv and operation `env:` behavior remain unchanged.

## Validation

Validation method: source review with a bounded focused-test attempt.

Evidence:

- `docs/features/configuration.md:14-26` and `docs/features/configuration.md:30-36` document both the source-bound credential model and the field fallthrough behavior that creates the mixed-source case.
- `internal/cli/config.go:437-447` shows `ExternalConfig` combines config, lifecycle, and connection fields into one effective provider configuration.
- `internal/cli/config.go:5716-5733` confirms `external.config` is not cleared when repo-local config sets lifecycle or connection fields but omits `external.config`.
- `internal/providers/external/lifecycle.go:37-58` confirms configured lifecycle `argv` and `steps` are expanded before command launch.
- `internal/providers/external/lifecycle.go:551-554` confirms effective config values are added to lifecycle template values while the sensitivity map starts empty.
- `internal/providers/external/lifecycle.go:604-625` confirms nested config values are flattened without sensitivity metadata.
- `internal/providers/external/lifecycle.go:627-638` confirms lifecycle argv rejection depends on the expansion being marked sensitive.
- `internal/providers/external/lifecycle.go:664-692` confirms env placeholders are marked sensitive, while config placeholders rely on the empty `templateCtx.sensitive` map.
- `docs/providers/external.md:157-177` confirms the intended argv safety behavior for environment-derived secrets and instructs operators to pass secrets through lifecycle `env:` rather than argv.

Counterevidence considered:

- `SECURITY.md` treats repository-configured commands, helpers, runtime arguments, and secrets explicitly passed through project configuration as expected behavior. This report does not rely on that repo-only case.
- `docs/providers/external.md:35-37` advises operators to keep secrets in the executable's credential store or environment rather than in Crabbox config. That limits the affected population, but it does not remove the source-boundary issue when a trusted inherited config value is present and repo-local lifecycle fields can select its argv destination.
- Crabbox rejects `{{env.*}}` values in lifecycle argv by default, which is a positive control for environment-derived secrets, but the same tracking is not applied to inherited `{{config.*}}` values.

Runtime validation note:

- A focused `go test ./internal/providers/external` run was attempted with `GOCACHE` and `GOMODCACHE` under `/tmp`, but this isolated verifier workspace could not download missing Go modules because network access is disabled (`proxy.golang.org` lookup failed). The verdict is therefore based on static source and documentation evidence from the checked target snapshot.

steipete:
## Summary

- track the trust source of External lifecycle and connection templates
- reject inherited `external.config` values in repository-selected argv/steps, including indirect `resourceName` and `cloudId` flows
- add an operation-scoped `allowConfigArgv` opt-in that requires an exact trusted lifecycle and connection-template contract
- advance persisted routing credential-boundary validation so legacy and repository-selected routes cannot bypass the new rule
- document the safe environment channel and compatibility contract

Closes https://github.com/openclaw/crabbox/issues/950

## Verification

- `gofmt -w $(git ls-files '*.go')`
- `git diff --check`
- `go vet ./...`
- `go test -race ./...`
- `go run golang.org/x/tools/cmd/deadcode@v0.45.0 -test ./...`
- `npm test --prefix worker`
- AutoReview clean after fixing three accepted routing/approval findings

## Live CLI proof

Built the exact candidate and exercised real layered user/repository configuration through `crabbox list`:

- inherited user-config value plus repository lifecycle argv: rejected before the helper ran
- exact trusted lifecycle/connection contract with `allowConfigArgv`: allowed and passed the non-secret sentinel to the helper
- repository-owned `external.config` plus repository lifecycle: allowed unchanged

The canary used local non-secret sentinels and created no remote resource or persistent credential.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- Prevented repository-defined External lifecycle commands from placing inherited `external.config` values on process arguments without an exact, trusted non-secret argv contract. Thanks @coygeek.
 - Prevented repository-controlled External SSH endpoint templates and adapter output from silently using ambient or operator-managed SSH credentials, with source-bound opt-ins for environment-derived fields and provider-returned destinations. Thanks @coygeek.
 - Made Sealos DevBox preflight work with tenant-scoped RBAC, rendered the runtime class, storage request, scheduling constraints, and SSH port contract required by hosted Sealos clusters, updated the SSHGate default to port 2233, bootstrapped missing sync tools, and cleaned local claims safely when a DevBox is already absent. Thanks @coygeek.
 - Changed portal logout to an authenticated, same-origin `POST` with a read-only `GET` confirmation page, preventing cross-site top-level navigation from clearing portal cookies or revoking isolated Code viewer sessions. Thanks @coygeek.
--- a/docs/features/configuration.md
+++ b/docs/features/configuration.md
@@ -788,7 +788,11 @@ not shell commands. Acquire steps can opt into release cleanup with
 `rollbackOnFailure: true`. Put credentials in an operation `env:` map, not in
 `argv` or `steps`; environment-derived argv values require the explicit
 `allowEnvArgv: true` compatibility opt-in, and environment-derived resource
-names require `connection.allowEnvResourceName: true`. See
+names require `connection.allowEnvResourceName: true`. Repository lifecycle
+commands cannot place inherited `external.config` values in argv without an
+exact lifecycle and config-derived connection-template contract carrying
+`allowConfigArgv: true` in trusted user config; repository-owned config values
+are unchanged. See
 [External Provider](../providers/external.md) for placeholders, output
 semantics, inventory formats, routing behavior, controller-safe raw
 `json-lease` identity attestation, and security guidance.
--- a/docs/providers/external.md
+++ b/docs/providers/external.md
@@ -192,6 +192,17 @@ later release still requires `allowEnvArgv` before placing `{{resourceName}}`
 back in argv. Do not use either opt-in for tokens, passwords, API keys, or
 other credential contents.
 
+Repository lifecycle commands also cannot place `{{config.*}}` values inherited
+from user config, an explicit config file, or `--external-config-json` in
+`argv` or `steps`. Pass credentials through the operation `env:` map. For an
+inherited non-secret value that must be an argument, set `allowConfigArgv: true`
+on that operation and repeat the exact full lifecycle plus `resourceName` and
+`cloudId` template contract in trusted user config. The repository cannot grant
+itself this opt-in, and changing any lifecycle command, argument, environment
+mapping, output mode, cleanup behavior, or config-derived connection template
+invalidates the approval. Repository-owned `external.config` values remain
+ordinary project automation.
+
 Environment-derived expansion in `connection.ssh` fields is rejected by
 default, including indirect use through an environment-backed `resourceName`,
 because values such as the SSH user, host, key path, ready check, or proxy
--- a/internal/cli/config.go
+++ b/internal/cli/config.go
@@ -487,6 +487,7 @@ type ExternalLifecycleOperation struct {
 	Steps             [][]string        `yaml:"steps,omitempty" json:"steps,omitempty"`
 	Env               map[string]string `yaml:"env,omitempty" json:"env,omitempty"`
 	AllowEnvArgv      bool              `yaml:"allowEnvArgv,omitempty" json:"allowEnvArgv,omitempty"`
+	AllowConfigArgv   bool              `yaml:"allowConfigArgv,omitempty" json:"allowConfigArgv,omitempty"`
 	Output            string            `yaml:"output,omitempty" json:"output,omitempty"`
 	NamePrefix        string            `yaml:"namePrefix,omitempty" json:"namePrefix,omitempty"`
 	RollbackOnFailure bool              `yaml:"rollbackOnFailure,omitempty" json:"rollbackOnFailure,omitempty"`
@@ -5854,10 +5855,12 @@ func applyFileConfigWithTrust(cfg *Config, file fileConfig, trusted bool) error 
 		}
 		if file.External.Lifecycle != nil {
 			cfg.External.Lifecycle = *file.External.Lifecycle
+			cfg.credentialProvenance.externalLifecycle = credentialSource
 		}
 		if file.External.Connection != nil {
 			cfg.External.Connection = *file.External.Connection
 			ssh := cfg.External.Connection.SSH
+			cfg.credentialProvenance.externalConnection = credentialSource
 			cfg.credentialProvenance.externalSSHConnection = credentialSource
 			if trusted {
 				outputContract, outputContractOK := externalProviderOutputContract(cfg.External)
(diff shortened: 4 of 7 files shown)


steipete:
Exact-head proof for `16eddbc6a665852e70db1ba8a43ffa262198fb64`:

- AutoReview: clean after fixing three accepted approval/routing bypasses.
- Local full gate: gofmt, diff check, go vet, go test -race, dead-code analysis, Worker format/lint/typecheck/tests/build, and docs build all passed.
- Exact-built CLI layered-config canary: an inherited user value was rejected before helper execution; an exact trusted lifecycle/connection contract with `allowConfigArgv` was allowed; repository-owned config remained allowed.
- Routing regressions cover legacy claim state, current claim-bound validated state, and repository-selected current-version routes.
- Cleanup: no remote resource or persistent credential.

PR #996, opened 2026-07-06
feat: add signed run receipts with crabbox run --attest and crabbox verify

yetval:
## Maintainer review update

Exact reviewed head: `c28da3ce2718b03c6cdbceb0e698e1e12250e154`, rebased onto current `main`.

- Protected both the receipt and effective signing key from collisions with lease/proof/capture/download/timing outputs, including symlink, hard-link, dangling-parent, and filesystem case aliases.
- Made first-use key minting atomic across concurrent processes; losing callers load the winning key instead of returning transient identities.
- Rejected `--attest-key` without `--attest` and validated override keys before any remote execution.
- Preserved delegated provider, lease, slug, run, and actions identity from session handles.
- Made the trust boundary visible in every result with `trust=self-signed`; a `PASS` is meaningful only after matching the fingerprint through a trusted channel.
- Added the maintainer-owned `Unreleased` changelog entry with contributor credit.

Exact-head verification: `go test -race ./...`, `go vet ./...`, `go build -trimpath -o /tmp/crabbox-pr996-current ./cmd/crabbox`, and `node scripts/check-command-docs.mjs` all pass. Four autoreview rounds addressed every accepted finding; the final review is clean with no actionable findings.

The maintainer exact-head proof below exercises the final trust marker and path hardening over real loopback SSH. The longer proof later in the body was captured by the contributor at head `7ec4e381` and covers tamper/schema rejection.

## Summary

Crabbox is built around run evidence: proof blocks (`--emit-proof`), content-addressed capsule manifests, timing JSON, JUnit summaries, artifact manifests with SHA-256 digests. None of that evidence is cryptographically signed, and there is no command to check any of it after the fact. A reviewer or an agent overseer has to trust pasted output.

This change adds the minimal provider-neutral primitive on top of what runs already collect:

- `crabbox run --attest <path>` writes `receipt.json` after a successful run: a flat JSON record of provider, lease, command, exit code, timing, and the SHA-256 of the combined stdout and stderr stream as observed by the client, signed with a per-user Ed25519 key.
- `crabbox verify <receipt.json>` recomputes the canonical bytes, checks the signature, and prints PASS or FAIL with the signer public key fingerprint and `trust=self-signed`. Exit 0 on PASS, 1 on signature mismatch, 2 for receipts that cannot be checked at all.
- The signing key is minted on first use at `<user config dir>/crabbox/attest/id_ed25519.pem` (PKCS8 PEM, 0600, same mint-once pattern as the per-lease SSH keys). `--attest-key <path>` signs with an existing PKCS8 PEM Ed25519 key and fails closed; it never creates a key at an override path.

Everything uses the Go standard library (`crypto/ed25519`, `crypto/x509`, `encoding/pem`); no new dependencies.

## Design notes

- The signature covers the canonical encoding of every receipt field except `signature` itself: the object minus `signature` is re-marshaled as compact JSON, which sorts map keys lexicographically. Because `public_key` is inside the signed payload, swapping in a different key also fails verification. Empty optional fields are omitted, never emitted as empty strings, so signer and verifier agree without per-field logic.
- The output digest is a fresh SHA-256 writer teed into the stream fan-out after the capture wrapping, so it sees the same bytes the console does, including under `--capture-stdout`. It deliberately does not hash the internal log ring buffer, which truncates at 8 MiB. Because the stdout and stderr streams write from separate goroutines, the digest is wrapped in a mutex-guarded writer, matching the synchronized run log buffer those streams already share, so `log_sha256` is always the hash of the bytes in arrival order and can never be corrupted by concurrent writes.
- Both run paths write receipts: the direct SSH path includes `run_id` and `log_sha256`; the delegated path preserves available session identity fields but omits `log_sha256` because a delegated `RunResult` carries only a bounded log excerpt, and hashing an excerpt would be misleading evidence. Dispatch stays by feature, not provider name, per `docs/vision.md`.
- `--attest` mirrors `--emit-proof`: path-valued, success-only, rejected with `--sync-only`, and preflighted with the same local output path checks.
- Hardening from the review rounds: `verify` rejects receipts whose JSON contains duplicate object keys anywhere (exit 2), because JSON parsers disagree on which duplicate wins and an edited receipt could show one value to a reader while the signature still verifies against another; beyond that, receipts are strictly schema validated before any signature decision: unknown fields, missing required fields, wrong types, non-integer number spellings, unsupported schema versions, invalid timestamps, and malformed digests are all rejected as malformed (exit 2), so `PASS` can only ever be printed for a well-formed, integrity-checked record. The receipt and effective signing-key paths are preflighted against every writable local output with filesystem-aware alias detection; `--attest`/`--attest-key` are forbidden in `watch` and `shard` forwarding, matching `--emit-proof`; missing parent directories use private modes; concurrent first use publishes exactly one stable key; and delegated receipts preserve every available session identity while omitting unavailable fields. The contributor changelog edit was dropped and a maintainer-owned entry was added during landing review.

## Config and secret implications

- New key file at `<user config dir>/crabbox/attest/id_ed25519.pem`, created 0600 in a 0700 directory through atomic no-replace publication. It is a signing identity, not a provider credential, and never leaves the machine.
- No secrets on argv: `--attest-key` takes a file path, never key material.
- No config schema changes and no coordinator or Worker changes.

## Verification

- `gofmt -l $(git ls-files '*.go')` clean
- `go vet ./...` clean
- `go run golang.org/x/tools/cmd/deadcode@v0.45.0 -test ./...` empty
- `go test -race ./...` all packages pass
- `go build -trimpath -o bin/crabbox ./cmd/crabbox` builds
- `node scripts/check-command-docs.mjs` passes (54 command docs, includes the new `verify` page and `crabbox verify --help` exit 0)
- `scripts/check-go-coverage.sh 90.0` passes
- New regression tests in `internal/cli/attest_test.go`: receipt round trip through the real `verify` command dispatch, empty-optional-field omission, a tamper table (mutated exit code, mutated command, foreign public key, corrupt signature encoding, deleted signature, duplicate `exit_code` key with the signed value last, duplicate `command` key with the signed value last, truncated JSON, missing file) asserting exit codes 1 vs 2, receipt path collision preflights against capture and download destinations, key mint idempotence with 0600/0700 permission checks, and `--attest-key` override semantics including fail-closed on a missing path and rejection of non-Ed25519 keys. `watch` and `shard` forbidden-flag suites cover `--attest` and `--attest-key` rejection. Digest-writer regressions run under the race detector: concurrent mixed stdout/stderr writes through the shared digest produce the exact expected hash with no data race (an unsynchronized hash in the same shape fails `-race` immediately), and a sequential mixed-stream case pins arrival-order hashing. Schema-validation regressions cover unsupported schema versions, unknown fields, decimal integer spellings, and invalid digests; a nested `--attest` path regression checks created directory modes; and a delegated-receipt regression covers a provider returning successful run metadata with no LeaseID, asserting the empty fields are omitted and the receipt still verifies.

## Exact-head real behavior proof

Built and exercised `c28da3ce2718b03c6cdbceb0e698e1e12250e154` with an isolated home directory, the static `ssh` provider explicitly pinned, a real local sshd, and temporary work/output paths. Temporary paths and the local username are redacted below; no cloud lease was created.

```text
$ git rev-parse HEAD
c28da3ce2718b03c6cdbceb0e698e1e12250e154

$ HOME=<isolated-home> /tmp/crabbox-pr996-current run --provider ssh --target macos \
    --static-host 127.0.0.1 --static-user <local-user> \
    --static-work-root <temporary-work-root> --no-sync \
    --attest <temporary-output>/receipt.json -- \
    /bin/sh -c 'printf "attest current-head\n"; uname -s'
using static target lease=static_127-0-0-1 slug=127-0-0-1 target=macos host=127.0.0.1 keep=false
running on 127.0.0.1 /bin/sh -c 'printf "attest current-head\n"; uname -s'
attest current-head
Darwin
artifact kind=receipt path=<temporary-output>/receipt.json bytes=520
lease cleanup stopped=true policy=auto lease=static_127-0-0-1 slug=127-0-0-1

$ HOME=<isolated-home> /tmp/crabbox-pr996-current verify <temporary-output>/receipt.json
PASS <temporary-output>/receipt.json signer=sha256:4ab06eb561fa17ba12b50bbe71d4c4d3a8abc4eb35b171e7c4263ef5be4a0b1c trust=self-signed

$ printf 'attest current-head\nDarwin\n' | shasum -a 256
0942612941e6d955b36963ac8de357e867e947d0db57e6c1f19070472a66aa52  -
$ jq -r '.log_sha256, .provider, .exit_code' <temporary-output>/receipt.json
sha256:0942612941e6d955b36963ac8de357e867e947d0db57e6c1f19070472a66aa52
ssh
0

$ HOME=<isolated-home> /tmp/crabbox-pr996-current run ... \
    --attest '<isolated-home>/Library/Application Support/crabbox/attest/id_ed25519.pem' -- /usr/bin/true
attest receipt and attest key paths must be different
exit status: 2
```

The collision attempt exits during local preflight: there is no target acquisition or remote command output. The independently computed hash exactly matches the receipt, and the verifier makes the self-signed trust boundary explicit.

## Contributor real behavior proof

Behavior addressed: run evidence in Crabbox cannot be verified after the fact; there is no signed receipt and no verify command, so any pasted run result is trust-me.
Real environment tested: the real `crabbox` binary built from pristine main d2ad413b101148400232c4f048da1227f9541ce0 and from this branch at head 7ec4e381, driving a real `crabbox run` over loopback SSH with the static ssh provider (real sshd, real sync, real remote command execution) on macOS; nothing simulated.
Exact steps or command run after this patch: `crabbox run --static-host 127.0.0.1 --static-user yuvald --attest receipts/nested/receipt.json -- /bin/sh -c 'echo attest demo; uname -s'` (a nested receipt path whose directories do not exist yet), then `crabbox verify` on the untouched receipt, after a one-field tamper, after a duplicate-key tamper that keeps the signed value last, and after injecting an unknown field.
Evidence after fix at contributor head `7ec4e381` (current output additionally includes `trust=self-signed`):
```
# BEFORE (pristine main d2ad413b101148400232c4f048da1227f9541ce0)
$ crabbox verify receipt.json
unexpected argument verify
exit status: 2
$ crabbox run --help | grep -E "^  -attest"
(no output; exit status 1: no such flags)

# AFTER (this branch, real run over loopback SSH with the static ssh provider)
$ crabbox run --static-host 127.0.0.1 --static-user yuvald --attest receipts/nested/receipt.json -- /bin/sh -c 'echo attest demo; uname -s'
running on 127.0.0.1 /bin/sh -c 'echo attest demo; uname -s'
attest demo
Darwin
artifact kind=receipt path=receipts/nested/receipt.json bytes=503
$ ls -la receipts/nested/receipt.json   (parent directories created by --attest)
-rw-------@ receipts/nested/receipt.json
$ cat receipts/nested/receipt.json
{
  "command": "/bin/sh -c 'echo attest demo; uname -s'",
  "command_ms": 640,
  "exit_code": 0,
  "generated_at": "2026-07-07T00:01:44Z",
  "lease_id": "static_127-0-0-1",
  "log_sha256": "sha256:32470cf9166f8353de6d7ffebf19d66a16ab9a0d34c2cad840ac30dda4865c21",
  "provider": "ssh",
  "public_key": "npCheZsJNUrNHucD7VmntRwVYel4vpiuprF6uRn6T9g=",
  "schema_version": 1,
  "signature": "fsbj5E/O1yi4H1QhxyKDzeKCsvq8Nj7oVLtOHeVhNHctaf2fKBfidiOJVvq3SSKJgrKacUkuKQeAK8YL/nhABA==",
  "slug": "127-0-0-1"
}
$ crabbox verify receipts/nested/receipt.json
PASS receipts/nested/receipt.json signer=sha256:c30e2ae902253e3191a54aa8b78da9170e9a3630820c0b97c6142f475fe81295
exit status: 0
$ printf 'attest demo
Darwin
' | shasum -a 256
32470cf9166f8353de6d7ffebf19d66a16ab9a0d34c2cad840ac30dda4865c21  -
$ sed -i '' 's/"exit_code": 0/"exit_code": 1/' receipts/nested/receipt.json && crabbox verify receipts/nested/receipt.json
FAIL receipts/nested/receipt.json signer=sha256:c30e2ae902253e3191a54aa8b78da9170e9a3630820c0b97c6142f475fe81295: signature mismatch
exit status: 1
$ duplicate exit_code key inserted with the signed value last, then crabbox verify
malformed receipt: duplicate key
exit status: 2
$ unknown field review_note injected, then crabbox verify
malformed receipt: unknown field "review_note"
exit status: 2
```
Observed result after fix: a real run over loopback SSH created the missing receipt directories with private modes and produced a signed receipt whose output digest independently matches the observed live output; verifying the untouched receipt prints PASS with the signer fingerprint and exit 0, flipping a single field flips the same command to FAIL signature mismatch with exit 1, and both a duplicate-key edit that keeps the signed value last and an injected unknown field are rejected outright as malformed with exit 2 instead of printing PASS, where pristine main rejects the verify command entirely and has no attest flags.
What was not tested: no delegated-provider live run (the delegated receipt path is exercised in the Go suite only); no Windows host; key rotation and multi-user trust decisions are out of scope for v1.

## Not in this change (deliberate v2 boundary)

A v1 receipt is client-side, self-signed evidence: it proves the record is unmodified since signing and was signed by the embedded key, whose fingerprint `verify` prints for out-of-band comparison. It does not prove who holds that key (no trust roots or identity binding), that the run occurred as described (the signer is the party that ran it), or that a receipt was not withheld and regenerated (no transparency log). Coordinator key custody, DSSE or in-toto envelope formats, trust roots, and a remote receipt registry are deliberately out of scope for this slice.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
+- Added provider-neutral Ed25519-signed run receipts through `crabbox run --attest` and integrity verification through `crabbox verify`, with collision-safe signing-key handling and explicit self-signed trust reporting. Thanks @yetval.
 - Documented a provider-neutral hermetic-agent evidence pattern with separate writer contexts, QA arbitration, required proof artifacts, and sync-safe local downloads. Thanks @zozo123.
 - Added `sync-plan --json` with candidate and dirty-delta sizes, configured guardrail status, deleted-path counts, and ranked file and directory hotspots for automation. Thanks @zozo123.
 - Added a CubeSandbox delegated-run provider with E2B-compatible lifecycle and envd execution, archive sync, CubeProxy routing, exact API-endpoint/sandbox-bound ownership claims, conflict-safe explicit adoption, and guarded cleanup. Thanks @zozo123.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Added `sync-plan --json` with candidate and dirty-delta sizes, configured guardrail status, deleted-path counts, and ranked file and directory hotspots for automation. Thanks @zozo123.
 - Added a CubeSandbox delegated-run provider with E2B-compatible lifecycle and envd execution, archive sync, CubeProxy routing, exact API-endpoint/sandbox-bound ownership claims, conflict-safe explicit adoption, and guarded cleanup. Thanks @zozo123.
 - Added coordinator-managed Daytona Linux leases with a Worker-held API key, exact ownership cleanup, expiring SSH-token refresh, CLI secret redaction, and production Cloudflare configuration.
+- Added signed run receipts: `crabbox run --attest <path>` writes an Ed25519-signed receipt of a successful run, and the new `crabbox verify` command checks receipt integrity and reports the signer fingerprint. Thanks @yetval.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,6 @@
 - Added `sync-plan --json` with candidate and dirty-delta sizes, configured guardrail status, deleted-path counts, and ranked file and directory hotspots for automation. Thanks @zozo123.
 - Added a CubeSandbox delegated-run provider with E2B-compatible lifecycle and envd execution, archive sync, CubeProxy routing, exact API-endpoint/sandbox-bound ownership claims, conflict-safe explicit adoption, and guarded cleanup. Thanks @zozo123.
 - Added coordinator-managed Daytona Linux leases with a Worker-held API key, exact ownership cleanup, expiring SSH-token refresh, CLI secret redaction, and production Cloudflare configuration.
-- Added signed run receipts: `crabbox run --attest <path>` writes an Ed25519-signed receipt of a successful run, and the new `crabbox verify` command checks receipt integrity and reports the signer fingerprint. Thanks @yetval.
 
 ### Fixed
 
--- a/docs/commands/README.md
+++ b/docs/commands/README.md
@@ -30,6 +30,7 @@ same order as the CLI help.
 - [events](events.md)
 - [attach](attach.md)
 - [results](results.md)
+- [verify](verify.md)
 - [cache](cache.md)
 - [status](status.md)
 - [list](list.md)
--- a/docs/commands/run.md
+++ b/docs/commands/run.md
@@ -344,8 +344,9 @@ Use `--attest <path>` to write a signed run receipt after a successful run: a
 flat JSON record of the provider, lease, command, exit code, timing, and the
 SHA-256 of the combined live output stream, signed with a per-user Ed25519 key
 minted on first use under the user config dir. Check receipts later with
-[`crabbox verify`](verify.md). Pass `--attest-key <path>` to sign with an
-existing PKCS8 PEM Ed25519 key instead of the default one.
+[`crabbox verify`](verify.md), then compare the reported signer fingerprint
+through a trusted channel. Pass `--attest-key <path>` to sign with an existing
+PKCS8 PEM Ed25519 key instead of the default one.
 
 ## Artifacts and downloads
 
--- a/docs/commands/run.md
+++ b/docs/commands/run.md
@@ -340,6 +340,13 @@ selected live console output, collected artifact paths, and the
 `--proof-template` or preset template. Keep proof templates in repo config so
 parser-sensitive PR wording stays project-owned.
 
+Use `--attest <path>` to write a signed run receipt after a successful run: a
+flat JSON record of the provider, lease, command, exit code, timing, and the
+SHA-256 of the combined live output stream, signed with a per-user Ed25519 key
+minted on first use under the user config dir. Check receipts later with
+[`crabbox verify`](verify.md). Pass `--attest-key <path>` to sign with an
+existing PKCS8 PEM Ed25519 key instead of the default one.
+
 ## Artifacts and downloads
 
 Use repeatable `--artifact-glob <glob>` to collect matching remote files after a
@@ -553,6 +560,8 @@ Run-specific flags:
 --preset-var name=value      Repeatable or comma-separated.
 --emit-proof <path>
 --proof-template <name>
+--attest <path>
+--attest-key <path>
 --preflight
 --preflight-tools <comma-separated tool names>
 --junit <comma-separated remote XML paths>
--- a/docs/commands/verify.md
+++ b/docs/commands/verify.md
@@ -3,7 +3,8 @@
 `crabbox verify` checks a signed run receipt produced by `crabbox run --attest`.
 It recomputes the receipt's canonical bytes, verifies the embedded Ed25519
 signature, and prints the signer's public key fingerprint, so a reviewer can
-confirm a pasted receipt has not been edited after the run.
+confirm a pasted receipt has not been edited after the run once they have
+matched the fingerprint through a trusted channel.
 
 ```sh
 crabbox run --attest receipt.json -- go test ./...
@@ -13,14 +14,14 @@ crabbox verify receipt.json
 On success it prints one line and exits 0:
 
 ```
-PASS receipt.json signer=sha256:6a5f...
+PASS receipt.json signer=sha256:6a5f... trust=self-signed
 ```
 
 If any signed field was modified after signing, it prints a failure line and
 exits 1:
 
 ```
-FAIL receipt.json signer=sha256:6a5f...: signature mismatch
+FAIL receipt.json signer=sha256:6a5f... trust=self-signed: signature mismatch
 ```
 
 A receipt that cannot be checked at all (unreadable file, invalid JSON,
@@ -35,7 +36,7 @@ The receipt is a flat JSON object. `schema_version`, `generated_at`,
 `provider`, `command`, `exit_code`, `command_ms`, and `public_key` are always
 present; `lease_id`, `slug`, `run_id`, `actions_url`, and `log_sha256` appear
 when the run produced them. `log_sha256` is the SHA-256 of the combined
-stdout and stderr stream as observed by the client during brokered SSH runs.
+stdout and stderr stream as observed by the client during SSH-backed runs.
 
 The signature covers the canonical encoding of every field except `signature`
 itself: the object is re-marshaled as compact JSON with lexicographically
@@ -52,7 +53,9 @@ PKCS8 PEM Ed25519 key instead; the override path is never auto-created.
 
 Verification trusts the key embedded in the receipt and reports its
 fingerprint. Binding that fingerprint to an identity is out of scope: compare
-it out of band with the signer.
+it out of band with the signer. The `trust=self-signed` marker is printed on
+every verification result so `PASS` cannot be mistaken for an identity or
+execution-provenance check.
 
 ## Non-goals
 
--- a/docs/commands/verify.md
+++ b/docs/commands/verify.md
@@ -1,0 +1,59 @@
+# verify
+
+`crabbox verify` checks a signed run receipt produced by `crabbox run --attest`.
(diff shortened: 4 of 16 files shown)


yetval:
All four review items are addressed in https://github.com/openclaw/crabbox/pull/996/commits/83070b1d and the PR body proof is refreshed against that head.

- [P1] Duplicate keys: `verify` now rejects receipts whose JSON contains duplicate object keys at any depth, before any signature decision, with `malformed receipt: duplicate key` and exit 2. The tamper table gained two regression cases where the signed value stays last (duplicate `exit_code`, duplicate `command`), and the live proof in the body shows the same edit against a real receipt now exiting 2 instead of printing PASS.
- [P2] Output collisions: `--attest` is preflighted against `--lease-output`, `--emit-proof`, `--capture-stdout`, `--capture-stderr`, and every `--download` destination via a shared labeled collision check that `--emit-proof` now also uses. Covered by a new collision regression test.
- [P2] Wrapper forwarding: `--attest` and `--attest-key` are forbidden in `watch` and `shard` forwarding, matching `--emit-proof`, with rows added to both existing forbidden-flag suites.
- The CHANGELOG edit is removed; that file is byte-identical to the branch point.

The v1 trust boundary question (self-signed, no trust roots, no transparency log) is stated in the body as a deliberate v2 boundary and stays with maintainers.

yetval:
@clawsweeper re-run

yetval:
The digest race is fixed in https://github.com/openclaw/crabbox/pull/996/commits/b7ada56f and the body proof is refreshed against that head.

- [P1] Serialization: the receipt digest is now a mutex-guarded writer shared by the stdout and stderr fan-outs, the same synchronization the run log buffer those streams already share uses. `log_sha256` is defined as the hash of the combined stream bytes in arrival order.
- Regression coverage runs under the race detector: one case drives concurrent mixed stdout/stderr writes through the shared digest and asserts the exact expected hash, and a sequential mixed-stream case pins arrival-order hashing. As validation of the finding, the previous shape (a bare shared hash behind both MultiWriters) fails `go test -race` immediately with WARNING: DATA RACE.
- Maintainer confirmation of the self-signed v1 boundary remains open, as documented in the body and the verify command page.

yetval:
@clawsweeper re-run

yetval:
@clawsweeper re-review

Both remaining findings are addressed and the body proof is regenerated against head https://github.com/openclaw/crabbox/pull/996/commits/7ec4e381c548ca56f6089c3ebd362382527d638d.

- Nested receipt paths: writeRunReceipt now creates missing parent directories with the same private mode proof output uses, before the atomic write. The live proof drives `--attest receipts/nested/receipt.json` end to end and shows the created 0700/0600 modes; a regression covers the nested path and directory mode.
- Delegated receipts without a LeaseID: lease_id is now treated like the other run-dependent fields (omitted when a delegated provider reports none, validated only when present), and the delegated block builds receipts through a writeDelegatedRunReceipt helper mirroring writeDelegatedRunProof. A regression covers a delegated provider returning successful run metadata with no LeaseID and asserts the receipt round trips to PASS with the empty fields omitted.

Also folded in since the last review: strict receipt schema validation before any signature decision (unknown fields, missing required fields, wrong types, non-integer spellings, unsupported schema versions, invalid timestamps and digests all exit 2), with the unknown-field rejection shown live in the proof block.

steipete:
@clawsweeper re-review

steipete:
@clawsweeper re-review

Exact current-head behavior proof is now in the PR body for `c28da3ce2718b03c6cdbceb0e698e1e12250e154`: real loopback SSH execution, independent output-digest match, `trust=self-signed` verification, and receipt/signing-key collision rejection before target acquisition. The maintainer review also remains clean after four autoreview rounds.

steipete:
Maintainer decision: accept the documented opt-in, self-signed v1 boundary as integrity evidence only. `PASS` is not identity binding or independent execution provenance; callers must establish the signer fingerprint through a trusted channel. Exact-head live proof, full CI, and maintainer review are complete, so this is ready to land.

PR #983, opened 2026-07-06
feat(external): support macOS desktop credentials

MuduiClaw:
Replaces https://github.com/openclaw/crabbox/pull/976 after its stack base merged into main.

## Summary

- allow the generic `external` provider to preserve Linux, macOS, native Windows, and WSL2 targets, including ARM64 hosts
- persist target and Windows-mode routing so later lease commands restore the captured platform without overriding explicit target flags
- add source-bound external desktop credentials via username plus password environment-variable name
- use provider credentials consistently for WebVNC, screenshots, input, native VNC handoff, daemon children, and routed leases
- authenticate macOS ARD or legacy VNC locally, then expose an already-authenticated RFB stream without sending the desktop password to the coordinator
- add authenticated `webvnc --preflight`, including RFB 3.3, 3.7, 3.8, and accepted alias-banner handling
- validate native-Windows work roots against drive roots, user profiles, protected directories, traversal/ambiguous components, short names, ADS, and reserved device aliases
- document macOS credentials, native-Windows work-root requirements, preflight, and routing behavior

## Maintainer hardening

- merged current `main` and resolved the WebVNC/RFB stack against the one-time portal handoff architecture
- preserved legacy Parallels password-only VNC while using ARD accounts for provider-supplied macOS credentials
- kept registered static macOS endpoints host-managed and preserved two macOS WebVNC slots for controller plus observer
- made a configured-but-missing desktop password environment variable fail closed instead of falling back to an unrelated remote password file
- scrubbed desktop credentials from every unrelated provider, adapter, SSH, shell, and controller child while keeping the one credential-owning daemon on stdin only
- made persisted credential routes fail closed on malformed state, symlinks/reparse points, insecure ownership or permissions, and same-path replacement; generated children are bound to the exact route digest and cleanup to a unique generation
- preserved explicit, persisted, and server-labelled work roots even when they equal another platform's default
- made server-authenticated macOS portal viewers credential-free, with clean ACL-controlled share URLs and correct RFB 3.7 `None` negotiation
- made foreground SSH tunnel cleanup terminate and verify the full ProxyCommand process tree, including when the SSH leader already exited
- made canceled controller children wait for verified process-group exit before removing their durable recovery handle
- resolved registered macOS desktop credentials before scrubbing the daemon environment and forwarded them only through the one-shot supervisor stdin gate
- kept `webvnc status` usable without loading desktop secrets
- added a changelog entry thanking @MuduiClaw

## Verification

[...]

PR #1043, opened 2026-07-10
fix(webvnc): preserve reset credentials (KIN-1161)

MuduiClaw:
Refs KIN-1161

## Summary

- preserve the external macOS desktop credential input when `webvnc reset` restarts the local bridge daemon
- keep the provider-routed desktop username and password environment reference in daemon arguments
- continue sending the password only through the existing one-shot supervisor stdin gate
- fail closed with exit code 2 when the configured desktop credential is unavailable

## Root cause

The reset path resolved the provider desktop credentials before mutating the portal, daemon, or remote desktop state, but then restarted the daemon with a nil credential input. The daemon environment had already been scrubbed, so the replacement bridge could not recover the configured secret.

## Validation

- `go test ./internal/cli -run 'Test(WebVNCResetDaemonLaunch|RegisteredWebVNCDaemonCredentialInput)' -count=1 -timeout=2m`
- `go test ./internal/cli -run 'TestWebVNCReset' -count=1 -timeout=2m`
- `go vet ./internal/cli`
- local autoreview: no accepted or actionable findings

## Security

The regression fixtures use synthetic values only. Credential values remain absent from argv and the daemon child environment, and the tests do not contain viewer URLs, tickets, cookies, signed URLs, or real credentials.


MuduiClaw:
Closing because the PR was opened against the wrong base. Commit 75b64da is the head of open PR #983, not openclaw/main, so this comparison included the full KIN-1161 provider stack. The reset-only delta is being reopened against MuduiClaw:codex/kin-1161-macfleet-provider.

PR #1071, opened 2026-07-13
feat: validate required run artifacts against a schema (`--require-artifact-schema`)

### LINKED ISSUE #1070: Validate required run artifacts against a schema (`--require-artifact-schema`)
## Problem

Crabbox's strongest post-run evidence gate today is `--require-artifact`, but by its own documentation it is *"an existence guard, not manifest validation or a data safety scanner"* (`docs/features/artifacts.md`). The check runs a remote command that only asks *"does this file exist?"* and returns `exit 7` when it is missing (`internal/cli/run_artifacts.go`).

That means a run **passes its proof gate even when the proof is meaningless**: an empty `{}`, a truncated write, the wrong shape, or a JSON object that is missing the very fields that make it evidence. For a tool whose contract is *reproducible, reviewable run evidence* (hermetic-agent proofs, test-result summaries, QA verdicts, manifests), "the file is present" is a weak guarantee. The interesting assertion is "the file is **valid**."

`docs/features/hermetic-agent-evidence.md` names this gap explicitly:

> Crabbox can require that the proof file exists; it does not validate the proof schema unless the repo command does so.

## Use case

A repository wants CI to fail when its evidence file is present but malformed:

```sh
crabbox run \
  --require-artifact-schema docs/metrics/hermetic-agents-e2e.json=schemas/hermetic.schema.json \
  -- ./scripts/run_hermetic_agents_demo.sh
```

If `hermetic-agents-e2e.json` is missing, unparseable, or missing required fields (`code_writer`, `test_writer`, `qa_verdict`), the run fails the gate with `exit 7` and a precise list of what was wrong — instead of silently "passing" because a zero-byte file happened to be there.

## Proposed shape (narrow, testable MVP)

A new **post-run gate** that validates a required artifact's JSON *content* against a declared schema. It is the existing existence gate, one notch deeper: present → **valid**.

### CLI surface

```
--require-artifact-schema REMOTE=SCHEMA    # repeatable
```

- `REMOTE` — path (relative to the run workdir) of the artifact to validate, reusing the same safe-path validation as `--require-artifact`.
- `SCHEMA` — a **local** schema file, read and parsed at preflight.
- Validating a file implies it must exist and parse, so the flag composes with (or stands in for) `--require-artifact` for that path.

### Schema language

A small, **dependency-free subset of JSON Schema** — the forward-compatible first slice of full JSON Schema:

| Keyword | Meaning |
| --- | --- |
| `type` | `object` / `array` / `string` / `number` / `integer` / `boolean` / `null` |
| `required` | required object properties |
| `properties` | per-property sub-schemas (recursed) |
| `items` | sub-schema applied to every array element (recursed) |
| `enum` | value must equal one of the listed JSON values |

The subset is **fail-closed**: a schema containing any validation keyword the validator cannot enforce (`pattern`, `minimum`, `additionalProperties`, `anyOf`, `$ref`, …) is rejected at preflight with `exit 2`, so a passing gate always means the supplied constraints were actually enforced — never silently skipped. Only annotation keywords (`$schema`, `$id`, `title`, `description`, `$comment`, `examples`, `default`, `deprecated`) are accepted and ignored. The supported subset can be widened deliberately over time. No new Go module dependency is added.

### Behaviour and exit codes

Validation happens **locally** on the artifact's bytes (fetched to memory via the existing base64 read path, never written to the user's tree) — which keeps the logic in Go, provider-agnostic, and unit-testable.

| Situation | Result |
| --- | --- |
| Command itself fails | command's own nonzero exit (unchanged) |
| Malformed / unreadable / unsupported-keyword **schema file** | `exit 2` (bad invocation, at preflight) |
| Required artifact missing / unfetchable | `exit 7` (gate failure) |
| Required artifact larger than the cap (5 MiB) | `exit 7` (gate failure, not read into memory) |
| Artifact present but content violates schema | `exit 7` (gate failure, with violations listed) |
| All schemas satisfied | run proceeds normally |

The gate runs only after command success, alongside the existing `--require-artifact` gate, and reuses its `exit 7` class.

### Evidence

Each result is recorded on the timing report under a new `omitempty` field so existing timing JSON is byte-identical when the flag is absent:

```json
{
  "schemaValidations": [
    {
      "artifact": "docs/metrics/hermetic-agents-e2e.json",
      "schema": "schemas/hermetic.schema.json",
      "valid": false,
      "violations": [
        "qa_verdict: missing required property \"qa_verdict\"",
        "count: expected type number, got string"
      ]
    }
  ]
}
```

## Provider scope

Phase 1 targets **SSH-backed Linux/Windows providers**, exactly like the existing artifact gate. Delegated-run providers reject the flag at preflight with `exit 2` ("not supported for provider=X yet"), mirroring how `--require-artifact` collection and other artifact features are already gated behind an explicit delegated capability. Delegated support is a documented follow-up.

## What is reuse vs. genuinely new

**Reuse:** the `stringListFlag` pattern, `validateRequiredRunArtifactGlobs` safe-path validation, the `runSSHOutput` + `remoteDownloadBase64Command` read path, the post-run gate ordering, the `exit 7` gate class, and the `TimingReport` JSON writer.

**New (bounded):** one `--require-artifact-schema` flag + preflight loader, one dependency-free JSON-Schema-subset validator (`internal/cli/run_artifact_schema.go`), one `TimingReport.SchemaValidations` field, and the delegated guard.

## Non-goals

- Not a data-safety / secrets scanner (that stays a separate concern).
- Not remote validation (validation is local Go).
- Not a replacement for JUnit/`results` parsing.
- Not full JSON Schema in v1 (documented forward-compatible subset only).

## Testing

The valuable core is a pure function `validateJSONAgainstSchema(doc []byte, schema) []violation` with **no lease, network, provider, or cloud** — a classic table test:

- valid document → no violations;
- missing required field → violation at the right path;
- wrong type / `enum` mismatch → violation;
- nested object and `items[i]` paths reported correctly;
- non-JSON / empty document → single `json` violation (not a crash);
- malformed **schema** → `exit 2`, distinct from `exit 7`;
- spec parsing: valid `REMOTE=SCHEMA`, missing `=`, unsafe remote path.

## Resolved decisions

- **Unsupported keywords: fail closed.** Rather than claim general JSON Schema compatibility and silently skip constraints, the parser rejects any unsupported validation keyword at preflight. A passing gate reliably means the supplied schema was enforced.
- **Fetched artifacts are bounded.** The remote read is capped at 5 MiB (reads one byte past the limit to detect and reject oversized artifacts), bounding SSH output and local memory.

## Open questions

1. Should `--require-artifact-schema REMOTE` implicitly add `REMOTE` to the `--require-artifact` set, or is the fetch's own existence failure enough? (MVP: the fetch failure covers existence — no implicit add.)
2. Should the 5 MiB artifact cap be configurable per run?
3. Delegated capability contract for a later phase.

dwin-gharibi:
Closes https://github.com/openclaw/crabbox/issues/1070

## Summary

Adds `--require-artifact-schema remote=schema.json`, a post-command evidence gate that validates a required JSON artifact instead of checking existence alone.

- Supports standard JSON Schema drafts 4, 6, 7, 2019-09, and 2020-12.
- Supports bounded local static references and embedded resources; external fetches and active recursive or dynamic references fail closed.
- Returns exit 2 for malformed or unsafe schemas and exit 7 for missing, oversized, malformed, or non-conforming artifacts.
- Emits bounded, value-redacted JSON Pointer diagnostics and records results in timing evidence.
- Documents schema, artifact, reference, regular-expression, work, and diagnostic limits.

## Safety and dependency notes

Schema definitions are capped at 1 MiB and artifacts at 5 MiB. JSON depth, value count, numeric work, reference expansion, validation work, regular-expression source/program work, and diagnostic output all have explicit limits. The regular-expression engine uses Go's linear-time RE2 implementation through a fail-closed ECMA-262 compatibility layer.

The implementation uses `github.com/santhosh-tekuri/jsonschema/v6` plus `github.com/dlclark/regexp2/syntax`. A compatibility correction for mixed legacy `dependencies` resources is proposed upstream at https://github.com/santhosh-tekuri/jsonschema/pull/262; this PR temporarily replaces the module with the public commit https://github.com/steipete/jsonschema/commit/7235d3ee96420a175998445eaf4f864f0aef5faf. The replacement is source-compatible and can be removed once upstream ships the fix. Both dependencies are active, non-archived projects; no credentials or new runtime configuration are introduced.

## Verification

- `go test -race ./...`
- `go vet ./...`
- `go mod verify`
- Worker format, lint, typecheck, build, and 1,059 tests
- Docs-site build
- Source-blind local-container behavior contract: matching artifact exit 0; invalid and missing artifacts exit 7; malformed schema exit 2 before provider access; artifact value absent from validation diagnostics
- Exact-head AWS SSH behavior matrix: matching artifact exit 0; type mismatch exit 7; missing artifact exit 7; 5 MiB plus one byte exit 7; malformed schema exit 2 before run creation; lease released afterward
- Structured code review: clean, no actionable findings
- Exact-head hosted CI at `b6a830872ecb262ec4259dd1d2c32957e1c8ae5c`: all jobs green, including Go race/coverage/build, Worker, release snapshot, Apple VM, Windows cancellation, scripts, connector lifecycles, docs, Docs UI Proof, and Connector E2E Smokes — https://github.com/openclaw/crabbox/actions/runs/29575888071

## Contributor credit

Preserves Dwin Gharibi's authorship with co-authored maintainer commits and the changelog credit to @dwin-gharibi.


dwin-gharibi:
@clawsweeper re-review

steipete:
Maintainer improvement pass pushed to the contributor branch at `3ff3ca87`.

Improvements:

- Made schema parsing genuinely fail-closed: exactly one JSON object, no trailing content, duplicate object names, null/invalid keyword shapes, duplicate `required` entries, duplicate `enum` values, or unsupported keywords.
- Preserved exact JSON number semantics, including values beyond `float64` precision and compact numbers with very large exponents.
- Removed rejected artifact values from diagnostics, capped retained/printed violations, and added an explicit truncation marker.
- Fixed native Windows short-read handling by reading until EOF or the byte limit.
- Restricted the artifact operand to one exact safe relative path, including rejecting glob and Windows drive/ADS syntax.
- Added regression coverage for all of the above while preserving the provider-neutral core/SSH-target boundary.

Proof: `go build ./...`, `go vet ./...`, focused tests, focused race tests, and docs-site generation pass. The full race suite passed the changed CLI package and all relevant packages; one unrelated timing-sensitive external-provider test missed its bound under suite load and passed immediately when retried alone with race.

Final Codex autoreview command: `.agents/skills/autoreview/scripts/autoreview --mode branch --base origin/main`. Verdict: clean, no accepted/actionable findings, overall confidence 0.91.

**Remaining gate — do not merge:** a maintainer product decision is still required on whether Crabbox should own this bespoke JSON-Schema-subset contract at all, versus narrowing the feature or using a maintained standards-compatible approach. If the feature is retained, it still needs redacted real SSH-backed CLI proof for valid, invalid, missing, oversized, and malformed-schema cases. This fork also needs maintainer approval before hosted workflows can run.

steipete:
Maintainer correctness pass pushed to the contributor branch at `d04a148c` (after merging current `origin/main`).

Two reproducible P2 fail-closed bugs found by live testing are fixed:

1. Schema keyword matching was case-insensitive through Go's JSON struct decoder. A later `"REQUIRED": []` could alias and overwrite a real `"required": ["proof"]` constraint. Schema nodes now use an exact, case-sensitive keyword allowlist. Unknown or mis-cased keywords fail preflight with exit 2, consistent with this tool's documented fail-closed subset.
2. Go's JSON decoder replaces invalid UTF-8 with U+FFFD. Raw schema and artifact bytes are now checked before decoding. Invalid UTF-8 schema input fails preflight with exit 2; invalid UTF-8 artifact input fails validation with exit 7.

Regression provenance:

```text
BEFORE
mis-cased required: expected invalid schema keyword shape, got no error
invalid UTF-8 schema: expected exit 2, got nil
invalid UTF-8 artifact: expected exit 7, got nil

AFTER
go test ./internal/cli -run 'TestParseArtifactSchemaRejectsInvalidKeywordShapes|TestLoadRequireArtifactSchemas|TestValidateArtifactSchemasWithReaderBehaviour' -count=1
ok github.com/openclaw/crabbox/internal/cli
```

Real CLI proof through a localhost SSH target, using the branch-built binary:

```text
valid             -> 0  schema valid.json: ok
violation         -> 7  schema violation.json: failed 1 check
missing           -> 7  schema missing.json: fetch failed
unparseable       -> 7  schema unparseable.json: failed 1 check
malformed-schema  -> 2  invalid schema: EOF
oversized         -> 7  artifact exceeds the 5242880-byte validation limit
```

Verification clean:

```text
go vet ./...
go test ./internal/cli -run 'ArtifactSchema|ValidateJSONAgainstSchema|DecodeBoundedBase64|RemoteBoundedReadBase64' -count=1
.agents/skills/autoreview/scripts/autoreview --mode branch --base origin/main
autoreview clean: no accepted/actionable findings reported (confidence 0.92)
```

Technically merge-ready. Still **do not merge** until the maintainer decides whether Crabbox should own this bespoke JSON-Schema-subset contract, use a maintained standards-compatible validator, or keep validation repo-owned. This pass intentionally does not make that product decision.

PR #1084, opened 2026-07-14
fix: make published releases immutable

steipete:
## What Problem This Solves

Fixes an issue where published release assets remained mutable and release operators could not prove the Homebrew formula on both native Apple Silicon and Intel hosts from one protected, repeatable workflow.

## Why This Change Was Made

Requires organization-enforced release immutability before the publication mutation, rejects mutable public records in every verifier, and adds a protected-default-branch workflow that binds the immutable release to its exact successful public verifier run. The workflow checks out only that verifier commit and runs the credential-free Homebrew verifier on native arm64 and x86_64 runners; invalid refs fail instead of appearing skipped-successful.

## User Impact

No CLI behavior changes. Release operators get tamper-resistant public assets and a reproducible two-architecture Homebrew closeout gate.

## Evidence

- `actionlint .github/workflows/verify-homebrew.yml`
- Fresh Codex autoreview: clean after fixing invalid-ref skip behavior and requiring GitHub release immutability.
- `node --test scripts/publish-release.test.js`: 17/17 passing, including disabled and repository-only immutability failures before mutation.
- Crabbox v0.38.1 public verifier run `29327262104` is the pinned proof source used to exercise the workflow after landing; v0.38.2 will be the first immutable release accepted by the gate.

### DIFF
--- a/.github/workflows/release-assets.yml
+++ b/.github/workflows/release-assets.yml
@@ -183,6 +183,7 @@ jobs:
             release.tag_name === process.env.RELEASE_TAG &&
             release.name === process.env.RELEASE_TAG &&
             release.draft === (process.env.EXPECTED_DRAFT === 'true') &&
+            release.immutable === (process.env.EXPECTED_DRAFT !== 'true') &&
             release.prerelease === false &&
             (release.draft ? release.published_at === null : typeof release.published_at === 'string') &&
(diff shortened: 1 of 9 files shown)


steipete:
@codex review

steipete:
@codex review

PR #1063, opened 2026-07-12
fix(cli): honor context cancel in provider IP wait loops

SebTardif:
## What Problem This Solves

Hetzner `waitForServerIP`, AWS `waitForServerIP`, DigitalOcean `waitForDropletIP`, and `waitForLoopbackVNC` waited between polls with bare `time.Sleep`, which ignores `ctx.Done()`. GCP and Azure already use `select` on context vs timer. A cancelled acquire/bootstrap could sit on the sleep for the full interval (3–5s, or 2s for VNC) even after the parent context was cancelled.

## Evidence

Sibling implementations already correct:

- `internal/cli/gcp.go` `WaitForServerIP` uses `select { case <-ctx.Done(); case <-time.After(...) }`
- `internal/cli/azure.go` same pattern
- Proxmox uses ticker + `deadlineCtx.Done()`

Before this change, Hetzner/AWS/DO used `time.Sleep` with no select.

### Unit tests

```bash
go test ./internal/cli/ -run TestSleepContext -count=1
go test ./internal/providers/digitalocean/ -run TestSleepContext -count=1
```

`TestSleepContextHonorsCancellation` fails to return promptly if sleep ignores context (hangs until the delay); with `sleepContext` it returns `context.Canceled` immediately.

### Related

- https://github.com/openclaw/crabbox/pull/1053 (AWS bootstrap host selection)
- Pattern mirror of GCP/Azure IP wait loops in-tree

## Real behavior proof

- **Behavior or issue addressed:** Provider IP wait loops (Hetzner/AWS/DigitalOcean) and VNC wait used bare `time.Sleep`, so context cancel during the delay was ignored until the full sleep elapsed.
- **Real environment tested:** macOS arm64, Go toolchain, branch `fix/wait-ip-context-cancel` at worktree `/private/tmp/crabbox-ip-wait`.
- **Exact steps or command run after this patch:**

  ```bash
  cd /private/tmp/crabbox-ip-wait
  go test ./internal/cli/ ./internal/providers/digitalocean/ -count=1 -run 'SleepContext|Sleep'
  go run /tmp/crabbox-cancel-demo.go   # waitForIP-shaped loop: 3s poll, cancel after 50ms
  ```

- **Evidence after fix:** terminal output from the patched build:

  ```console
  === CONT  TestSleepContextHonorsCancellation
  --- PASS: TestSleepContextHonorsCancellation (0.00s)
  PASS
  ok  github.com/openclaw/crabbox/internal/cli
[...]

PR #1092, opened 2026-07-14
fix(cli): honor context cancel during SSH readiness backoff

SebTardif:
## What problem this solves

`waitForSSHReady` is the shared SSH readiness loop for bootstrap, sync, and command paths. It checked cancellation at loop entry but used a bare ten-second sleep between failed probes, delaying cancellation by the full backoff.

## Improved fix

- replace the bare retry sleep with the existing context-aware helper
- preserve `context.Cause`, so lease-watch failures are not downgraded to generic `context.Canceled`
- synchronize the regression test on the real failed-probe progress boundary instead of sleeping for an assumed 200ms
- retain contributor credit in the 0.38.4 Unreleased changelog

## Verification

- `go test -race ./internal/cli -run 'TestWaitForSSHReadyPreservesCancellationCauseDuringBackoff' -count=20`
- `go test -race ./internal/cli -count=1` (193.850s)
- `go vet ./...`
- built-CLI, source-blind behavior contract: a real `provider=ssh` run reached SSH readiness backoff, remained alive before cancellation, then exited nonzero 0.114s after SIGINT; no readiness timeout and zero isolated-home files
- full combined-diff autoreview: clean, confidence 0.96

## Risk

Shared SSH readiness only. Probe, success, deadline, and provider routing behavior are unchanged. Cancellation now returns promptly while preserving the exact existing cancellation cause.

## Related

- Same defect class: https://github.com/openclaw/crabbox/pull/1063
- Original loop provenance: `a52b45d45` (2026-04-30)

### DIFF
--- a/internal/cli/ssh.go
+++ b/internal/cli/ssh.go
@@ -163,7 +163,9 @@ func waitForSSHReady(ctx context.Context, target *SSHTarget, stderr io.Writer, p
 			lastPorts = strings.Join(probes, ",")
 			fmt.Fprintln(stderr, sshWaitProgressMessage(target, phase, reachablePort, transportPort, lastPorts, time.Since(start), time.Until(deadline)))
 		}
-		time.Sleep(10 * time.Second)
+		if err := sleepContext(ctx, 10*time.Second); err != nil {
+			return err
+		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


steipete:
Maintainer hardening pushed at exact head `e742bcf75557461e2022d014c45102a82115855d`.

The test now waits for the production progress boundary immediately before backoff, removing the blind 200ms delay. The production path also returns `context.Cause(ctx)`, preserving lease-watch failures instead of degrading them to generic cancellation.

Proof: focused race test 20x; full `internal/cli` race suite green in 193.850s; vet/build green; built CLI entered a real SSH readiness wait and exited nonzero 0.114s after SIGINT with no timeout or isolated-home residue; combined-diff autoreview clean (0.96).

@vincentkoc please provide the required independent last-push approval on this exact head. Auto-merge is armed for exact-head CI.

@clawsweeper re-review

PR #1091, opened 2026-07-14
chore(release): prepare 0.38.4

steipete:
Closes #1089.

PR #1090 is incorporated here; this is the sole v0.38.4 landing path.

## Summary

- restore protected Homebrew proof downloads with GitHub's supported Actions artifact media type
- align the release runbook and regression assertion with that request contract
- freeze the v0.38.4 changelog date
- align the private Worker package metadata with v0.38.4

## Verification

- `node --test scripts/release-workflow.test.js` (22 passed)
- full-branch autoreview clean on exact head `079ecbb84a0dfebe92eb7298ac5301806dc8c2de`
- both published v0.38.3 native proof ZIPs downloaded successfully with the corrected request shape; credential-free native arm64 Homebrew verification then passed end-to-end
- package and lockfile versions checked for exact equality
- exact-head CI, CodeQL, connector smokes, and Release Check required before merge

## Security and configuration

No secret, permission, dependency, or configuration changes. The protected workflow retains read-only Actions and contents permissions.

PR #1094, opened 2026-07-14
fix: prevent sparse Testbox sync corruption

### LINKED ISSUE #1093: Blacksmith Testbox reuse corrupts sparse worktree sync
## Summary

Reusing a Blacksmith Testbox from a sparse Git worktree can corrupt the remote checkout when Blacksmith's delegated sync advances from targeted manifest sync to its full rsync fallback. Paths omitted locally by sparse checkout are interpreted as deletions even though Git does not consider them deleted.

## Reproduction

1. Create a sparse worktree from a repository with thousands of tracked files.
2. Warm a `blacksmith-testbox` lease.
3. Run Crabbox once from the sparse worktree. The initial targeted sync correctly sends only the changed files.
4. Reuse the same lease for another run after the delegated sync threshold advances.
5. Inspect the remote checkout with `git status --short`.

## Observed

The reused sync falls back to a full rsync and produces thousands of phantom deletions for tracked paths outside the sparse checkout. In one real repository, 2,521 files were incorrectly marked deleted.

## Expected

Sparse-checkout omissions must never be treated as user deletions. Crabbox should either provide Blacksmith a materialized full checkout that preserves the sparse worktree's real changes, or fail closed before delegated sync can mutate the Testbox.

## Version

Confirmed with Crabbox 0.38.3. The same repository synced correctly when invoked from a materialized full checkout.

## Safety impact

This can make remote tests run against a checkout that differs radically from the user's local Git state. It is especially risky because the first run may succeed and the corruption appears only when reusing the lease.

steipete:
Closes https://github.com/openclaw/crabbox/issues/1093

Merge after https://github.com/openclaw/crabbox/pull/1091 so the 0.38.4 release remains serialized.

## What Problem This Solves

Fixes an issue where users running Blacksmith Testbox from a sparse Git checkout could receive a remotely corrupted workspace when a later delegated sync interpreted intentionally omitted tracked paths as deletions.

## Why This Change Was Made

Crabbox now detects Git's canonical sparse-checkout setting and fails before invoking Blacksmith's delegated sync. The provider documentation directs users to materialize a temporary full checkout, preserving the provider boundary while preventing an unsafe remote mutation.

## User Impact

Blacksmith Testbox runs from full checkouts are unchanged. Runs started from sparse checkouts now stop with an actionable error instead of risking thousands of phantom remote deletions.

## Evidence

- `go test -count=1 ./internal/cli ./internal/providers/blacksmith` — passed.
- `go test -count=1 -run TestIsSparseGitCheckout ./internal/cli` — passed.
- `go test -count=1 ./internal/providers/blacksmith` — passed.
- `git diff --check` — passed.
- Structured autoreview — clean twice; no accepted or actionable findings.
- Source-blind CLI validation — sparse fixture exited 2 with the full-checkout remediation and never invoked the provider stub; disabling sparse mode reached the stub and propagated its exit 99.

### DIFF
--- a/docs/providers/blacksmith-testbox.md
+++ b/docs/providers/blacksmith-testbox.md
@@ -47,6 +47,11 @@ crabbox status --provider blacksmith-testbox --id tbx_123
 crabbox stop --provider blacksmith-testbox tbx_123
 ```
 
+Run delegated sync from a full Git checkout. Crabbox rejects sparse checkouts
+before invoking Blacksmith because paths intentionally omitted by sparse checkout
+can otherwise be misread as deletions during a later full sync. Materialize a
+temporary full checkout when the source workspace must stay sparse.
+
 Keep a Testbox between runs via a JSON session handle:
 
 ```sh
--- a/internal/cli/repo.go
+++ b/internal/cli/repo.go
@@ -23,6 +23,19 @@ type Repo struct {
 	BaseRef   string
 }
 
+// IsSparseGitCheckout reports whether Git intentionally omits tracked paths
+// from the working tree. Delegated sync tools must not treat those omissions as
+// deletions from a complete checkout.
+func IsSparseGitCheckout(root string) bool {
+	if strings.TrimSpace(root) == "" {
+		return false
+	}
+	return strings.EqualFold(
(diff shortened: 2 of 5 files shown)


steipete:
@clawsweeper review

PR #1097, opened 2026-07-15
fix(worker): drain pipes before cmd.Wait to stop stdout truncation

anagnorisis2peripeteia:
fix(worker): own container-runner pipe ends to prevent stdout truncation and exec hangs

## Summary
Give the cloudflare-container-runner exec path **self-owned `os.Pipe()` write-ends** so `cmd.Wait` never closes the read-ends. A long-running foreground command is drained in full (no truncation), and a lingering background descendant that inherited a write-end is bounded by a **drain-until-idle** policy rather than the process's runtime — so exec never hangs and never stalls.

## What to review
| file | change |
|---|---|
| worker/cloudflare-container-runner/main.go | self-owned `os.Pipe()` ends; drain-until-idle (progress counter → 300ms idle close, 5s absolute cap) |
| worker/cloudflare-container-runner/stdout_truncation_test.go | full-output, prompt-return-under-descendant, noisy-descendant-bounded, bursty-idle-reset, long-foreground no-truncation |

## How it works
`runCommand` creates `rOut/wOut`, `rErr/wErr`, sets `cmd.Stdout=wOut`/`cmd.Stderr=wErr`, and after `Start()` closes the parent's write-end copies so only the child and its descendants hold write-ends. The copiers read the parent-owned read-ends, which `cmd.Wait` never touches. Sequence: start copiers → `cmd.Wait` (child may run arbitrarily long; every byte is delivered) → a drain loop that returns instantly on real EOF, else closes the read-ends once (`sync.Once`) after **300ms with no copier progress** (buffered output drained, descendant quiet), bounded by a **5s absolute cap** (a continuously-writing descendant). Zero truncation by construction — the idle timer only fires after progress stops, and a copier blocked in a slow response write (downstream backpressure) is tracked as in-flight rather than idle, so the drain never closes mid-flush.

## Motivation
Fixes stdout truncation. The two prior iterations were correctly rejected: `wg.Wait()` before `cmd.Wait()` introduced an unbounded hang when a background descendant held the inherited pipe (repro `sleep 60 & echo done` → context deadline); a fixed 5s grace then made every such command stall 5s. This drain-until-idle design resolves all three: no truncation, no hang, no stall.

## Testing
- `go build ./...` + `go vet ./...` at repo root and in the runner module; `gofmt` clean.
- `go test -race -run TestRunCommand -count=3 .` (runner module) — green ×3.
- Red→green: the prompt-return assertion (return under `pipeDrainGrace/2`) **fails on the fixed-grace version** (5.1s) and **passes now** (0.55–0.64s). The 1 MiB full-delivery and long-foreground no-truncation guarantees are retained.

## Real-behavior proof
Captured live against the built runner (a standalone HTTP server; no coordinator/lease needed) — base vs this branch:
```
BASE (pre-fix)  1 MiB stream: 1048576/1048576 delivered    sleep 60 & echo done: returned 0.46s, output intact
HEAD (branch)   1 MiB stream: 1048576/1048576 delivered    sleep 60 & echo done: returned 0.80s, output intact
```
The truncation itself is a **race masked by socket buffering** — a slow network client cannot force it (the TCP send buffer + HTTP buffering absorb ~1 MiB before backpressure reaches the copiers), so it does not reproduce as a live base-vs-head differential. The deterministic regression reproduces it directly at the real `runCommand`/`os.Pipe`/subprocess level (base loses 49,152 bytes of 1 MiB; branch delivers all), and this branch eliminates the race by construction. The live capture above confirms the branch delivers full output and returns promptly under a background descendant.

### DIFF
--- a/worker/cloudflare-container-runner/main.go
+++ b/worker/cloudflare-container-runner/main.go
@@ -14,6 +14,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"syscall"
 	"time"
 )
@@ -23,6 +24,37 @@ import (
 const minStreamLifetime = 75 * time.Millisecond
 const finalStreamFlushDelay = 100 * time.Millisecond
 const streamHeartbeatInterval = 15 * time.Second
+
+// After the direct child is reaped, a background descendant may have
+// inherited a stdout/stderr write-end (e.g. `sleep 30 & echo done`) and still
+// hold it open, so the owned read-ends never reach a natural EOF on their own.
+// runCommand handles that with an IDLE-AWARE DRAIN, not a single fixed grace:
+//
+//   - pipeDrainIdle is how long the drain waits for a lull with NO new bytes
+//     copied before concluding the child's real output is fully flushed and
+//     only a quiet (or absent) descendant remains. Once idle elapses with no
+//     progress, we close the read-ends ourselves to unblock the copiers. A
+//     descendant that keeps writing (even slowly) keeps resetting this timer,
+//     so genuine trailing output is never cut short.
+//   - pipeDrainPoll is the polling interval used to sample the shared `copied`
+//     byte counter and decide whether progress was made since the last check.
+//   - pipeDrainGrace is an ABSOLUTE CAP on the whole drain phase, in case a
+//     descendant writes continuously (e.g. `(while true; do echo x; done) &`)
+//     and so never goes idle: once this cap elapses (even mid-progress) we
+//     close the read-ends and return anyway. It is the same 5s constant the
+//     bounded-select design used, now acting as a backstop rather than the
+//     primary timer.
+//
+// None of this bounds the foreground command's own runtime: runCommand owns
+// the pipe read-ends, so cmd.Wait never closes them, and a long-running
+// foreground command is fully drained however long it takes — the drain phase
+// (and these three timers) only start once cmd.Wait has already returned. All
+// three are vars so tests can lower them.
+var (
+	pipeDrainIdle  = 300 * time.Millisecond
+	pipeDrainPoll  = 100 * time.Millisecond
+	pipeDrainGrace = 5 * time.Second
+)
 
 type execRequest struct {
 	Command   string            `json:"command"`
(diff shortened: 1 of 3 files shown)


steipete (dismissed):
Blocking correctness regression: moving the pipe wait before command wait can hang indefinitely when a background descendant inherits stdout or stderr.

Production-path repro on exact head: run a command containing sleep 60 in the background under a two-second context. runCommand returned context deadline exceeded after two seconds in three of three race-enabled runs. Current main returns promptly because command wait closes the inherited pipes.

Please use a bounded post-parent-exit drain policy that preserves buffered output without waiting indefinitely for descendant-owned pipes. Keep the full-output regression and add a background-descendant regression.

anagnorisis2peripeteia:
@clawsweeper re-review

anagnorisis2peripeteia:
@clawsweeper re-review

steipete (approved):

steipete:
The pre-fix bursty-stream regression was deterministic:

```text
go test ./worker/cloudflare-container-runner/ -run 'TestRunCommandIdleDrainDeliversBurstyBackgroundOutput' -count=1
--- FAIL: TestRunCommandIdleDrainDeliversBurstyBackgroundOutput
    stdout_truncation_test.go:437: stdout = "done\nBURST_A\n", missing "BURST_B" (idle timer must reset on progress, not truncate a slow-but-live bursty stream)
FAIL
```

Root cause: the post-`cmd.Wait` idle deadline was effectively anchored to the start of the drain. Copier progress did not extend the quiet window, so a live descendant emitting spaced bursts could have later output cut off when the read ends were closed.

The latest PR head resets `idleDeadline` whenever `copied` advances, while retaining the `writesInFlight` guard, prompt natural-EOF completion, and the separate absolute `pipeDrainGrace` cap for a continuously writing detached descendant.

Repeated proof on the latest PR head (`7135c1455ebca3ad37ae897e9e940d5fb667851c`):

```text
for i in 1 2 3 4 5; do
  (cd worker/cloudflare-container-runner && go test . -run TestRunCommandIdleDrainDeliversBurstyBackgroundOutput -count=1)
done

RUN:1  PASS
RUN:2  PASS
RUN:3  PASS
RUN:4  PASS
RUN:5  PASS
```

Additional verification:

```text
go build ./... && go vet ./...                                      PASS
(cd worker/cloudflare-container-runner && go build ./... && go vet ./...) PASS
(cd worker/cloudflare-container-runner && go test . -count=1)       PASS (6.746s)
```

The explicit prompt-return, long-foreground-output, and absolute-cap tests also pass without hanging.

PR #1099, opened 2026-07-15
fix(cli): clean up abandoned egress opens to prevent conn+goroutine leak

anagnorisis2peripeteia:
fix(cli): clean up abandoned mediated-egress opens on timeout or cancel

## Summary
When a mediated-egress `openRemote` times out (the 20s `egressOpenTimeout`) or its context is canceled, the client no longer retains the pending entry or leaves a late host-side connection and copy goroutine behind. The long-lived egress bridge stays usable for subsequent opens.

## What to review
| file | change |
|---|---|
| internal/cli/egress.go | client `openRemote` timeout/cancel path: finish the pending open, remove it from the pending map, and signal the host to close the destination conn + its copy goroutine |
| internal/cli/egress_openremote_leak_test.go | regression driving the real client path through the real timeout |

## How it works
On the timeout (`<-timer.C`) or cancel (`<-ctx.Done()`) branch of `openRemote`, the client finishes the open, drops the id from the pending map, and emits the `close` the host already handles (`hostOpen` registers `conns[id]`; a `close` tears down the destination conn + `copyConnToBridge`). The host-side teardown path is pre-existing and unchanged — this fix makes the client actually send that `close` and clean its own pending state.

## Testing
- `go build ./...` + `go vet` + `go test -race` (internal/cli) green.
- The regression drives `connectEgressBridge` + `clientReadLoop` + `openRemote` over a **real** WebSocket through the **real** 20s `egressOpenTimeout` against a blackholed dial, and fails deterministically on the base revision (pending entry + host conn + copy goroutine leak) / passes on this branch.

## Real-behavior proof
The regression exercises the real client code over a real WebSocket; the host peer is faked because a live host requires a running coordinator + provisioned lease. This contribution was developed without coordinator or provider access, so the deterministic base-vs-branch reproduction over the real client path is the strongest evidence available here — the only uncovered slice (the host reacting to `close`) is pre-existing, unchanged host code. A maintainer with a coordinator + lease can confirm the live mediated-egress cleanup if required.

### DIFF
--- a/internal/cli/egress.go
+++ b/internal/cli/egress.go
@@ -597,7 +597,14 @@ func (b *egressBridge) clientReadLoop(ctx context.Context) error {
 		}
 		switch msg.Type {
 		case "open_ok":
-			b.finishOpen(msg.ID, nil)
+			if !b.finishOpen(msg.ID, nil) {
+				// Late open_ok for an open we already abandoned (timed out or
+				// canceled): the host registered the destination conn and its
+				// copy goroutine, so tell it to close, or they leak for the life
+				// of the bridge. Async — a synchronous host write here would
+				// stall the single read loop (and every stream) on a full socket.
+				go b.closeHostConn(msg.ID)
+			}
 		case "error":
 			b.finishOpen(msg.ID, errors.New(msg.Error))
 			b.closeConn(msg.ID)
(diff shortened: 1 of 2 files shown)


steipete (dismissed):
Deep-reviewed and improved at exact head 9ac7d8b7.

The timeout/cancellation cleanup now removes pending client state, closes host connections created by late successful opens, and also removes pending state when the initial WebSocket write fails. The regression drives the real client bridge and WebSocket path through the production 20-second timeout.

Proof:
- focused race regression green
- structured autoreview clean, confidence 0.94
- changelog credit included

Approved. Independent last-push approval and exact-head CI remain the merge gates.

anagnorisis2peripeteia:
@clawsweeper re-review

Issue #1096, opened 2026-07-15
Tracking: latent concurrency bugs in the CLI + container-runner (goroutine/timer leaks, races, ctx handling)


anagnorisis2peripeteia:
A systematic concurrency audit of `internal/cli` and `worker/cloudflare-container-runner` surfaced a set of latent bugs the `-race` CI does not catch — goroutine/timer leaks, output-truncation races, a lost-update logic race, and mishandled cancellation. Each is independently reproduced with a failing test on `main` (`808d96f`) and will be sent as its own focused PR referencing this issue.

**How these were found:** a systematic concurrency audit (I'm building a data-race / goroutine-leak / TOCTOU hunter, `signalkrebs`). The tool surfaces *candidates*; every finding below is then independently reproduced with a failing test on `main` and gated before filing — the proof is manual, not generated.

None cross a security boundary (the destructive-cleanup class is being reported separately); these are reliability/correctness. Happy to split this into per-file issues if the maintainers prefer.

### Findings (each → its own PR)
- [x] **`worker/cloudflare-container-runner/main.go:204`** — *stdout truncation.* `runCommand` calls `cmd.Wait()` before `wg.Wait()`; `cmd.Wait` closes the pipes on exit, racing the `copyPipe` goroutines and silently dropping buffered output (`os.ErrClosed` masked as EOF). Repro loses ~16–64 KiB deterministically. → **#1097** (open)
- [x] **`internal/cli/run_recorder.go:195`** — *goroutine + ticker leak.* `runRecorder.Failed` marks the run finished but, unlike `Finish`, never calls `stopTelemetrySampler`, so the 15s telemetry sampler goroutine leaks on every failure path. → **#1098** (open)
- [x] **`internal/cli/egress.go:669`** — *goroutine + conn leak.* `openRemote` abandons a timed-out/canceled open without deleting `pending[id]` or closing the host conn, leaking the destination TCP conn + its `copyConnToBridge` goroutine for the life of the bridge. → **#1099** (open)
- [x] **`internal/cli/code.go:418`** — *goroutine + websocket leak.* `codeBridge` has no `closed` flag, so an in-flight `openUpstreamWebSocket` can register a new upstream + spawn `readUpstreamWebSocket` after `Close` ran, leaking the goroutine and local code-server websocket. → **#1100** (open)
[...]

PR #1104, opened 2026-07-15
fix(cli): deliver server-first egress banner racing the CONNECT handshake

anagnorisis2peripeteia:
## Summary

Fix the server-first mediated-egress race tracked in https://github.com/openclaw/crabbox/issues/1096 without blocking the shared WebSocket reader.

The local proxy stream is registered before the remote open can emit data. Inbound frames then use a bounded, ordered queue per stream until the local CONNECT handshake is complete. A slow or abandoned stream therefore cannot stall unrelated multiplexed connections.

## Maintainer improvements

- Replaced the contributor's shared `sync.Cond` wait with asynchronous per-stream delivery.
- Bounded each client queue to 256 KiB; overflow closes only that stream and its peer.
- Preserved frame order, including draining final data before a terminal close.
- Preserved the rejected-open `502 Bad Gateway` response.
- Closed the established host connection when the local proxy handshake fails.
- Added changelog credit and preserved Cameron Beeley's authorship and co-author trailer.

## Regression coverage

- CONNECT response strictly precedes an intact server-first banner.
- A blocked stream cannot stop the real shared WebSocket read loop from delivering another stream.
- Multiple frames preserve order.
- Terminal close drains earlier frames.
- Queue overflow is isolated to the offending stream.
- Rejected opens return `502` through the real read loop.
- Handshake write failure notifies and closes the host peer.

Focused race tests passed 50 consecutive runs during development and 20 consecutive runs on the rebased exact head. Full `go test -race ./...`, `go vet ./...`, native build, Windows cross-compile, Worker format/lint/typecheck/Vitest/build, and docs build passed locally.

Structured autoreview accepted two early findings (terminal-frame ordering and host cleanup on handshake failure). Both were fixed with regressions. Final exact-head branch review: clean, no actionable findings, confidence 0.94.

## Live behavior proof

Validated the built branch binary through the production coordinator path with a disposable AWS lease and `github.com` as the sole allowlist entry:

```text
egress: host=true client=true allow=github.com
allowed_status=HTTP/1.1 200 Connection Established
allowed_banner=SSH-2.0-3477046
concurrent_status=HTTP/1.1 200 Connection Established
concurrent_banner=SSH-2.0-3477046
rejected_status=HTTP/1.1 502 Bad Gateway
```

For the concurrency probe, the first allowed CONNECT socket remained unread while the second completed through the same bridge. Cleanup was also verified: local host daemon stopped, remote client stopped, coordinator reported `host=false client=false`, the remote process probe found no client, and the disposable lease was released.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Fixed
 
+- Delivered server-first mediated-egress bytes after the local proxy handshake without letting one slow stream block unrelated connections. Thanks @anagnorisis2peripeteia.
 - Serialized complete per-lease egress host-daemon starts and stops and atomically replaced the remote client, preventing concurrent lifecycle commands and ordinary restarts from leaving untracked or stale processes. Thanks @anagnorisis2peripeteia.
 - Closed code-server WebSockets whose upstream dial completes after bridge shutdown, preventing orphaned connections and reader goroutines. Thanks @anagnorisis2peripeteia.
 - Stopped failed runs' telemetry samplers promptly so long-lived CLI processes do not retain ticker goroutines or continue probing released leases. Thanks @anagnorisis2peripeteia.
--- a/internal/cli/egress.go
+++ b/internal/cli/egress.go
@@ -32,6 +32,7 @@ const (
 	egressRemoteLog         = "/tmp/crabbox-egress-client.log"
 	egressMaxMessageBytes   = 2 * 1024 * 1024
 	egressCopyChunkBytes    = 32 * 1024
+	egressClientQueueBytes  = 8 * egressCopyChunkBytes
 	egressOpenTimeout       = 20 * time.Second
 	egressDialTimeout       = 15 * time.Second
 	egressRemoteReadyWait   = 5 * time.Second
@@ -50,6 +51,15 @@ type egressProxyMessage struct {
 
 type egressOpenResult struct {
 	err error
+}
+
+type egressClientConn struct {
+	conn            net.Conn
+	ready           bool
+	queue           [][]byte
+	queuedBytes     int
+	draining        bool
+	closeAfterDrain bool
 }
 
 func (a App) egress(ctx context.Context, args []string) error {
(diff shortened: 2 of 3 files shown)


steipete (dismissed):
Deep-reviewed and improved at exact head 26950fef.

Server-first mediated-egress data now waits behind the local proxy handshake instead of being dropped. The repaired error path unregisters the early socket before waking the proxy handler, preserving HTTP 502 responses for rejected opens. Bridge shutdown also wakes blocked readers.

Proof:
- five focused protocol and lifecycle regressions green for 20 consecutive runs
- structured autoreview clean, confidence 0.92
- changelog credit included

Approved. Independent last-push approval and exact-head CI remain the merge gates.

anagnorisis2peripeteia:
@clawsweeper re-review

steipete:
Landed after maintainer redesign and live behavior proof.

- Exact reviewed head: `8f92d0dfcb1a7b3577281eb63b5ef822e1093f48`
- Merge commit: `435ea00ac2e13e5f7cedd9a3a6f22a9c3d7b6b48`
- Hosted exact-head CI: 12/12 green across https://github.com/openclaw/crabbox/actions/runs/29527524012 and https://github.com/openclaw/crabbox/actions/runs/29527524005
- Local proof: focused race coverage stressed 50x during repair and 20x on the rebased head; full `go test -race ./...`, vet, native build, Windows cross-compile, Worker gate, and docs build passed
- Live proof: through a disposable coordinator-backed AWS lease, an allowed CONNECT returned `200 Connection Established` before an intact `SSH-2.0-3477046` banner; a second connection succeeded while the first remained unread; a disallowed target returned `502 Bad Gateway`
- Cleanup proof: local host daemon stopped, remote client stopped, coordinator reported both bridge sides disconnected, remote process probe found no client, and the lease was released
- Final structured autoreview: clean, no actionable findings, confidence 0.94

The merged implementation uses bounded ordered delivery per stream, so one slow connection cannot block the shared reader. It also drains final data before close and cleans the host peer after local handshake failure. Contributor authorship and changelog credit are preserved. Thanks @anagnorisis2peripeteia.

steipete:
Maintainer review of the merged-with-`main` result (Codex autoreview, `internal/cli/egress.go`). The `data`-frame race fix is correct, but the same early registration exposes a second, symmetric race on the **close** path — flagging before merge:

Registering the socket before `openRemote` means the existing `case "close"` path can now find and close the connection while `notReady` is still true, but only `data` frames honor the readiness barrier. A legitimate upstream that accepts and immediately closes produces `open_ok` followed by `close`; the read loop can process both before `handleProxyConn` is scheduled, so the CONNECT client receives EOF instead of `200 Connection Established`.

Fix direction: make the inbound `close` path wait for (or record) readiness the same way inbound `data` does, with a regression test covering the `open_ok`-then-immediate-`close` ordering. Once that lands I'll re-run the review and land it. The rest of the change (including the merge against the now-landed #1099/#1103 egress work) is clean and race-verified.

PR #1109, opened 2026-07-16
feat: add Zed Remote Projects handoff

zozo123:
Related: https://github.com/zed-industries/zed/pull/61102

## What Problem This Solves

Resolves a problem where users opening a Crabbox workspace in Zed Remote Projects had to assemble the connection workflow themselves: resolve the lease, synchronize the checkout, paste the SSH command, discover the remote folder, and keep the lease active while Zed runs its own SSH processes.

## Why This Change Was Made

Adds a provider-neutral `crabbox zed` command that reuses the existing SSH target resolution and lease-activity paths. It verifies the current repository's synced remote folder, prints a Zed-compatible SSH command plus the exact folder, and holds the lease activity until Ctrl-C.

The command does not modify `~/.ssh/config` or Zed settings. It rejects Windows remote targets and token-as-username credentials because Zed remote servers support Linux/macOS and editor settings should not persist provider secrets. The emitted executable is unquoted for compatibility with current Zed releases while every SSH argument remains shell-quoted.

## User Impact

Users can now run:

```sh
crabbox run --id <lease> --sync-only
crabbox zed --id <lease>
```

Then paste the printed command into Zed's **Connect New Server** flow, open the printed folder, and keep the command running for lease activity. The configured hard TTL still applies, and stopping the command does not release the lease.

## Evidence

- `go test -race -p 1 ./...`
- `go vet ./...`
- `./scripts/check-docs.sh`
- Live local-container acceptance: provisioned a Linux lease, synced 1,510 files (22.9 MiB), emitted a command containing a spaced identity path and the exact remote folder, held the session, exited cleanly on Ctrl-C, and released the lease/container.
- Security/config implications: no credential output for token-auth providers and no persistent SSH or editor configuration changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,7 @@ crabbox prewarm                                      # lease + Actions hydration
 crabbox run --id blue-lobster -- pnpm test:changed
 crabbox connect blue-lobster                         # open an interactive SSH session
 crabbox ssh --id blue-lobster
+crabbox zed --id blue-lobster                       # prepare a Zed Remote Projects session
 crabbox stop blue-lobster
 ```
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -99,6 +99,7 @@ See [history](commands/history.md), [logs](commands/logs.md),
 ```text
 crabbox connect <id>                          open an interactive SSH session
 crabbox ssh --id <id>                          print the SSH command
+crabbox zed --id <id>                          prepare a lease for Zed Remote Projects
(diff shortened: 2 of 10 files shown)


steipete:
Superseded by #1127, which lands the same Zed handoff as a general 'crabbox open --editor' surface (zed + VS Code/Cursor) with maintainer review. Thanks @zozo123 — your approach is credited in the commit.

PR #1118, opened 2026-07-16
test(cli): cover immediate egress close handshake

steipete:
Follow-up to https://github.com/openclaw/crabbox/pull/1104.

## Summary

- document the invariant that an inbound close arriving immediately after open_ok must remain queued until the CONNECT response is written
- add a deterministic WebSocket regression that sends open_ok followed immediately by close before the proxy handshake can complete
- assert the CONNECT client receives 200 Connection Established before EOF

## Regression proof

- controlled before replay, with the readiness term removed: new test fails with `io: read/write on closed pipe`
- restored readiness gate: new test passes under `-race`, including `-count=5`

## Verification

- `go build ./...`
- `go vet ./...`
- `go test ./internal/cli/ -race -run 'Egress|Banner|AbandonedOpen|Openremote|Daemon' -count=1`
- `go test ./internal/cli/ -race -run '^TestEgressImmediateUpstreamCloseFollowsProxyHandshake$' -count=5`
- `.agents/skills/autoreview/scripts/autoreview --mode uncommitted` — clean, no accepted/actionable findings

No merge requested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Fixed
 
+- Delivered server-first mediated-egress bytes after the local proxy handshake without letting one slow stream block unrelated connections. Thanks @anagnorisis2peripeteia.
 - Serialized complete per-lease egress host-daemon starts and stops and atomically replaced the remote client, preventing concurrent lifecycle commands and ordinary restarts from leaving untracked or stale processes. Thanks @anagnorisis2peripeteia.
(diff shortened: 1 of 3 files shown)


steipete:
Closing this. The close-frame race I opened it for was in a superseded head of #1104 — the version that actually merged (`435ea00a`) was rewritten to a per-stream ordered-bounded-queue design that already buffers server-first bytes until the local proxy handshake completes and preserves ordering through terminal close frames. An independent Codex review of the merged commit confirms it's correct, so there's no bug to fix here. The author's own `egress_banner_buffer_test.go` already covers the behavior; a second test isn't worth the review burden.

Issue #1123, opened 2026-07-17
Tracking: cross-process lease/testbox TOCTOU in tart / firecracker / blacksmith cleanup paths


anagnorisis2peripeteia:
A systematic concurrency audit of the provider cleanup paths surfaced three cross-process **TOCTOU**
bugs the `-race` CI does not catch. Each makes a **destructive decision from a stale pre-operation
snapshot without re-checking ownership before acting** — delete a lease claim, kill a VM, or stop a
testbox. When a second crabbox process (for example a concurrent CI job for the same repo/config)
creates or reclaims a resource inside that check→act window, the first process's cleanup destroys the
second's healthy, in-use resource. The trigger is **two concurrent crabbox runs** — no attacker required.

**How these were found:** the same systematic concurrency audit that produced #1096 (I'm building a
data-race / goroutine-leak / TOCTOU hunter, `signalkrebs`). The tool surfaces *candidates*; every
finding below is then independently reproduced with a failing test on `main` (`808d96f`) and gated
before filing — the proof is manual, not generated.

Unlike #1096 (reliability/correctness), these **do** cross the SECURITY.md boundary — "destructive
provider actions against resources Crabbox cannot strongly identify as its own." This is the
`[security]` destructive-cleanup class the project already hardened for other providers in
https://github.com/openclaw/crabbox/pull/446 ("require orphan sweep ownership proof", AWS/Azure) and
https://github.com/openclaw/crabbox/pull/781 ("require claims for provider mutations", Freestyle/Islo);
it reaches three providers those sweeps did not cover. Per SECURITY.md ("send a pull request with the
fix"), each is sent as its own fix PR so the fix and the report arrive together.

Each fix reuses the project's existing ownership-proof primitives and ships a regression test that
FAILS on base `808d96f` and PASSES with the fix (a paired base/HEAD comparison — stronger than a live
run for a timing race, which almost never hits the exact interleaving).

### Fixes (each → its own PR)
- [ ] **tart** — `(*backend).Cleanup` orphan sweep deletes a concurrent `Acquire`'s claim: it snapshots
  the VM list *before* reading claims, so a claim registered during a slow `stopVM`/`deleteVM` reads as
  a "missing-instance" orphan and its running VM's lease + SSH key are deleted. Fix: snapshot orphan
  candidates before listing instances + guard the delete with `core.RemoveLeaseClaimIfUnchanged`. →
  https://github.com/openclaw/crabbox/pull/1124
[...]

PR #1125, opened 2026-07-17
fix(firecracker): guard Cleanup claim removal against a concurrent reclaim

anagnorisis2peripeteia:
`(*backend).Cleanup` decides via `shouldCleanupRecord`, releases the VM, then calls the **unguarded** `core.RemoveLeaseClaim`. A concurrent `Resolve`/`Touch` that reclaims the same lease inside that check→destroy window has its freshly-rewritten claim deleted out from under the live session. Trigger is two concurrent runs; no attacker. Same `[security]` class as https://github.com/openclaw/crabbox/pull/446.

**Fix:** guard both removals with `core.RemoveLeaseClaimIfUnchanged` against the pre-cleanup snapshot (mirrors `namespaceinstance` backend.go:455,473).

**Verify:** `go test ./internal/providers/firecracker/ -run TestCleanupTOCTOU -race` — `TestCleanupTOCTOUDeletesConcurrentlyReclaimedClaim` fails on base, passes with this fix. `go vet ./... && go build ./cmd/crabbox`.

Part of https://github.com/openclaw/crabbox/issues/1123.

### DIFF
--- a/internal/providers/firecracker/backend.go
+++ b/internal/providers/firecracker/backend.go
@@ -436,7 +436,19 @@ func (b *backend) Cleanup(ctx context.Context, req CleanupRequest) error {
 		if err := b.releaseStateRecord(ctx, cfg, record, true); err != nil {
 			return err
 		}
-		core.RemoveLeaseClaim(record.LeaseID)
+		// Guard the claim removal against a concurrent Resolve/Touch that
+		// reclaimed this lease after our cleanup snapshot: RemoveLeaseClaimIfUnchanged
+		// declines to delete a claim rewritten in the check->destroy window,
+		// leaving the reclaiming process's live lease intact. (namespaceinstance
+		// backend.go:455,473 uses the same guard.)
+		if claim := claims[record.LeaseID]; claim.LeaseID != "" {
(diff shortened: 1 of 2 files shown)


steipete:
Closing after maintainer lifecycle review.

The reported interleaving does not preserve a live Firecracker session. The injected reclaim runs from `cleanupNetwork`, after `releaseStateRecord` has already completed `stopRecordedProcess`. Cleanup then removes the lease state directory. A claim rewritten at that point therefore points to a stopped VM with no local state; the provider's own `Resolve` path rejects it as a stale local claim.

I verified that behavior directly by extending the submitted regression with a post-cleanup `Resolve` assertion and running it 20 times under the race detector. The submitted patch preserved the claim every time, and every preserved claim produced the expected `stale local claim but no local state` failure. In other words, the test's “live lease” premise is false.

The proposed compare-and-remove also happens after destructive teardown. If metadata changes during teardown, it leaves a claim for a VM that was already stopped and deleted. `Touch` only updates metadata; it cannot revive the process. A valid `Resolve` checks the exact local state and process before publishing a claim, so it cannot reclaim the stopped resource through this window.

No code landed. Thanks @anagnorisis2peripeteia for the report and deterministic interleaving harness; it made the lifecycle distinction easy to prove.

PR #678, opened 2026-06-25
feat(cua): add experimental read-only diagnostics provider

coygeek:
## Summary

Adds an experimental CUA provider for read-only diagnostics and existing-sandbox inventory.

The provider is intentionally registered as service control, not delegated execution. It supports:

- `crabbox doctor --provider cua`
- `crabbox list --provider cua`
- `crabbox status --provider cua --id <existing-sandbox-id>`
- trusted local configuration, environment-only credentials, provider metadata, documentation, and a non-mutating smoke test

All remote lifecycle mutation fails closed before the CUA SDK can perform it. Create, warmup, run, stop, cleanup, and bridge deletion are disabled.

## Safety gate

The pinned upstream create API has no idempotency key or client-assigned identity echoed by create, list, and get. A create that succeeds server-side but times out client-side can therefore leave a billed sandbox that Crabbox cannot identify safely.

Deletion is also disabled because the current API cannot atomically bind deletion to an immutable sandbox identity; verify-then-delete by mutable name has a replacement race.

Tracking: https://github.com/openclaw/crabbox/issues/381

## Verification

- `go build ./...`
- `go vet ./...`
- `go test ./internal/providers/cua/... ./internal/cli/ -run 'CUA|Provider|Registry' -count=1`
- `go test ./internal/providers/cua/... ./internal/providers/all -count=1`
- `node --test scripts/live-cua-smoke.test.js`
- `scripts/live-cua-smoke.sh`
- `node scripts/check-provider-matrix.mjs`
- `.agents/skills/autoreview/scripts/autoreview --mode branch --base origin/main` — clean, no accepted/actionable findings

Merge remains pending a maintainer nod.

Issue #381, opened 2026-06-15
Add CUA as a delegated-run provider


coygeek:
## Summary

Add a first-class `cua` provider so Crabbox can run commands in CUA cloud sandboxes.

The first implementation should treat CUA as a delegated-run provider, not as a normal SSH lease provider. CUA's public surface is centered on sandbox lifecycle, a computer-server transport, shell execution, file operations, screenshots, VNC/display URLs, and multi-OS computer-use APIs. It does not hand Crabbox a stable Crabbox-managed SSH target with per-lease keys, rsync, and Crabbox-owned command transport. A delegated provider lets Crabbox use CUA's real sandbox contract while keeping Crabbox core responsible for provider selection, local claims, sync manifests, timing, normalized list/status/stop rendering, and option validation.

## Why this fits Crabbox

Crabbox already separates provider integrations into:

- SSH lease providers, where Crabbox owns SSH readiness, rsync, command execution, heartbeats, VNC/code surfaces, and release.
- Delegated-run providers, where the provider owns execution and Crabbox owns config, claims, timing, local repo sync policy, normalized list/status/stop, and user-facing workflow consistency.

CUA maps much more naturally to the delegated-run side:

- CUA can create or connect to a named sandbox through the Python Sandbox SDK.
- CUA can run shell commands through `Sandbox.shell.run(...)` or the cloud computer-server `/cmd` path.
- CUA can read and write files inside the sandbox through the Sandbox SDK files interface.
- CUA can list, inspect, suspend/resume/restart, and delete sandboxes.
- CUA can expose display/VNC URLs, screenshots, and computer-use controls, but those are not the same as Crabbox's existing SSH/VNC data plane and should not be advertised as Crabbox desktop/browser/code support in v1.

The initial provider should support the highest-confidence Crabbox workflow first: cloud Linux sandbox command execution with archive sync and cleanup. CUA's macOS, Windows, Android, local runtime, screenshots, display URLs, snapshots, and agent-control features are valuable follow-ups, but they should be added only when Crabbox can expose them honestly without pretending they are SSH leases.

## Evidence from current CUA and Crabbox source

Current CUA source and generated docs expose the primitives Crabbox needs:

- CUA is described as "agent-ready sandboxes for any OS" with `Sandbox.ephemeral(Image.linux())`, `Image.macos()`, `Image.windows()`, and `Image.android()` examples.
- Cloud CUA requires `CUA_API_KEY`; local Linux Docker can run without a cloud account.
[...]

PR #1134, opened 2026-07-17
feat(providers): add local Lume macOS leases

madhavajay:
## Summary

- add a native `lume` SSH-lease provider for local Apple-silicon macOS VMs
- clone one stopped golden image into one isolated VM per Crabbox lease
- rotate SSH host keys on first clone boot, pin a per-lease host identity, and inject only a unique public key
- serialize acquisition and enforce Apple's two-running-macOS-guest host limit
- identity-fence the detached `lume run` owner, work around Lume 0.3.16 stop/readiness defects, and verify stop/delete before claim cleanup
- add claim-safe release and multi-storage cleanup, trusted configuration, ARM64 target validation, docs, metadata, and live-smoke dispatch
- add reusable macOS image hooks for first-boot identity and optional Cua Driver startup, plus noninteractive developer-tool setup improvements

## Why

Crabbox needs a local macOS backend that can reuse layered Lume images without baking credentials into them. Each session should have its own VM, filesystem, machine identity, SSH key, and host-key trust while still supporting normal Crabbox sync/run/release behavior.

This is separate from #678, which adds a cloud Linux CUA provider; this PR covers local macOS VMs on user-owned Apple-silicon infrastructure.

## Validation

- `go test -race ./internal/providers/lume ./internal/providers/all -count=1`
- `go vet ./internal/providers/lume ./internal/providers/all`
- full `go test ./internal/cli -count=1`
- Windows ARM64 cross-compilation of the Lume provider
- docs, provider-matrix, shell syntax, plist, and Node live-smoke checks
- real Apple-silicon lifecycle with two concurrent leases from the same stopped golden image:
  - distinct VM names, IPs, platform UUIDs, SSH keypairs, and ED25519 host-key fingerprints
  - independent Crabbox commands at the same guest path
  - Cua Accessibility, Screen Recording, and actual desktop capture passed in both VMs
  - Tailscale remained installed but logged out with `NeedsLogin` in both VMs
  - concurrent release left no VM, owner process, claim, key, known-hosts, private log, or temporary password-helper residue

## Current boundaries

- macOS/ARM64 only; Lume is not used for Windows or Linux guests
- the prepared guest currently needs Lume's default bootstrap account/password; custom bootstrap passwords are not yet supported
- desktop/webVNC bridging and Crabbox-managed Tailscale enrollment are intentionally not advertised yet
- reusable image layers remain credential-free; credentials are injected only into disposable clones after boot


madhavajay:
Sorry, my clanker is a little too excited here, let me do some more testing.
I saw this: https://github.com/openclaw/crabbox/issues/381 and this https://github.com/openclaw/crabbox/pull/678

But it appears they are focused on cua with linux first.
I started using crabbox to give xcode over ssh to my linux agents which worked okay but im running into limits and want better isolation so when I saw cua mentioned on x, I figured it was worth trying.

I'll report back shortly once I have a firm idea on how this goes.

madhavajay:
@steipete omg legend, thank you. ❤️ I was busy adding fast native display to lume over here: https://github.com/trycua/cua/pull/2401 and came back to find this done.

If you use mac crabboxes I highly recommend the native display.

PR #1140, opened 2026-07-17
docs(agents): note release gotchas

steipete:
Adds a terse Releasing section to AGENTS.md capturing the pitfalls that broke the 0.39.0 cut: the tag annotation must be the bare version, every version-carrying file must be bumped (changelog + worker/package.json + both worker/package-lock.json root entries), the producer needs a merged authorize-source record, and the producer must run credential-free. Points at docs/RELEASING.md as the authority.


steipete:
Good point — updated to point at the full credential check at the top of scripts/build-release-candidate.sh (GitHub, Homebrew-tap, and Actions tokens + codesign identity + notary profile) rather than an abbreviated list, so it can't go stale.

PR #1146, opened 2026-07-17
fix(apple-container): guard Cleanup orphan sweep against a concurrent Acquire's claim

anagnorisis2peripeteia:
## Summary
Guards the `apple-container` provider's `Cleanup` against a cross-process TOCTOU race that deletes a
concurrent crabbox run's **live** lease claim. `(*backend).Cleanup` snapshots containers via
`listContainers()` **before** it reads lease claims via `core.ListLeaseClaims()`, then its orphan sweep
removes — via the **unguarded** `core.RemoveLeaseClaim` — any claim absent from that stale container
view. A claim registered by a concurrent `Acquire` (a second `crabbox` run on the same Mac) for a live
container is visible to the newer claim read but absent from the older container view, so it is swept as
"missing container" and that live lease's claim is destroyed. The trigger is two concurrent `crabbox`
runs; no attacker.

Part of https://github.com/openclaw/crabbox/issues/1144. Same `[security]` destructive-cleanup class already hardened for other
providers in https://github.com/openclaw/crabbox/pull/1124 (tart),
https://github.com/openclaw/crabbox/pull/446 (aws/azure), and
https://github.com/openclaw/crabbox/pull/781 (freestyle/islo).

## What to review
| file | change |
|---|---|
| `internal/providers/applecontainer/backend.go` | +23 / −7 — snapshot claims before listing containers; guard the orphan-sweep removal with `RemoveLeaseClaimIfUnchanged` |
| `internal/providers/applecontainer/cleanup_toctou_test.go` | +218 — two regression tests (fresh-claim + reclaim), base-fails / HEAD-passes |

## How it works
Mirrors the merged tart fix (https://github.com/openclaw/crabbox/pull/1124):
- Snapshot orphan-candidate claims **before** `listContainers()`, so a claim newer than the container
  view cannot be misclassified as a "missing container" orphan.
- Guard the sweep removal with `core.RemoveLeaseClaimIfUnchanged`, so a lease reclaimed mid-cleanup is
  spared.
- Leave the stored testbox key in place — `Acquire` creates/reuses the key before publishing its claim,
  so a missing-container sweep cannot safely delete it.

The delete-loop path (a container this `Cleanup` itself removes, from a fresh claim read) is unchanged.

## Evidence (deterministic before/after)
apple-container `Cleanup` requires macOS on Apple silicon and provisions real containers, and this is a
**timing race** a happy-path lifecycle run cannot reliably trigger — so live host_smoke is deferred to
[...]

PR #1124, opened 2026-07-17
fix(tart): guard Cleanup orphan sweep against a concurrent Acquire's claim

anagnorisis2peripeteia:
## Summary

- snapshot Tart orphan-claim candidates before listing instances so claims created after the instance view cannot be misclassified as missing
- remove a candidate only when its persisted claim still exactly matches the snapshot
- retain stored SSH keys during missing-instance claim cleanup because Tart creates or reuses the key before it publishes the replacement claim; safe key garbage collection needs a wider acquisition fence
- cover both post-snapshot claim creation and concurrent rebinding with compact deterministic regressions

Part of https://github.com/openclaw/crabbox/issues/1123.

## Maintainer improvements

- reduced the original 213-line regression file to focused shared fixtures
- strengthened assertions to verify rebound claim contents, successful orphan-claim deletion, and intentional key retention
- isolated both claim and key paths from the developer's real config directories
- added changelog credit for @anagnorisis2peripeteia

## Verification

- base proof: `TestCleanupPreservesClaimCreatedAfterInstanceSnapshot` fails on current `main` because cleanup removes the live claim
- `go test -race ./internal/providers/tart/ -run 'TestCleanup(Preserves|RemovesOrphanedClaimsWithoutDeletingStoredKey)' -count=50`
- `go test -race ./internal/providers/tart/ -count=1`
- `go vet ./...`
- `go build -trimpath -o bin/crabbox ./cmd/crabbox`
- final whole-branch AutoReview clean with no accepted/actionable findings (confidence 0.91)

The repository-wide race command also exposed pre-existing cross-package claim-store leakage when packages share one temporary home; the exact Tart package is isolated and green, and hosted exact-head CI remains the merge authority.

Issue #1144, opened 2026-07-17
Tracking: cross-process orphan-sweep TOCTOU in provider Cleanup paths (apple-container / local-container / external / apple-vm)


anagnorisis2peripeteia:
A systematic audit of the provider **`Cleanup` orphan-sweep paths** surfaced three cross-process
**TOCTOU** bugs the `-race` CI does not catch. Each snapshots live resources (a container/VM/lease
list) **before** reading lease claims, then its orphan sweep removes — via the **unguarded**
`core.RemoveLeaseClaim` — any provider claim absent from that stale snapshot. When a second crabbox
process (e.g. a concurrent CI job for the same repo/config) registers a claim for a genuinely live
resource inside that snapshot→sweep window, the first process's cleanup deletes the second's healthy
lease claim. The trigger is **two concurrent crabbox runs** — no attacker required.

**How these were found:** the same systematic concurrency audit that produced
https://github.com/openclaw/crabbox/issues/1123 (I'm building a data-race / goroutine-leak / TOCTOU
hunter, `signalkrebs`). After the tart/firecracker/blacksmith round, I scanned **every** provider
`Cleanup` for the same orphan-sweep-by-listing shape (`ListLeaseClaims` compared against a stale
resource snapshot, then an unguarded claim removal). The tool surfaces *candidates*; every finding
below is then independently reproduced with a failing test on `main` (`f46ce01f`) and gated before
filing — the proof is manual, not generated.

This is the same `[security]` destructive-cleanup class the project already hardened in
https://github.com/openclaw/crabbox/pull/446 ("require orphan sweep ownership proof", AWS/Azure),
https://github.com/openclaw/crabbox/pull/781 ("require claims for provider mutations", Freestyle/Islo),
and https://github.com/openclaw/crabbox/pull/1124 (tart). It reaches three providers those did not
cover. Per SECURITY.md ("send a pull request with the fix"), each is sent as its own fix PR so the fix
and the report arrive together.

Each fix mirrors the merged tart fix (https://github.com/openclaw/crabbox/pull/1124): snapshot
orphan-candidate claims **before** listing resources, and guard the removal with
`core.RemoveLeaseClaimIfUnchanged` so a lease reclaimed mid-cleanup is spared. Each ships regression
tests that FAIL on base `f46ce01f` and PASS with the fix.

### Fixes (each → its own PR)
- [x] **apple-container** — `(*backend).Cleanup` orphan sweep deletes a concurrent Acquire's claim; the
[...]

PR #1147, opened 2026-07-17
fix(local-container): guard Cleanup orphan sweep against a concurrent same-scope Acquire's claim

anagnorisis2peripeteia:
## Summary
Guards the `local-container` provider's `Cleanup` against a cross-process TOCTOU race that deletes a
concurrent crabbox run's **live** lease claim. `(*backend).Cleanup` snapshots containers via
`listContainers()` **before** it reads lease claims via `core.ListLeaseClaims()`, then its orphan sweep
removes — via the **unguarded** `core.RemoveLeaseClaim` — any provider claim on the current
runtime/context/host scope that is absent from that stale container view. A claim registered by a
concurrent `Acquire` on the **same scope** (a second `crabbox` run against the same docker context on the
same host) for a live container is visible to the newer claim read but absent from the older container
view, so it is swept as "missing container" and that live lease's claim is destroyed. The trigger is two
concurrent same-scope `crabbox` runs; no attacker.

Part of https://github.com/openclaw/crabbox/issues/1144. Same `[security]` destructive-cleanup class
already hardened in https://github.com/openclaw/crabbox/pull/1124 (tart),
https://github.com/openclaw/crabbox/pull/446 (aws/azure), and
https://github.com/openclaw/crabbox/pull/781 (freestyle/islo); sibling to the apple-container fix
https://github.com/openclaw/crabbox/pull/1146.

## What to review
| file | change |
|---|---|
| `internal/providers/localcontainer/backend.go` | +29 / −7 — snapshot claims before listing containers; guard the orphan-sweep **claim** removal with `RemoveLeaseClaimIfUnchanged`, and **retain** the stored testbox key (a concurrent Acquire prepares it before publishing its claim) |
| `internal/providers/localcontainer/cleanup_toctou_test.go` | +326 — three regression tests: fresh-claim reorder, reclaim guard-decline, and key-retention against a concurrent Acquire; scope-forced, base-fails / HEAD-passes |
| `internal/providers/localcontainer/backend_test.go` | +12 / −4 — two existing orphan-claim tests updated to assert the stored key is **retained** (not deleted) by the sweep |

## How it works
Mirrors the merged tart fix (https://github.com/openclaw/crabbox/pull/1124):
- Snapshot orphan-candidate claims **before** `listContainers()`, so a claim newer than the container
  view cannot be misclassified as a "missing container" orphan.
- Guard the sweep removal with `core.RemoveLeaseClaimIfUnchanged`, so a lease reclaimed mid-cleanup is
  spared.
- A genuine orphan has its **claim** removed but its **stored testbox key retained**. A concurrent
  Acquire/reclaim of the same lease prepares or reuses the key (via `EnsureTestboxKeyForConfig`) *before*
  it publishes the replacement claim, so the old claim can still match our snapshot at the instant the key
  already belongs to a live reacquisition — deleting it in the sweep would leave that container unreachable
  over SSH (an availability bug). This mirrors the merged tart fix
  (https://github.com/openclaw/crabbox/pull/1124), which retains per-lease state for this reason; the
  apple-container sibling (https://github.com/openclaw/crabbox/pull/1146) applies the same key retention. A
  genuinely dead lease's key is a harmless small residue in the state dir. The container-removal path
  (which removes a container that is actually present) still deletes the key, unchanged.

**Scope note:** local-container's sweep is scope-filtered (`localContainerClaimMatchesScope`), so the race
is gated to two runs sharing the same docker runtime/context/host — the normal same-machine case. The
regression tests force a non-empty scope (via a faked `docker context show`) so the scope-matched path is
actually exercised.

This is a **timing race** a happy-path lifecycle run cannot reliably trigger, so live host_smoke is
deferred to maintainer CI (as with https://github.com/openclaw/crabbox/pull/1124). Instead the race is
reproduced **deterministically** with three regression tests (the snapshot reorder, the guard-decline, and
the key-retention availability case) that FAIL on base `cf5081fc` and PASS with this change:

```
# BASE (this change stashed) — the sweep deletes the concurrent same-scope claim, and deletes a key
# a concurrent Acquire prepared:
$ go test -race -run 'TestCleanupOrphanSweep|TestCleanupRetainsKey' ./internal/providers/localcontainer/
--- FAIL: TestCleanupOrphanSweepDeletesConcurrentAcquireClaim
    Cleanup's orphan sweep deleted the claim registered by a concurrent same-scope Acquire ... present=false
    remove claim lease=cbx_lcfresh12345 slug=fresh-slug reason=missing container
--- FAIL: TestCleanupOrphanSweepGuardDeclinesReclaimedCandidate
--- FAIL: TestCleanupRetainsKeyPreparedByConcurrentAcquire
    cleanup deleted a stored key prepared by a concurrent Acquire (availability regression)
FAIL

# HEAD (this change) — all three pass; full package suite green:
$ go test -race ./internal/providers/localcontainer/
ok  github.com/openclaw/crabbox/internal/providers/localcontainer
```

`TestCleanupOrphanSweepDeletesConcurrentAcquireClaim` exercises the reorder (a fresh concurrent claim);
`TestCleanupOrphanSweepGuardDeclinesReclaimedCandidate` exercises the `RemoveLeaseClaimIfUnchanged` guard
(a reclaim mid-cleanup). `go build ./...`, `go vet ./...`, `gofmt -l`, and the full `go test -race ./...`
suite are clean.

## Not included (scope)
apple-container is fixed in https://github.com/openclaw/crabbox/pull/1146; the same class in the `external`
provider is a sibling PR to follow under https://github.com/openclaw/crabbox/issues/1144. Providers that
already guard the sweep (tart https://github.com/openclaw/crabbox/pull/1124, aws/azure
https://github.com/openclaw/crabbox/pull/446, freestyle/islo https://github.com/openclaw/crabbox/pull/781)
or re-check ownership before deleting (multipass/hyperv/kubevirt) are unaffected.

## Instrumented live proof (real Docker via colima)

Committed as opt-in `//go:build smoke` tests
(`internal/providers/localcontainer/cleanup_live_smoke_test.go`), run with:

```
CRABBOX_LIVE=1 go test -tags smoke -run TestLiveLocalContainer -v ./internal/providers/localcontainer/
```

They drive a **real** Docker-compatible runtime and prove both facets against real runtime state.

**(A) Key retention (the availability case).** Create a real container for a lease, register its claim,
write a stored key (as a concurrent Acquire would prepare it before publishing its claim), delete the real
container so the claim is a genuine missing-container orphan, then run the real `Cleanup`. The orphan claim
is removed, but the key is **retained**:

```text
    LIVE PROOF (real Docker runtime): genuine orphan claim removed, key RETAINED for the concurrent Acquire:
        local-container cleanup removed=0 claims_removed=1 checked=0
--- PASS: TestLiveLocalContainerCleanupRetainsKeyForConcurrentAcquire (3.53s)
```

`claims_removed=1` (the genuine orphan claim) with **no key deleted** — so a concurrent Acquire that
reuses this lease keeps its SSH credentials.

**(B) Guard decline.** A reclaim injected during Cleanup's real `docker ps` snapshot window makes the guard
decline; the live lease's claim survives:

```text
    LIVE PROOF (real Docker runtime): guard declined the reclaimed claim during Cleanup's real docker ps window:
        skip claim lease=<lease> slug=livelcguard reason=changed-during-cleanup err=lease <lease> claim changed; retry
        local-container cleanup removed=0 claims_removed=0 checked=0
--- PASS: TestLiveLocalContainerCleanupGuardDeclinesReclaim (0.09s)
```

The smoke skips unless `CRABBOX_LIVE=1` and refuses to run while any other crabbox container is present, so
it never touches unrelated containers; the claim store is an isolated temp dir. Excluded from normal CI by
the `smoke` build tag.

## Review history
https://gist.github.com/377b39f0dadc748db3a8678a38b64c1a


anagnorisis2peripeteia:
@clawsweeper re-review

anagnorisis2peripeteia:
@clawsweeper re-review — fixed the P1 availability bug (orphan sweep now RETAINS the stored key like the merged tart fix #1124; a concurrent Acquire prepares the key before publishing its claim), added TestCleanupRetainsKeyPreparedByConcurrentAcquire + an instrumented live Docker proof (a genuine orphan claim is removed while a concurrent Acquire's key is retained, and the guard declines a mid-cleanup reclaim), hardened the new tests for config-isolation/safety, and reworded the retention comment to cite the merged tart precedent (apple-container retention is the #1146 sibling). Output in the body.

PR #1148, opened 2026-07-17
fix(external): guard Cleanup orphan sweep against a concurrent same-scope Acquire's claim

anagnorisis2peripeteia:
## Summary
Guards the `external` provider's `Cleanup` against a cross-process TOCTOU race that deletes a concurrent
crabbox run's **live** lease claim. `(*leaseBackend).Cleanup` reads the live lease set from the external
helper's protocol `list` invoke, then removes any provider claim on the current external scope that is
absent from that list — via the **unguarded** `core.RemoveLeaseClaim` / `core.RemoveExternalRouting`. A
claim registered by a concurrent `Acquire` on the **same external scope** (a second `crabbox` run against
the same external helper) is visible to the newer claim read but absent from the older `list` snapshot, so
it is swept as "missing" and that live lease's claim (and its routing) is destroyed. The trigger is two
concurrent same-scope `crabbox` runs; no attacker.

Part of https://github.com/openclaw/crabbox/issues/1144. Same `[security]` destructive-cleanup class
already hardened in https://github.com/openclaw/crabbox/pull/1124 (tart),
https://github.com/openclaw/crabbox/pull/446 (aws/azure), and
https://github.com/openclaw/crabbox/pull/781 (freestyle/islo); siblings to the apple-container
(https://github.com/openclaw/crabbox/pull/1146) and local-container
(https://github.com/openclaw/crabbox/pull/1147) fixes under the same umbrella.

## What to review
| file | change |
|---|---|
| `internal/providers/external/backend.go` | +25 / −8 — snapshot orphan-candidate claims **before** the protocol `list` invoke; guard the orphan-sweep **claim** removal with `RemoveLeaseClaimIfUnchanged`, and **retain** the external routing (a concurrent Acquire persists it before publishing its claim) |
| `internal/providers/external/cleanup_toctou_test.go` | +261 — three regression tests: fresh-claim reorder, reclaim guard-decline, and routing-retention against a concurrent Acquire; scope-forced, base-fails / HEAD-passes |
| `internal/providers/external/cleanup_live_smoke_test.go` | +145 — opt-in `//go:build smoke` instrumented live proof: a real external helper subprocess + real routing file through the real Cleanup path (tagged out of normal CI) |

## How it works
Mirrors the merged tart fix (https://github.com/openclaw/crabbox/pull/1124):
- Snapshot orphan-candidate claims via `core.ListLeaseClaims()` **before** the protocol `list` invoke, so
  a claim newer than the list view cannot be misclassified as a "missing" orphan.
- Guard the sweep removal with `core.RemoveLeaseClaimIfUnchanged`, so a lease reclaimed mid-cleanup is
  spared.
- A genuine orphan has its **claim** removed but its **external routing retained**. A concurrent Acquire
  persists the lease's routing (`PersistValidatedExternalRouting`, early in `Acquire`) *before* it publishes
  the replacement claim (`claimLeaseForRepo`), so the old claim can still match our snapshot at the instant
  the routing already belongs to a live reacquisition — deleting it in the sweep would leave that lease
  unroutable (an availability bug). This mirrors the merged tart fix
  (https://github.com/openclaw/crabbox/pull/1124), which retains per-lease state for this reason; the
  apple-container (https://github.com/openclaw/crabbox/pull/1146) and local-container
  (https://github.com/openclaw/crabbox/pull/1147) siblings apply the same retention. A genuinely dead
  lease's routing is a harmless small residue.

**Scope note:** the external sweep is scope-filtered (`claim.ProviderScope != b.claimScope()`), so the
race is gated to two runs sharing the same external-provider scope. The regression tests force a non-empty
scope so the scope-matched path is actually exercised.

This is a **timing race** a happy-path lifecycle run cannot reliably trigger (the external `Acquire`
publishes its claim only after the helper round-trip, outside the sweep's snapshot→remove window), so a
full third-party-provider lifecycle is deferred to maintainer CI (as with
https://github.com/openclaw/crabbox/pull/1124). The race is reproduced **deterministically** with three
regression tests (the snapshot reorder, the guard-decline, and the routing-retention availability case)
that FAIL on base `cf5081fc` and PASS with this change — and an **instrumented live proof against a real
helper subprocess** is included below (see "Instrumented live proof"):

```
# BASE (this change stashed) — the sweep deletes the concurrent same-scope claim, and deletes routing a
# concurrent Acquire persisted:
$ go test -race -run 'TestCleanupOrphanSweep|TestCleanupRetainsRouting' ./internal/providers/external/
--- FAIL: TestCleanupOrphanSweepDeletesConcurrentAcquireClaim
    Cleanup's orphan sweep deleted the claim registered by a concurrent same-scope Acquire: present=false (want present)
--- FAIL: TestCleanupOrphanSweepGuardDeclinesReclaimedCandidate
    orphan sweep removed a lease reclaimed during Cleanup: present=false (want present)
--- FAIL: TestCleanupRetainsRoutingForConcurrentAcquire
    cleanup deleted external routing persisted by a concurrent Acquire (availability regression)
FAIL

# HEAD (this change) — all three pass; full package suite green:
$ go test -race ./internal/providers/external/
ok  github.com/openclaw/crabbox/internal/providers/external
```

`TestCleanupOrphanSweepDeletesConcurrentAcquireClaim` exercises the reorder (a fresh concurrent claim);
`TestCleanupOrphanSweepGuardDeclinesReclaimedCandidate` exercises the `RemoveLeaseClaimIfUnchanged` guard
(a reclaim mid-cleanup). `go build ./...`, `go vet ./...`, `gofmt -l`, and the full `go test -race ./...`
suite are clean.

## Not included (scope)
apple-container is fixed in https://github.com/openclaw/crabbox/pull/1146 and local-container in
https://github.com/openclaw/crabbox/pull/1147. Providers that already guard the sweep (tart
https://github.com/openclaw/crabbox/pull/1124, aws/azure https://github.com/openclaw/crabbox/pull/446,
freestyle/islo https://github.com/openclaw/crabbox/pull/781) or re-check ownership before deleting
(multipass/hyperv/kubevirt) are unaffected.

## Instrumented live proof (real external helper subprocess)

Committed as an opt-in `//go:build smoke` test
(`internal/providers/external/cleanup_live_smoke_test.go`), run with:

```
CRABBOX_LIVE=1 go test -tags smoke -run TestLiveExternal -v ./internal/providers/external/
```

Unlike the hermetic tests (which return canned JSON from an in-process fake), this drives a **real external
helper subprocess** that implements the versioned JSON protocol on stdin/stdout, plus a **real routing file
on disk**, through the real `Cleanup` code path. A genuine missing-lease orphan claim is removed while the
external routing a concurrent Acquire persisted before publishing its claim is **retained**:

```text
--- RUN   TestLiveExternalCleanupRetainsRoutingForConcurrentAcquire
    LIVE PROOF (real external helper subprocess + real routing file): genuine orphan claim removed, external routing RETAINED for the concurrent Acquire:
        external-helper: handled protocol operation=cleanup (returning 0 live leases)
        external-helper: handled protocol operation=list (returning 0 live leases)
--- PASS: TestLiveExternalCleanupRetainsRoutingForConcurrentAcquire (0.27s)
ok  github.com/openclaw/crabbox/internal/providers/external
```

The helper subprocess genuinely received and answered both protocol operations (`cleanup` then `list`),
reporting zero live leases so the pre-existing claim is a real missing-lease orphan; `Cleanup` removed that
claim but left the routing file on disk. The smoke skips unless `CRABBOX_LIVE=1`, isolates
`HOME`/`XDG_CONFIG_HOME`/`XDG_STATE_HOME` to temp dirs (`isolateCrabboxState`) so it never touches real
config, and is excluded from normal CI by the `smoke` build tag.

## Review history
https://gist.github.com/a80a0a68d1a3e81ec0ad688b9dc38284


anagnorisis2peripeteia:
@clawsweeper re-review

anagnorisis2peripeteia:
@clawsweeper re-review — added an instrumented live proof: an opt-in `//go:build smoke` test drives a **real external helper subprocess** implementing the JSON protocol (handled `cleanup`+`list`, 0 live leases) plus a real routing file on disk, through the real `Cleanup` path — a genuine missing-lease orphan claim is removed while the routing a concurrent Acquire persisted is **retained**. Output is in the body under 'Instrumented live proof'. (The earlier B/B/A review already validated the patch, reproduced the base-fails, and cleared security; proof was mock_only only for lack of a live helper, now addressed.)

PR #1149, opened 2026-07-17
fix(apple-container): require an exact claim before state-based cleanup (P0 data loss)

### LINKED ISSUE #1132: [security] Apple-container cleanup deletes unclaimed stopped containers with incomplete ownership labels
# [security] Apple-container cleanup deletes unclaimed stopped containers with incomplete ownership labels

## Summary

`crabbox cleanup --provider apple-container` can force-delete an unrelated stopped container based only on `crabbox=true` and `provider=apple-container`. These labels admit the container to discovery, and the stopped-state branch authorizes deletion before cleanup checks for a local lease claim.

Consequently, a stopped container with the namespace labels but no lease identifier or exact resource-bound claim is deleted rather than skipped. This violates the defensive ownership boundary used for other destructive lifecycle operations.

## Affected Components

- `internal/providers/applecontainer/backend.go:543-559` selects containers using only the `crabbox` and `provider` labels.
- `internal/providers/applecontainer/backend.go:860-885` returns a positive cleanup decision for every non-running/non-ready container before evaluating `hasClaim`.
- `internal/providers/applecontainer/backend.go:321-365` executes `container delete --force` for that decision without a subsequent ownership check.
- `internal/providers/applecontainer/backend_test.go:394-416` is an upstream inventory-filter test, but it does not exercise destructive cleanup or incomplete ownership metadata.
- `docs/security.md:343-369` documents the broader fail-closed resource-ownership model and treats discovery metadata as insufficient destructive ownership proof.

## Attack Path

1. A local process or user able to create containers in the same Apple container runtime creates or leaves a stopped container labeled `crabbox=true` and `provider=apple-container`. No Crabbox lease label or local claim is required.
2. The victim runs `crabbox cleanup --provider apple-container` without `--dry-run`.
3. `listContainers` admits the container based on the two labels.
4. `shouldCleanup` sees `status=stopped` and returns `true` before reaching its missing-claim branch.
5. Cleanup invokes `container delete --force <id>`.

The issue is limited to the local Apple-container provider and requires an explicit cleanup invocation. `--dry-run` prevents deletion, `keep=true` prevents deletion, and unclaimed containers reported as `running` or `ready` are skipped. Those countercontrols do not protect a stopped container with the two discovery labels.

## Impact

The cleanup command can irreversibly delete a container that Crabbox has not established it owns. This destroys container-local writable state and prevents the container from being restarted, causing integrity and availability loss. The behavior does not demonstrate access to container contents, deletion of unrelated host files or external volumes, privilege escalation, or a remote attack path.

### CVSS Assessment

| Metric | v3.1 | v4.0 |
|--------|------|------|
| **Score** | 6.6 / 10.0 | 5.2 / 10.0 |
| **Vector** | CVSS:3.1/AV:L/AC:L/PR:L/UI:R/S:U/C:N/I:H/A:H | CVSS:4.0/AV:L/AC:L/AT:N/PR:L/UI:A/VC:N/VI:H/VA:H/SC:N/SI:N/SA:N |
| **Calculator** | [CVSS v3.1 Calculator](https://www.first.org/cvss/calculator/3.1#CVSS:3.1/AV:L/AC:L/PR:L/UI:R/S:U/C:N/I:H/A:H) | [CVSS v4.0 Calculator](https://www.first.org/cvss/calculator/4.0#CVSS:4.0/AV:L/AC:L/AT:N/PR:L/UI:A/VC:N/VI:H/VA:H/SC:N/SI:N/SA:N) |

The scores reflect a local attack requiring privileges sufficient to place a labeled container in the same runtime and active victim interaction to run cleanup. Scope remains within that local runtime, confidentiality is unaffected, and forced deletion can have high integrity and availability impact. Material preconditions and the `--dry-run`, `keep=true`, and running-state countercontrols reduce practical exploitability but do not restore ownership validation.

Require destructive Apple-container cleanup to establish ownership before considering container state. Prefer an exact local claim whose provider, lease ID, and bound container ID match the current candidate, and revalidate that binding immediately before deletion.

Treat the two namespace labels as discovery hints only. If reclaiming containers left behind before claim persistence is required, place that behavior behind an explicit adoption/reclaim operation or another durable ownership binding. At minimum, incomplete candidates lacking a lease identifier must be skipped rather than inferred to be Crabbox-owned from the namespace labels alone.

Add regression coverage for an unclaimed stopped container with only the namespace labels, a conflicting claim, and an exact matching claim. Preserve coverage for `keep=true`, dry-run operation, expired claims, and intentional orphan-recovery behavior.

## Validation

A temporary focused unit test created only for local validation passed a stopped server with the two namespace labels and `hasClaim=false` to `shouldCleanup`. It failed because the function returned `true` with `reason="container state=stopped"` instead of `false` with `reason="missing claim"`.

A second temporary mocked cleanup test supplied `container ls` output containing a stopped `unclaimed-container` with the two labels and no claim. Cleanup issued exactly `container delete --force unclaimed-container`, confirming the destructive command path. These temporary tests are not present upstream and were removed after validation.

The existing upstream inventory-filter and unclaimed-release tests passed with:

`go test -count=1 -run '^(TestListContainersFiltersByLabel|TestReleaseLeaseRejectsUnclaimedContainer)$' ./internal/providers/applecontainer`

A full unmodified package run was also attempted but was not green because the unrelated upstream `TestAcquirePostCreateFailureKeepsRetainedContainerKey` test stopped after its mocked list call and did not record its expected run call. That failure did not exercise cleanup and does not affect the focused reproduction.

steipete:
## Summary

- require an exact local claim matching provider, lease ID, and container ID before Apple-container cleanup considers container state
- preserve unclaimed or conflicting stopped containers while retaining cleanup of exactly claimed stopped containers
- document the fail-closed cleanup behavior and add the security fix to the changelog

## Security impact

The `crabbox=true` and `provider=apple-container` labels remain discovery hints, but no longer authorize `container delete --force`. Cleanup now fails closed before state-based deletion unless the local claim is resource-bound to the current container.

No config, secret, or migration changes.

This fix is intentionally separate from https://github.com/openclaw/crabbox/issues/1144 and https://github.com/openclaw/crabbox/issues/1146.

Fixes https://github.com/openclaw/crabbox/issues/1132

## Regression proof

Before the fix, `TestCleanupPreservesUnclaimedStoppedAppleContainer` failed after recording:

```text
container delete --force unclaimed-container
```

After the fix, the regression test passes. Additional coverage verifies that a conflicting resource-bound claim is skipped and an exact matching claim still permits stopped-container cleanup.

## Verification

```text
go build ./...
go vet ./...
go test -race ./internal/providers/applecontainer/... -count=1
```

Codex autoreview:

```text
.agents/skills/autoreview/scripts/autoreview --mode branch --base origin/main
autoreview clean: no accepted/actionable findings reported
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 
 ### Fixed
 
+- Prevented Apple-container cleanup from force-deleting stopped containers without an exact resource-bound local claim. Thanks @coygeek.
 - Limited failed Blacksmith warmup cleanup to Testbox IDs emitted by that invocation, preventing config-matched concurrent Testboxes from being stopped. Thanks @anagnorisis2peripeteia.
 - Prevented Tart cleanup from deleting lease claims and stored SSH keys created or rebound by concurrent acquisitions. Thanks @anagnorisis2peripeteia.
 - Failed Node coordinator startup on malformed trusted-proxy CIDRs, warned on untrusted forwarded client headers, and kept environment-backed provider failures out of top-level request logs.
--- a/docs/providers/apple-container.md
+++ b/docs/providers/apple-container.md
@@ -137,9 +137,9 @@ variable consumed by the bootstrap script.
 6. `status`, `list`, and `stop` inspect or remove labeled containers via
    `container ls --all --format json`, `container inspect`, and
    `container delete --force`.
-7. `cleanup --provider apple-container` removes stopped containers and running
-   non-`keep` containers whose local claim is stale past the idle timeout plus a
-   safety grace period, and prunes orphaned claims.
+7. `cleanup --provider apple-container` removes exactly claimed stopped containers
+   and exactly claimed running non-`keep` containers whose local claim is stale
+   past the idle timeout plus a safety grace period, and prunes orphaned claims.
 
 ## Limits and caveats
 
--- a/internal/providers/applecontainer/backend.go
+++ b/internal/providers/applecontainer/backend.go
@@ -865,24 +865,29 @@ func shouldCleanup(server core.Server, claim core.LeaseClaim, hasClaim bool, now
 	if strings.EqualFold(labels["keep"], "true") {
 		return false, "keep=true"
 	}
+	if !hasClaim {
+		return false, "missing claim"
+	}
+	leaseID := strings.TrimSpace(labels["lease"])
+	containerID := strings.TrimSpace(server.CloudID)
+	if claim.Provider != providerName || claim.LeaseID != leaseID || leaseID == "" || strings.TrimSpace(claim.CloudID) != containerID || containerID == "" {
+		return false, "claim mismatch"
+	}
 	if !strings.EqualFold(server.Status, "running") && server.Status != "ready" {
 		return true, "container state=" + blank(server.Status, "unknown")
 	}
-	if hasClaim {
-		lastUsed, err := time.Parse(time.RFC3339, claim.LastUsedAt)
-		if err != nil || lastUsed.IsZero() {
-			return false, "claim active"
-		}
-		idle := time.Duration(claim.IdleTimeoutSeconds) * time.Second
-		if idle <= 0 {
-			return false, "claim active"
-		}
-		if now.After(lastUsed.Add(idle).Add(12 * time.Hour)) {
-			return true, "claim expired"
-		}
+	lastUsed, err := time.Parse(time.RFC3339, claim.LastUsedAt)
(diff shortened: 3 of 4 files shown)


steipete:
Already addressed in 25842310: both new cleanup regression tests now skip on non-darwin/arm64 hosts (`if runtime.GOOS != "darwin" || runtime.GOARCH != "arm64" { t.Skip(...) }`), matching the existing apple-container tests. The Go job passes on Linux (tests skip) and runs+passes on macOS.

PR #1151, opened 2026-07-17
fix(apple-vm): guard Cleanup orphan sweep against a concurrent same-scope Acquire's claim

anagnorisis2peripeteia:
## Summary
Guards the `apple-vm` provider's `Cleanup` against a cross-process TOCTOU race that can delete a
concurrent crabbox run's **live** lease claim and strand its prepared SSH key. `(*backend).Cleanup` lists
VM instances via `listInstances()` **before** it reads lease claims, then its orphan sweep removes — via
the **unguarded** `core.RemoveLeaseClaim` plus a stored-key deletion — any apple-vm claim whose lease is
absent from that stale instance view. In the read-to-remove window, a claim/key that a concurrent
`Acquire` prepares before publishing (Acquire creates the instance and writes the SSH key before its claim
is visible) can be swept as "missing instance." The trigger is two concurrent same-scope `crabbox` runs on
one Mac (apple-vm is the local provider); no attacker.

Part of https://github.com/openclaw/crabbox/issues/1144 — apple-vm was missed in the original audit
because its `Cleanup` reads claims through a `providerClaims()` wrapper rather than a direct
`core.ListLeaseClaims()` call, so the list-symbol scan never flagged it (the correct audit key is the
unguarded `core.RemoveLeaseClaim`, not the list op). Same `[security]` destructive-cleanup class already
merged for tart (https://github.com/openclaw/crabbox/pull/1124), apple-container
(https://github.com/openclaw/crabbox/pull/1146), local-container
(https://github.com/openclaw/crabbox/pull/1147), and external
(https://github.com/openclaw/crabbox/pull/1148).

## What to review
| file | change |
|---|---|
| `internal/providers/applevm/backend.go` | +25 / −4 — snapshot orphan-candidate claims **before** `listInstances()`; guard the orphan-sweep **claim** removal with `RemoveLeaseClaimIfUnchanged` (and `VerifyLeaseClaimUnchanged` in dry-run), and **retain** the stored testbox key |
| `internal/providers/applevm/cleanup_toctou_test.go` | +497 — five regression tests (reclaim guard-decline, key-retention, dry-run no-plan, dry-run genuine-removal, foreign-provider skip), scope-forced |
| `internal/providers/applevm/cleanup_live_smoke_test.go` | +287 — `//go:build smoke` instrumented live proof driving the real `crabbox-apple-vm-helper` (see Live proof below) |
| `CHANGELOG.md` | +1 — apple-vm entry under `0.39.1 - Unreleased → Fixed`, matching the three siblings |

## How it works
Mirrors the merged apple-container fix (https://github.com/openclaw/crabbox/pull/1146):
- Snapshot orphan-candidate claims via `core.ListLeaseClaims()` **before** `listInstances()`, so a claim
  newer than the instance view cannot be misclassified as a "missing instance" orphan.
- Guard the sweep's claim removal with `core.RemoveLeaseClaimIfUnchanged` (and `VerifyLeaseClaimUnchanged`
  in `--dry-run`), so a lease reclaimed mid-cleanup is spared and dry-run does not plan its removal.
- **Retain** the stored testbox key rather than deleting it in the sweep: `Acquire` creates or reuses the
  key before publishing its claim, outside this claim CAS, so deleting it here can strand a concurrent
  live reacquisition. The instance-removal path (which deletes an instance that is actually present) still
  removes the key, unchanged.

**Startup-grace note:** apple-vm's sweep already skips claims within `unclaimedInstanceGrace` (3h). Because
a reclaim rewrites `LastUsedAt = now`, a claim reclaimed *after* the pre-fix code has already listed
instances lands back inside that grace and is skipped — so the live-claim-deletion window this closes is
the narrow one where a claim/key exists at listing time but its instance does not yet, which the CAS guard
+ key retention cover. The tests therefore backdate their orphan claims past the grace so the guarded path
is actually exercised.

Three regression tests FAIL on base `5f6d7220` (with this test file applied) and PASS at HEAD, verified
under `-race`:

```
$ go test -race -run 'TestCleanupOrphanSweepGuardDeclinesReclaimedCandidate|TestCleanupRetainsKeyPreparedByConcurrentAcquire|TestCleanupDryRunDoesNotPlanReclaimedCandidateRemoval' ./internal/providers/applevm/

# BASE 5f6d7220:  --- FAIL (x3)
#   RetainsKey…                  base deletes the prepared key ("prepared key was removed by cleanup") — the destructive defect
#   GuardDeclinesReclaimed…      base grace-skips the reclaimed claim ("reason=startup grace period"); no changed-during-cleanup disposition
#   DryRunDoesNotPlanReclaimed…  same: dry-run does not emit the reclaimed-skip disposition
# HEAD:  ok  github.com/openclaw/crabbox/internal/providers/applevm
```

The destructive base failure the hermetic tests pin is the **stored-key deletion** (the reclaim-guard and
dry-run tests lock in the guard's `changed-during-cleanup` disposition, which base lacks). `go build ./...`,
`go vet` on both touched packages, and `gofmt -l` are clean; the full apple-vm package passes with the race
detector.

**Mutation coverage (gomu, diff-scoped):** 19 mutants on the changed lines. The two dry-run/foreign-provider
tests were added to close real survivors — they kill the sweep's provider-filter mutants (so a broken filter
can't sweep another provider's claim) and the dry-run guard's `would remove` path; each kill is confirmed by
direct mutant injection. The residual survivors are equivalent (`LeaseID == "" ` vs `<= ""`), unreachable
(empty-lease), or fail-safe (swallowing a `ListLeaseClaims` error sweeps nothing — no wrong removal).

## Live proof (instrumented, against the real helper)
`internal/providers/applevm/cleanup_live_smoke_test.go` (`//go:build smoke`, `CRABBOX_LIVE=1`) drives the
**real** `crabbox-apple-vm-helper` binary — built from `cmd/crabbox-apple-vm-helper` and exec'd exactly as
Cleanup shells out to it. Cleanup's only helper calls are `list` and `delete`, both of which read/remove
on-disk instance metadata, so **no Virtualization.framework VM is booted** and the guarded orphan-sweep runs
end-to-end against the real helper subprocess on any Apple-silicon Mac (entitlements/base image absent). The
genuine orphan is a claim whose lease has no instance metadata, so the real `list` reports
`{"instances":null}` — a real "missing instance" by the helper's own view. Two sub-proofs; the reclaim is a
controlled rebind injected during the real `list` window (a blind timing race can't hit the sub-200ms
window on its own). All temp-isolated HOME/XDG/state-root, so real Cleanup can only ever see this test's own
claim:

```
$ CRABBOX_LIVE=1 go test -tags smoke -run TestLiveAppleVMCleanup -v ./internal/providers/applevm/
--- PASS: TestLiveAppleVMCleanupGuardPreservesReclaimedClaim (0.50s)
    LIVE PROOF (real crabbox-apple-vm-helper): guard declined the reclaimed claim during Cleanup's real `list` window:
        skip claim lease=cbx_livesmoke_guard reason=changed-during-cleanup err=lease cbx_livesmoke_guard claim changed; retry
        apple-vm cleanup removed=0 claims_removed=0 checked=0
--- PASS: TestLiveAppleVMCleanupRetainsKeyOnGenuineOrphanRemoval (0.49s)
    LIVE PROOF (real crabbox-apple-vm-helper): genuine orphan removed, prepared key retained:
        remove claim lease=cbx_livesmoke_retain reason=missing instance
        apple-vm cleanup removed=0 claims_removed=1 checked=0
PASS
ok  	github.com/openclaw/crabbox/internal/providers/applevm	1.408s
```

## Not included (scope)
- No live proof exercises a booted VM: Cleanup never calls `start`, so the orphan-sweep path this PR touches
  has no VM dependency; the live proof drives the real helper's `list`/`delete` instead. A full VF boot lane
  stays with the maintainer's provisioned CI.
- The instance-removal path's unguarded `RemoveLeaseClaim` + key deletion is correct and unchanged (it
  deletes an instance that is actually present); it matches the merged siblings' scope.
- The other providers (tart, apple-container, local-container, external, aws/azure, freestyle/islo) are
  already fixed/merged. This PR is the apple-vm-specific repair under
  https://github.com/openclaw/crabbox/issues/1144.

## Review history
https://gist.github.com/9022e6f1fda2ba84c6bd7c07b184902c

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Prevented apple-container orphan cleanup from deleting claims reclaimed during its resource snapshot, including same-value rewrites, while retaining stored SSH keys when ownership cannot be proven safe to remove. Thanks @anagnorisis2peripeteia.
 - Prevented local-container orphan cleanup from deleting claims reclaimed during its resource snapshot while retaining stored SSH keys when ownership cannot be proven safe to remove. Thanks @anagnorisis2peripeteia.
 - Prevented external-provider orphan cleanup from deleting claims reclaimed during its resource snapshot while retaining routing state when ownership cannot be proven safe to remove. Thanks @anagnorisis2peripeteia.
+- Prevented apple-vm orphan cleanup from deleting claims reclaimed during its resource snapshot while retaining stored SSH keys when ownership cannot be proven safe to remove. Thanks @anagnorisis2peripeteia.
 
 ## 0.39.0 - 2026-07-17
 
--- a/docs/features/provider-live-smoke.md
+++ b/docs/features/provider-live-smoke.md
@@ -155,7 +155,7 @@ hermetic lifecycle tests, `scripts/live-smoke.sh`, dedicated live runners, and
 `//go:build smoke` tests. Regenerate it with
 `node scripts/generate-provider-matrix.mjs`; docs CI rejects drift.
 
-Current coverage: 76 providers; 4 with convention-named hermetic lifecycle tests, 56 with a live runner, 7 with tagged Go smoke tests, and 19 with none of those lifecycle surfaces.
+Current coverage: 76 providers; 4 with convention-named hermetic lifecycle tests, 56 with a live runner, 8 with tagged Go smoke tests, and 19 with none of those lifecycle surfaces.
 
 | Provider | Hermetic lifecycle | Live runner | Tagged Go smoke |
 | --- | --- | --- | --- |
@@ -163,7 +163,7 @@ Current coverage: 76 providers; 4 with convention-named hermetic lifecycle tests
 | [anthropic-sandbox-runtime](../providers/anthropic-sandbox-runtime.md) | — | dedicated + matrix | — |
 | [apple-container](../providers/apple-container.md) | — | matrix | yes |
 | [apple-machine](../providers/apple-machine.md) | — | — | — |
-| [apple-vm](../providers/apple-vm.md) | — | matrix | — |
+| [apple-vm](../providers/apple-vm.md) | — | matrix | yes |
 | [ascii-box](../providers/ascii-box.md) | — | — | — |
 | [aws](../providers/aws.md) | — | matrix | — |
 | [aws-lambda-microvm](../providers/aws-lambda-microvm.md) | — | dedicated + matrix | — |
--- a/internal/providers/applevm/backend.go
+++ b/internal/providers/applevm/backend.go
@@ -385,6 +385,14 @@ func (b *backend) Cleanup(ctx context.Context, req core.CleanupRequest) error {
 	if err := requireHost(); err != nil {
 		return err
 	}
+	// Snapshot orphan-candidate claims before listing instances so a claim
+	// registered by a concurrent Acquire cannot be compared against an older
+	// instance view and misclassified as a "missing instance" orphan. Only claims
+	// that predate our instance view can be genuine orphans.
+	orphanCandidates, err := core.ListLeaseClaims()
+	if err != nil {
+		return err
+	}
 	instances, err := b.listInstances(ctx, cfg)
 	if err != nil {
 		return err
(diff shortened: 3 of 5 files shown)


anagnorisis2peripeteia:
@clawsweeper review

PR #1174, opened 2026-07-20
docs: add pi coding agent to the Agent Skills compatibility catalog

zozo123:
## What Problem This Solves

Contributors using the **pi** coding agent ([earendil-works/pi](https://github.com/earendil-works/pi)) had no entry in the AI-agents integration catalog, so it was unclear whether `crabbox init`'s generated Agent Skill is discoverable by pi or whether an extra skill copy is required. The `docs/integrations/agents.md` compatibility table — the audited source of truth for which Agent Skills clients find the default `.agents/skills` path — omitted pi entirely.

## Why This Change Was Made

pi is a terminal coding-agent harness that implements the Agent Skills standard. It discovers project skills at the default `.agents/skills` path (from the working directory up through the git-repo root), so the Skill written by `crabbox init` works with pi without an additional copy — the same situation as Zed and OpenCode. This is a docs-only, integration-only change per the [authoring contract](https://github.com/openclaw/crabbox/blob/main/docs/integrations/authoring.md); no core, CLI, or `SKILL.md` changes are involved, so there is no canonical/projection drift.

The new row records the facts a contributor needs:
- discovers the default `.agents/skills` path (cwd → git-repo root);
- also scans the native `.pi/skills` project path and the global `~/.pi/agent/skills` and `~/.agents/skills` roots;
- invoke with `/skill:crabbox` after trusting the workspace; slash-command invocation is gated by the `enableSkillCommands` setting, and `--no-skills` disables discovery.

The audit-date note is updated to record that pi's documentation was audited on 2026-07-20.

## User Impact

pi users can now see, from the integration catalog, that `crabbox init` output is discovered automatically and how to invoke it — no speculative extra skill copy needed.

## Evidence

Facts cross-verified against three authoritative pi sources (pi.dev/docs/latest/skills, the repo's `packages/coding-agent/docs/skills.md`, and its README), which agree on all discovery paths, invocation, trust requirement, and gating.

Docs CI gates run locally:

```
$ node scripts/check-docs-links.mjs
checked 233 markdown files: internal links ok

$ node scripts/check-agent-skills.mjs
validated publishable Crabbox Agent Skill: 477 lines, canonical and projection identical
```

### DIFF
--- a/docs/integrations/agents.md
+++ b/docs/integrations/agents.md
@@ -45,7 +45,8 @@ the directory containing `SKILL.md`.
 
 The [Agent Skills standard](https://agentskills.io/specification) defines
 `SKILL.md`, not a mandatory search path. Official client documentation was
-audited on 2026-07-18; host behavior can change independently of Crabbox.
+audited on 2026-07-18 (pi on 2026-07-20); host behavior can change
+independently of Crabbox.
 
 | Client | Discovers the default `.agents/skills` path | Notes |
 | --- | --- | --- |
@@ -58,6 +59,7 @@ audited on 2026-07-18; host behavior can change independently of Crabbox.
 | [Gemini CLI](https://geminicli.com/docs/cli/using-agent-skills/) | Yes, from the workspace root | Launch Gemini from the repository root, trust the workspace, then use `/skills list` or `/skills reload`. |
 | [Google Antigravity](https://antigravity.google/docs/skills) | Yes | `.agent/skills` is a legacy alternative. |
 | [OpenCode](https://opencode.ai/docs/skills/) | Yes | Walks from the current directory to the worktree root; [`opencode debug skill`](https://opencode.ai/docs/cli/) shows the resolved inventory. |
+| [pi](https://pi.dev/docs/latest/skills) | Yes, from cwd through the git-repo root | Also scans the native `.pi/skills` project path and the global `~/.pi/agent/skills` and `~/.agents/skills` roots. Trust the workspace, then invoke with `/skill:crabbox`; slash-command invocation is gated by the `enableSkillCommands` setting, and `--no-skills` disables discovery. |
 
 The same shared path is also documented by [Roo Code](https://docs.roocode.com/features/skills),
 [Amp](https://ampcode.com/manual#agent-skills),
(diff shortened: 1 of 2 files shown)


zozo123:
## Heads-up: the red `Go` check was an unrelated flake — now green ✅

This PR is **docs-only** (`docs/integrations/agents.md`, +3/−1), but CI was red on the **`Go`** job. For the record, here's the end-to-end diagnosis and fix.

### What was failing

```
--- FAIL: TestEgressDaemonConcurrentStartDoubleProvision
    egress_daemon_double_provision_test.go:125: iteration 2: supervisor pid … remained alive after stop
FAIL	github.com/openclaw/crabbox/internal/cli
```

A process-lifecycle timing test in `internal/cli` — nothing to do with the docs change.

### Root cause

`stopDaemonProcess` (`internal/cli/daemon_unix.go`) signals the supervisor process group with **`SIGKILL`, which the kernel delivers asynchronously**, then returns immediately:

```go
func stopDaemonProcess(process *os.Process, pid int) error {
	if err := syscall.Kill(-pid, syscall.SIGKILL); err != nil && err != syscall.ESRCH {
		return process.Kill()
	}
	return nil // returns before the kill has actually torn the process down
}
```

So `stopEgressHostDaemon` can return `stopped=true` while the supervisor is still tearing down. The test then asserted liveness **the instant `stop` returned**, so under CI load it occasionally observed the not-yet-dead process and failed. Synchronous assertion on an inherently asynchronous outcome.

### Fix (`81ddbfb8`)

Poll for teardown with a bounded 3s timeout instead of checking synchronously:

```go
for _, pid := range pids {
	deadline := time.Now().Add(3 * time.Second)
	for egressDaemonTestAlive(pid) && time.Now().Before(deadline) {
		time.Sleep(20 * time.Millisecond)
	}
	if egressDaemonTestAlive(pid) {
		t.Fatalf("iteration %d: supervisor pid %d remained alive 3s after stop", i, pid)
	}
}
```

Deterministic — `SIGKILL` is guaranteed to terminate the process, and a reaped/zombie process already reads as dead via `egressDaemonTestAlive`. **Test-only; production `stop` semantics are unchanged.** `gofmt`-clean.

### Result

All checks green on `81ddbfb8` — `Go`, `Docs`, `Scripts`, `Release Check`, `Worker`, `Apple VM`, `Connector Lifecycles`, and the sandbox/mesh jobs. Merge state is **CLEAN**.

> **Note on scope:** this adds one test file to an otherwise docs-only PR. Happy to split the flake fix into its own PR if the maintainers would prefer to keep this one strictly docs — just say the word.

PR #1191, opened 2026-07-24
feat(tart): support random clone serials

malpern:
Adds an opt-in `--tart-random-serial` setting (plus YAML and environment configuration) that runs `tart set <vm> --random-serial` before other clone configuration.

This lets managed macOS clones receive unique hardware identities instead of inheriting the base VM identity.

Tests:
- `go test ./internal/providers/tart`
- focused Tart config file/environment tests

Note: the full `internal/cli` package test run was interrupted after 147s because an unrelated long-running test did not complete; the two config tests affected by this change pass directly.

### DIFF
--- a/docs/providers/tart.md
+++ b/docs/providers/tart.md
@@ -55,11 +55,18 @@ tart:
   cpus: 4
   memory: 8192
   # disk: 80  # only set to resize beyond the base image default
+  # usbPassthrough: true  # macOS 27+ host; requires a signed Tart build
 ```
 
 Environment variables: `CRABBOX_TART_IMAGE`, `CRABBOX_TART_USER`,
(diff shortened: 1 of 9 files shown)


steipete:
Thanks for the contribution. I am closing this combined branch because it carries three independently reviewable product changes: Tart random serials, macOS VNC double-click behavior, and USB passthrough. The USB slice also adds a physical-device access boundary, so these should not share one merge contract.

Please extract the random-serial improvement into a fresh focused PR from current `main`. Keep only the `tart set <vm> --random-serial` configuration path, its CLI/environment/YAML plumbing, focused tests, and provider documentation. The YAML example should explicitly document `tart.randomSerial`, and the PR should include redacted live Tart proof showing that an opted-in clone receives a different serial from its source image.

Please leave USB passthrough and VNC double-click out of that PR; either can be proposed separately with its own behavior and security proof.

PR #1203, opened 2026-07-28
fix(runpod): make new pods accept the configured SSH key

### LINKED ISSUE #1202: RunPod pods cannot authenticate with the configured SSH key
## Summary

With `provider: runpod`, Crabbox can create a pod that reaches `RUNNING` and exposes a public TCP mapping for port 22, but SSH key authentication fails even when `ssh.key` points to a private key with a valid adjacent `.pub` file.

In two Crabbox 0.38.0 provisioning attempts, the created pods exposed port 22 but reported `PUBLIC_KEY` as the literal string `null`. SSH readiness could not authenticate with the configured key.

## Reproduction

Configure a RunPod provider and a local SSH key:

```yaml
provider: runpod
ssh:
  key: /path/to/id_ed25519
```

Ensure `/path/to/id_ed25519.pub` exists, then create a RunPod lease:

```sh
crabbox warmup --provider runpod --keep
```

The pod reaches `RUNNING` with a public port 22 mapping, but the configured key is not authorized and SSH readiness cannot complete.

## Root cause

The RunPod acquire path sets `StartSSH: true` on its internal deploy input, but that field is not serialized by `runpodDeployPayload`. The configured `ssh.key` is only used later by the local SSH client; its public half is never added to the `POST /pods` request.

The request therefore has no `env.PUBLIC_KEY` value:

```text
ssh.key -> local SSH identity
        \-> (missing) POST /pods env.PUBLIC_KEY
```

RunPod's official container startup script only appends `PUBLIC_KEY` to `~/.ssh/authorized_keys` and starts SSH when that value is present:

https://github.com/runpod/containers/blob/0e47720fce6ccf75703e5f5068e90dc817492ae7/container-template/start.sh

The RunPod REST API supports an `env` object on pod configuration:

https://docs.runpod.io/api-reference/pods/POST/pods

This makes the failure deterministic at the request boundary: a configured Crabbox key cannot become an authorized pod key because its public value is absent from the deploy payload.

## Expected behavior

Before creating a paid pod, Crabbox should read the public half of the configured SSH key and fail clearly if it is missing or empty. The RunPod deploy request should serialize that value as `env.PUBLIC_KEY`, allowing the official image to populate `authorized_keys` and start SSH for the same identity Crabbox uses locally.

## Impact

Affected RunPod leases can consume provisioning time and reach a nominally running state while remaining unusable to Crabbox's SSH transport. Exporting `PUBLIC_KEY` in the caller's shell does not repair this path because the provider constructs the pod environment explicitly.

morluto:
Closes https://github.com/openclaw/crabbox/issues/1202

## What Problem This Solves

Fixes an issue where users provisioning RunPod leases with a configured `ssh.key` would receive a running pod with an exposed SSH port but no matching authorized key. The pod could report `PUBLIC_KEY` as `null`, leaving Crabbox's SSH readiness check unable to authenticate.

## Why This Change Was Made

The RunPod provider now reads the configured key's adjacent `.pub` file before creating the pod and serializes that exact value as `env.PUBLIC_KEY` in the `POST /pods` request. Missing or empty public-key files fail before the paid resource is created.

This keeps key ownership unchanged: the private key remains local, and only its public half is included in the provider request.

```text
ssh.key.pub
    |
    v
RunPod POST /pods
  env.PUBLIC_KEY
    |
    v
official image startup -> ~/.ssh/authorized_keys -> sshd
```

## User Impact

RunPod pods created by Crabbox can authorize the same SSH identity that Crabbox uses for readiness, sync, and command execution. Invalid local key configuration now fails before pod creation instead of producing an unreachable running lease.

## Evidence

- Two Crabbox 0.38.0 provisioning attempts reached `RUNNING` with public port 22 but could not authenticate; the pod environment showed `PUBLIC_KEY` as `null`.
- The regression test proves the configured `.pub` value reaches the acquire request and JSON serialization preserves it as `env.PUBLIC_KEY`.
- `go test ./internal/providers/runpod -count=1` — pass.
- `go test -race ./internal/providers/runpod ./internal/providers/all -count=1` — pass.
- `go vet ./internal/providers/runpod ./internal/providers/all` — pass.
- `go build -trimpath -o /tmp/crabbox-runpod-key-check ./cmd/crabbox` — pass.

## Live provider proof

Validated commit `1a3b1aabb466ef48bcbf46b500de9e6d594e1037` with the patched Crabbox binary and a disposable SSH key created only for this test.

The one-shot RunPod lease completed SSH readiness, repository sync, and a remote command:

```text
provider=runpod
lease_id=cbx_929b8e20f493
remote_command=CRABBOX_PR1203_LIVE_OK
exit_status=0
run_status=succeeded
lease_stopped=true
post_run_pod_count=0
```

The missing-sidecar preflight was also exercised with the disposable public-key file temporarily absent:

```text
exit_status=2
error=read SSH public key <redacted-path>.pub: no such file or directory
post_attempt_pod_count=0
```

The private key remained local. The evidence above omits the disposable key value, API credentials, host/IP, port, local paths, and provider endpoints. RunPod reported no remaining pods after either check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### Fixed
 
 - Kept ready-pool reconciliation rollout-compatible with older CLIs and coordinators, preserved unexpired in-flight claims across policy changes, and required actual ready capacity for `pool ensure` success.
+- Authorized RunPod SSH access with the configured public key while rejecting missing, empty, or invalid key files before creating a paid pod. Thanks @morluto.
 
 ## 0.40.1 - 2026-07-23
 
--- a/internal/cli/doctor.go
+++ b/internal/cli/doctor.go
@@ -341,7 +341,7 @@ func (a App) doctor(ctx context.Context, args []string) error {
 		if _, err := os.Stat(cfg.SSHKey); err != nil {
 			record("missing", "ssh-key", cfg.SSHKey, map[string]string{"path": cfg.SSHKey})
 			ok = false
-		} else if _, err := publicKeyFor(cfg.SSHKey); err != nil {
+		} else if _, err := PublicKeyFor(cfg.SSHKey); err != nil {
 			record("missing", "ssh-key", cfg.SSHKey+".pub", map[string]string{"path": cfg.SSHKey + ".pub"})
 			ok = false
 		} else {
--- a/internal/cli/lease.go
+++ b/internal/cli/lease.go
@@ -27,7 +27,10 @@ func newRunID() string {
 	return "run_" + strings.TrimPrefix(newLeaseID(), "cbx_")
 }
 
-func publicKeyFor(privatePath string) (string, error) {
+func PublicKeyFor(privatePath string) (string, error) {
+	if strings.TrimSpace(privatePath) == "" {
+		return "", exit(2, "ssh key path is not configured")
+	}
 	pub := privatePath + ".pub"
 	data, err := os.ReadFile(pub)
 	if err != nil {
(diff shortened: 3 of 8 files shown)


morluto:
@clawsweeper re-review

steipete:
Thanks, @morluto — I pushed a small maintainer follow-up on top of your commit.

I moved adjacent public-key loading and validation into the shared core helper, reused the existing OpenSSH type guard from KubeVirt, added a clear error for an unconfigured key path, and covered missing, empty, and private-key-content `.pub` files with zero-deploy assertions. This keeps your RunPod fix focused while ensuring bad key material is rejected before a paid pod is created.

Your redacted live RunPod proof—successful SSH, sync, and command execution plus zero pods after the missing-sidecar failure—was the deciding evidence for moving this forward. Thank you for doing that careful validation.

PR #1233, opened 2026-08-04
Add Run Cloud sandbox provider

t0tl:
## Summary

- add Run Cloud as a Crabbox provider, including registration, CLI/config wiring, doctor checks, metadata, and documentation
- provision persistent exposed sandboxes with the supported `runcloud/agent-base` image
- route SSH through `runcloud sandbox proxy` and support sync/run/inspect/list/stop lifecycle operations
- bootstrap compatibility prerequisites for older images, including the `runcloud` user, Git, and rsync
- make failed/interrupted live-smoke cleanup retain and destroy the exact requested sandbox

## Why

Run Cloud's retired `runcloud/box` image is no longer available. Crabbox therefore targets `runcloud/agent-base` explicitly and uses the authenticated Run Cloud SSH proxy rather than assuming a public SSH endpoint.

The live investigation also found a production image issue: the currently published agent-base guest lacks the guest-agent SSH WebSocket bridge on port 2222. The companion [Newly PR #3553](https://github.com/newly-app/newly/pull/3553) fixes that image and its SSH path.

## Validation

- `go test -race -count=1 ./internal/providers/runcloud ./internal/providers/all`
- `go build ./cmd/crabbox`
- `scripts/check-docs.sh`
- all 65 tests in `scripts/live-smoke.test.js`, including a retained-lease lifecycle proving warmup → status/inspect/ssh/cache/run → exactly one stop
- exact-head Run Cloud API/CLI live lifecycle against a real shared-dev sandbox: create, expose, exec, list, and destroy all passed in [Newly run 30933203610](https://github.com/newly-app/newly/actions/runs/30933203610)
- manual Crabbox live smoke reached real create/expose and then correctly waited for SSH readiness; it could not complete through the currently published guest image because port 2222 is absent; the exact sandbox was destroyed and no active Crabbox sandbox was left behind

## Remaining live boundary

Keep this draft until Newly #3553 is merged and its normally tested `runcloud/agent-base` artifact is published. Then rerun `scripts/live-smoke.sh --provider runcloud` to prove the Crabbox binary's complete SSH/sync/run path against the released image. The provider tests and Run Cloud lifecycle prerequisites are green; the integrated Crabbox-over-edge-SSH proof is intentionally not claimed yet.


steipete:
Thanks @t0tl for the substantial provider work and the live investigation.

Closing this draft rather than carrying the old integration forward. Its selected default image never completed the required Crabbox SSH/sync/run path, the external image-fix dependency is no longer publicly resolvable, and review also found that repository config could redirect the locally executed credentialed CLI to an arbitrary binary. Since then, Crabbox's provider, credential-provenance, lifecycle-claim, and live-proof contracts have moved substantially.

A future Run Cloud integration would be welcome as a fresh PR against current `main`, built around a current named API/CLI and released proxy-capable image, with the CLI path restricted to trusted configuration and a complete warmup → SSH → sync → run → cleanup proof.

PR #1235, opened 2026-08-05
fix: defer coordinator lease cleanup

fuller-stack-dev:
## Summary

Ordinary managed-provider release previously performed provider deletion inside the request, coupling client latency and request lifetime to provider I/O. This change makes the release transition durable first and leaves deletion, retry, and finalization to the coordinator alarm.

- Persist ordinary releases as `state=released` with `releaseDeletesServer=true` and an immediately expired `cleanupStartedAt` claim before returning.
- Keep the existing cleanup fields and claim timestamp so rollback readers discover the queued work and stale completions remain fenced.
- Preserve synchronous force-admin deletion: it reclaims an expired ordinary-release claim, while a genuinely active cleanup owner still receives `409 cleanup_in_progress`.
- Route JSON release, workspace and private-AWS release, portal release, and ready-pool drain/release through the same release operation.
- Keep `delete:false` non-destructive and repeated delete idempotent, with provider failures retained as visible retry state.

## Compatibility

This uses the existing lease cleanup fields and alarm reconciliation path. It requires no storage migration, protocol change, package-version change, or CLI workaround.

## Verification

- Worker formatting, lint, and TypeScript checks passed.
- Fleet tests: 551/551 passed.
- Private AWS workspace tests: 8/8 passed.
- Full Worker suite: 1,231 passed, 3 skipped.
- Worker dry-run build passed.
- Source-blind release contract: 8/8 clauses passed, backed by 559/559 lifecycle probes.
- `git diff --check`, exact-diff TruffleHog scan, and final structured P1 review passed with no actionable findings.

## Residual risk

No live provider was contacted in this repair pass. Real-provider deletion latency and eventual-consistency timing remain provider-dependent, but they now run behind the durable alarm retry path instead of the release request.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- Deferred ordinary coordinator lease provider cleanup to durable alarm-owned retries while preserving synchronous force-admin deletion and visible retry state. Thanks @fuller-stack-dev.
 - Made future Linux developer-image preparation use root-owned Corepack state while source, candidate, and promoted smoke checks exercise Corepack and pnpm as the runtime user. Thanks @fuller-stack-dev.
 - Made local sync report actionable non-Git workdir diagnostics and fail before lease acquisition, resolution, preparation, or ready-pool borrowing. Thanks @bunlongheng.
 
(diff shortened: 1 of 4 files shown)


fuller-stack-dev:
@clawsweeper re-review

fuller-stack-dev:
@clawsweeper re-review

PR #1236, opened 2026-08-06
fix(actions): full resync rehydrates adopted workspaces

vincentkoc:
## What Problem This Solves

An adopted GitHub Actions workspace could retain a readiness marker across
`--full-resync`. Crabbox would destructively replace the remote tree while the
old marker still described the workspace as hydrated, allowing a user command
to start against a newly recreated but unhydrated checkout.

## Root Cause and Fix

The full-resync path treated the hydration marker as authoritative through the
reset and did not guarantee that local hydration could reconstruct the
canonical lease workspace afterward. Crabbox now prepares the workflow inputs
and hydration script before remote mutation, invalidates only the readiness
marker, resets and syncs the workspace, clears stale hydration state, rehydrates
the canonical workspace, and only then executes the user command.

The prepared workflow snapshot is deliberately reused after reset. If the
workflow does not declare `crabbox_job`, the filtered plan clears the expected
job so execution does not require marker data the workflow could not receive.

## Compatibility and Ownership

The accepted compatibility contract is fail-closed: a command-bearing adopted
workspace run stops before reset when hydration is disabled, the workflow is
unsupported, or the adopted workdir is not the canonical local-hydration
workspace. `--sync-only` remains available to invalidate, reset, and sync an
adopted workdir without hydration or a user command.

The complete invalidation → reset → sync → clear → hydrate → command lifecycle
runs under the reusable-workspace owner lock introduced by
https://github.com/openclaw/crabbox/pull/1246. Marker invalidation uses the
existing idempotent SSH retry behavior. Target defaults are normalized before
command selection, so POSIX and WSL2 use the POSIX marker path while native
Windows uses the ProgramData PowerShell path.

## Verification

- Focused ordering, fail-closed, sync-only, prepared-snapshot, POSIX, WSL2,
  native-Windows, and owner-lock tests
- `go test -race ./internal/cli ./internal/providers/daytona ./internal/providers/blacksmith`
- `go test -p 1 ./...`
- `go vet ./...`
- `scripts/test-go-modules.sh`
- `scripts/check-go-coverage.sh 90.0` (`90.0%`)
- `scripts/check-docs.sh`
- Source-blind CLI validation of canonical rehydration, failure-before-mutation,
  repeated sync-only behavior, and WSL2 target handling
- Linux Testbox proof for the WSL2 normalization fixture
- TruffleHog and final P1 autoreview: clean

There is no configuration migration, secret migration, or credential handling
change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- Invalidated adopted Actions workspace readiness markers before full resync and rehydrated the canonical workspace before running commands. Thanks @vincentkoc.
 - Stopped sparse-checkout and skip-worktree omissions from deleting in-scope remote files, and kept staged gitlink removals out of file-deletion manifests. Thanks @vincentkoc.
 - Deferred ordinary coordinator lease provider cleanup to durable alarm-owned retries while preserving synchronous force-admin deletion and visible retry state. Thanks @fuller-stack-dev.
 - Made future Linux developer-image preparation use root-owned Corepack state while source, candidate, and promoted smoke checks exercise Corepack and pnpm as the runtime user. Thanks @fuller-stack-dev.
--- a/docs/commands/run.md
+++ b/docs/commands/run.md
@@ -239,6 +239,14 @@ Standalone `crabbox actions hydrate --id ...` acquires the same remote workspace
 owner as ordinary reused runs, so hydration cannot overlap a sync, command,
 collection, or cleanup from another client.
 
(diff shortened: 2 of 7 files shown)


vincentkoc:
@clawsweeper re-review

vincentkoc:
Maintainer disposition: accepted.

The documented fail-closed compatibility change is intentional. A
command-bearing adopted-workspace `--full-resync` must stop before destructive
reset when canonical rehydration is unavailable. Canonical automatic hydration
and `--sync-only` are the intended recovery paths. No further code change is
requested.

Issue #1231, opened 2026-08-03
Tracked files under dist/ target/ coverage/ etc. are silently dropped from sync (built-in excludes match every path component)


salmonumbrella:
## What happens

A file that is tracked in git is silently dropped from the sync manifest if any
directory on its path is named `dist`, `target`, `coverage`, `node_modules`,
`test-results`, `playwright-report`, `.build`, `.cache`, `.tmp`, `.local`,
`.venv`, `__pycache__`, or any other bare entry in the built-in
`sync.exclude` defaults — at any depth. Nothing warns; the file simply is not
there on the box.

Minimal repro, `crabbox 0.40.0`:

```console
$ git init -q . && printf 'hello\n' > src/main.txt \
    && mkdir -p pkg/dist target coverage \
    && printf 'tracked, not build output\n' | tee pkg/dist/keep.txt target/keep.txt coverage/keep.txt >/dev/null \
    && git add -A && git commit -qm init

$ git ls-files
coverage/keep.txt
pkg/dist/keep.txt
src/main.txt
target/keep.txt

$ crabbox sync-plan
sync candidate: 1 files, 6 B
top files:
  6 B        src/main.txt
```

Four tracked files, one synced. No warning, and `sync-plan` reports the reduced
number as if it were the whole repo.

## Why

The manifest starts from git and is correct: `repo.go:516` runs
`git ls-files --cached --others --exclude-standard -z`, which lists all four.

The files are dropped afterwards by `pathExcluded` (`repo.go:854`) →
`excludeMatches` (`repo.go:908`). For a pattern containing no `/`,
`excludeMatches` splits the path and compares against **every component**:

```go
if !strings.Contains(exclude, "/") {
    for _, part := range parts {
        if part == exclude {
            return true
        }
        ...
    }
}
```

So the built-in `dist` matches `pkg/dist/keep.txt`, `target` matches
`target/keep.txt`, and so on, at any depth. `crabbox config show --json`
confirms the defaults are bare, unanchored names:

```
".git", "._*", "node_modules", ".ignored", ".turbo", ".next", ".vite",
".parcel-cache", ".rollup.cache", "dist", "dist-runtime", "coverage",
"playwright-report", "test-results", ".cache", ".tmp", ".local", ".swiftpm",
".build", "apps/*/.build", ".pnpm-store", ".npm", ".yarn/cache", ".venv",
"__pycache__", ".pytest_cache", ".mypy_cache", ".ruff_cache", ".gradle",
[...]

PR #1241, opened 2026-08-06
fix(sync): stop sparse checkouts from deleting scoped files

vincentkoc:
## Summary

This repairs sparse and skip-worktree sync manifests so a partial local checkout cannot make an in-scope tracked file look like a remote deletion.

The root cause was that manifest construction started from paths visible in the working tree and index output. A tracked path omitted by sparse-checkout rules or skip-worktree state could therefore disappear from the upload without being distinguished from an intentional deletion. Staged submodule removal had a related type-loss problem: after `git rm --cached`, the stage-zero gitlink entry is gone even though the cached deletion still needs to be recognized as a gitlink rather than a remote file deletion.

## Behavior

- Fail closed before upload only for hidden tracked non-gitlink paths that survive `sync.include` and ordered excludes.
- Keep out-of-scope hidden paths ignored, preserve ordered re-inclusion, and leave the existing exclude semantics unchanged.
- Keep symlinks file-like, omit gitlinks from file uploads and deletions, and reject mixed file/gitlink unmerged modes in effective scope.
- Classify staged deletions from cached raw-diff preimage modes, so mode `160000` removals are not emitted as remote file deletions while ordinary staged file deletions remain deletions.
- Preserve the existing whole-checkout guard used by Blacksmith-style workflows.

Git 2.41 and newer uses `git sparse-checkout check-rules` to distinguish sparse omissions from intentional deletions. On older Git, an ambiguous missing path fails closed only when it remains in effective manifest scope; fully materialized paths and paths outside scope continue normally.

The actionable non-Git preflight remains the first Git-source check: synced runs outside a Git repository still stop with the existing `git init` / `--no-sync` guidance.

## Verification

- `gofmt` on the touched Go files
- focused sparse, skip-worktree, gitlink, mixed-mode, deletion, and delegated-upload regression tests
- focused race-detector run
- full `go test ./internal/cli -count=1`
- `go vet ./...`
- `scripts/check-docs.sh`
- `git diff --check`
- exact-bundle TruffleHog scan and maintainer review
- source-blind standalone CLI contract covering scope changes, staged gitlink deletion, mixed modes, symlinks, and non-Git diagnostic precedence

No configuration, secret, credential, data, or version migration is required.

Historical context: https://github.com/openclaw/crabbox/issues/1093

Non-goal: this does not change the exclude semantics discussed in https://github.com/openclaw/crabbox/issues/1231

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- Stopped sparse-checkout and skip-worktree omissions from deleting in-scope remote files, and kept staged gitlink removals out of file-deletion manifests. Thanks @vincentkoc.
 - Deferred ordinary coordinator lease provider cleanup to durable alarm-owned retries while preserving synchronous force-admin deletion and visible retry state. Thanks @fuller-stack-dev.
 - Made future Linux developer-image preparation use root-owned Corepack state while source, candidate, and promoted smoke checks exercise Corepack and pnpm as the runtime user. Thanks @fuller-stack-dev.
 - Made local sync report actionable non-Git workdir diagnostics and fail before lease acquisition, resolution, preparation, or ready-pool borrowing. Thanks @bunlongheng.
--- a/docs/commands/sync-plan.md
+++ b/docs/commands/sync-plan.md
@@ -28,6 +28,11 @@ matches what an actual sync would ship:
 Ordered exclude rules are applied before size accounting; a later `!pattern`
 can re-include a path matched by an earlier rule.
 
+The same preflight rejects tracked non-gitlink paths hidden by sparse-checkout
+or `skip-worktree` state only when they remain in the effective manifest after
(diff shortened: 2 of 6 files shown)


vincentkoc:
@clawsweeper review

Please review exact head `6e15331d0bacc61e19f2423235e74d9e0b56c880`.

vincentkoc:
@clawsweeper re-review

Please review exact head `c5ddf0be2066d0eaa0a30cd7a7ca1b1afe40dba9`.

vincentkoc:
@clawsweeper re-review

Please review exact head `5dfcd8bcaccc8d854474e66e00a912c9bcad955f`.

The documented Git-before-2.41 fail-closed behavior is the intended compatibility tradeoff: an ambiguous missing in-scope tracked path stops before upload rather than risking remote deletion from a partial manifest.

PR #1311, opened 2026-08-12
fix(cli): keep sensitive generated output private

### LINKED ISSUE #1240: [security] Windows attestation key can inherit cross-user read access from permissive AppData ACLs
## Summary

On Windows, Crabbox creates its automatically managed Ed25519 attestation key
under `%AppData%\crabbox\attest\id_ed25519.pem` using POSIX-style `0700` and
`0600` modes, but it does not create or verify a current-user-only Windows
discretionary access-control list (DACL).

Go documents that `os.Chmod` on Windows uses only the `0200` bit to control the
read-only attribute; it does not translate `0600` or `0700` into an owner-only
DACL. Microsoft documents that files and directories created without an
explicit security descriptor inherit applicable access-control entries (ACEs)
from their parent. As a result, when `%AppData%` is redirected, customized, or
otherwise below a permissive inheritable DACL, Crabbox can create the private
key with cross-user read access.

Another local OS user who can reach and read that path can copy the PKCS#8
Ed25519 private key and forge receipts that verify under the same public key and
fingerprint. The integrity impact is meaningful when that fingerprint has
previously been compared or trusted out of band.

This is **conditional and is not a default Windows-profile exposure**. A
normally configured per-user profile ACL prevents another standard user from
reading `%AppData%`. The issue is that Crabbox does not establish or validate
the promised per-user boundary for the credential it generates, so a valid
Windows deployment using a redirected or customized profile can silently
weaken that boundary.

The issue is confirmed at revision
[`0171ae2ee19fbf18ef0c2d5a2d04d8fe56a437f2`](https://github.com/openclaw/crabbox/commit/0171ae2ee19fbf18ef0c2d5a2d04d8fe56a437f2).
The introduction point and first fixed revision, if any, have not been bounded.

## Security boundary and prerequisites

Crabbox's security policy lists "generated credentials being exposed beyond the
current OS user by Crabbox" as an in-scope boundary. The affected file is a
Crabbox-generated signing credential, not an arbitrary operator-selected output
file.

All of the following conditions are required for the persistent key-disclosure
and trusted-fingerprint forgery path:

1. Crabbox runs on Windows.
2. The operator uses `crabbox run --attest ...` without supplying a separately
   secured `--attest-key`, causing Crabbox to mint or reuse its managed key.
3. The managed key is created beneath an AppData ancestry with an inheritable
   ACE that grants the attacker, or a group containing the attacker, the
   necessary directory traversal and file-read access.
4. The ACE propagates through the `crabbox` and `attest` directories and onto
   the temporary key file; no protected child DACL, effective deny ACE, EFS
   boundary, or equivalent control blocks access.
5. The selected filesystem supports Crabbox's same-directory hard-link
   publication path. If the link operation fails, the persistent key is not
   successfully published through this path.
6. For identity-relevant receipt forgery, a relying party has previously
   trusted or compared the victim key's fingerprint out of band. Unpinned
   receipts remain explicitly self-signed.

The primary countercontrol is the standard per-user Windows profile ACL. Under
normal profile permissions, another ordinary user cannot simply traverse and
read the key. Windows nevertheless supports profile-folder redirection, and
administrators can customize ACLs, so a permissive inheritable ancestry is a
real deployment condition rather than an impossible state.

### 1. Crabbox derives the managed key path from `%AppData%`

[`internal/cli/attest.go:42-79`](https://github.com/openclaw/crabbox/blob/0171ae2ee19fbf18ef0c2d5a2d04d8fe56a437f2/internal/cli/attest.go#L42-L79)
calls `os.UserConfigDir`, appends `crabbox/attest/id_ed25519.pem`, generates an
Ed25519 key, marshals it as PKCS#8 PEM, and passes it to the private-output
helper:

```go
func attestKeyPath() (string, error) {
	base, err := os.UserConfigDir()
	if err != nil {
		return "", err
	}
	return filepath.Join(base, "crabbox", "attest", "id_ed25519.pem"), nil
}

// ...
_, key, err := ed25519.GenerateKey(rand.Reader)
der, err := x509.MarshalPKCS8PrivateKey(key)
encoded := pem.EncodeToMemory(&pem.Block{Type: "PRIVATE KEY", Bytes: der})
created, err := writePrivateRunOutputFileIfAbsent(path, encoded)
```

Go documents that `UserConfigDir` is `%AppData%` on Windows. Crabbox's
documented managed-key location therefore resolves to:

```text
%AppData%\crabbox\attest\id_ed25519.pem
```

### 2. Directory creation and `Chmod` do not establish a Windows DACL

The platform-neutral helper creates the directory with `0700` at
[`internal/cli/run_local_output.go:11-18`](https://github.com/openclaw/crabbox/blob/0171ae2ee19fbf18ef0c2d5a2d04d8fe56a437f2/internal/cli/run_local_output.go#L11-L18).
The Windows-specific wrapper then calls `os.Chmod(path, 0700)` at
[`internal/cli/run_local_output_windows.go:11-16`](https://github.com/openclaw/crabbox/blob/0171ae2ee19fbf18ef0c2d5a2d04d8fe56a437f2/internal/cli/run_local_output_windows.go#L11-L16).
There is no security descriptor construction, DACL protection, or ACL readback
in this path.

The official [`os.Chmod` documentation](https://pkg.go.dev/os#Chmod) is
explicit: on Windows, only the `0200` owner-writable bit is used, to control the
file's read-only attribute, and the other mode bits are unused. Thus:

- `0700` makes the directory writable; it does not mean "current user only";
- `0600` makes the file writable; it does not mean "current user only";
- neither operation removes inherited allow ACEs or marks the DACL protected.

The current official Go Windows implementation calls `CreateDirectory` without
a supplied security descriptor for ordinary directory creation; see
[`syscall.Mkdir`](https://go.dev/src/syscall/syscall_windows.go#L623).

Microsoft's [File Security and Access Rights](https://learn.microsoft.com/en-us/windows/win32/fileio/file-security-and-access-rights)
documentation states that, when a file or directory is created without an
explicit security descriptor, its default ACL is inherited from the parent.
Microsoft's [ACE inheritance rules](https://learn.microsoft.com/en-us/windows/win32/secauthz/ace-inheritance-rules)
explain how object- and container-inheritable ACEs propagate to files and
directories. A protected DACL is the relevant way to prevent inherited ACEs
from modifying the child DACL; see
[`SE_DACL_PROTECTED`](https://learn.microsoft.com/en-us/windows/win32/secauthz/security-descriptor-control)
and [Automatic Propagation of Inheritable ACEs](https://learn.microsoft.com/en-us/windows/win32/secauthz/automatic-propagation-of-inheritable-aces).

### 3. The temporary key file is born with inherited security

[`internal/cli/run_local_output.go:20-54`](https://github.com/openclaw/crabbox/blob/0171ae2ee19fbf18ef0c2d5a2d04d8fe56a437f2/internal/cli/run_local_output.go#L20-L54)
performs this sequence:

```go
file, err := os.CreateTemp(dir, "."+filepath.Base(path)+".crabbox-*")
// ...
if err := file.Chmod(0o600); err != nil { /* ... */ }
if _, err := io.Copy(file, bytes.NewReader(data)); err != nil { /* ... */ }
// ...
if err := os.Link(tempPath, path); err != nil { /* ... */ }
```

Go's [`os.CreateTemp` implementation](https://go.dev/src/os/tempfile.go#L26)
creates the temporary file through `OpenFile`. The Windows open path supplies no
application-defined security descriptor; see the official
[`syscall.Open` implementation](https://go.dev/src/syscall/syscall_windows.go#L364).

We therefore reach the decisive state before the private bytes are written: the
temporary file has been created without a Crabbox-supplied protected DACL, and
`Chmod(0600)` has not changed its inherited DACL. If the inherited ACE grants
the other local user effective read access, writing the PKCS#8 bytes places the
complete signing key in a file that user can read.

### 4. Hard-link publication preserves the file's security descriptor

The final key name is created with `os.Link(tempPath, path)`, after which the
temporary pathname is removed. Microsoft documents for
[`CreateHardLinkW`](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createhardlinkw)
that security information belongs to the underlying file and that creating a
hard link does not modify the file's security descriptor. The final
`id_ed25519.pem` name therefore refers to the same file with the same inherited
DACL; publication does not repair it.

If a key already exists, [`internal/cli/attest.go:164-187`](https://github.com/openclaw/crabbox/blob/0171ae2ee19fbf18ef0c2d5a2d04d8fe56a437f2/internal/cli/attest.go#L164-L187)
reads and parses it without checking its owner, DACL, protected-DACL control
bit, inherited ACEs, or effective access. Reuse therefore does not detect or
repair an already exposed managed key.

### 5. Key disclosure permits same-fingerprint receipt forgery

Receipt generation derives the public key from the private key, embeds it in
the receipt, and signs the canonicalized receipt at
[`internal/cli/attest.go:427-475`](https://github.com/openclaw/crabbox/blob/0171ae2ee19fbf18ef0c2d5a2d04d8fe56a437f2/internal/cli/attest.go#L427-L475).
Verification reconstructs those bytes, computes the public-key fingerprint,
and calls `ed25519.Verify` at
[`internal/cli/attest.go:478-524`](https://github.com/openclaw/crabbox/blob/0171ae2ee19fbf18ef0c2d5a2d04d8fe56a437f2/internal/cli/attest.go#L478-L524).

Once we carry the disclosed private key into the same signing operation, we can
produce arbitrary receipt fields plus a valid signature under the victim's
public key. The verifier reports the same signer fingerprint because that
fingerprint is derived from the same key pair. No cryptographic break is
needed.

Crabbox correctly labels verification as `trust=self-signed`; it does not claim
that an unpinned key proves identity or execution provenance. The supported
integrity consequence is narrower: theft defeats any identity or continuity a
relying party previously attached to that fingerprint through an out-of-band
comparison.

### 6. Crabbox already contains the required Windows security pattern

This is not a limitation of the available Windows APIs or dependencies.
[`internal/cli/ssh_transport_permissions_windows.go:11-72`](https://github.com/openclaw/crabbox/blob/0171ae2ee19fbf18ef0c2d5a2d04d8fe56a437f2/internal/cli/ssh_transport_permissions_windows.go#L11-L72)
gets the current user's SID, installs a DACL with
`PROTECTED_DACL_SECURITY_INFORMATION`, then reads back and validates the owner
and DACL.

Similarly,
[`internal/cli/webvnc_bootstrap_file_windows.go:17-122`](https://github.com/openclaw/crabbox/blob/0171ae2ee19fbf18ef0c2d5a2d04d8fe56a437f2/internal/cli/webvnc_bootstrap_file_windows.go#L17-L122)
creates a credential-bearing directory and file with an explicit
`D:P(A;;GA;;;<current-user-SID>)` security descriptor and verifies
`SE_DACL_PROTECTED`. The attestation path does neither.

## Impact

Successful exploitation gives another local OS user:

- **complete disclosure of the managed Ed25519 private signing key**; and
- **the ability to forge receipts under the same public key and fingerprint**,
  defeating any prior out-of-band trust in that signer fingerprint.

It does **not** by itself grant a Crabbox coordinator token, provider
credential, remote shell, or proof that a claimed run actually occurred.
Availability is unaffected. The broader inherited-ACL behavior may also affect
ordinary run-output files written through the shared helper, but that is not
included in this security impact or score.

### CVSS Assessment

| Metric | v3.1 | v4.0 |
|--------|------|------|
| **Score** | 6.3 / 10.0 | 7.2 / 10.0 |
| **Severity** | Medium | High |
| **Vector** | CVSS:3.1/AV:L/AC:H/PR:L/UI:N/S:U/C:H/I:H/A:N | CVSS:4.0/AV:L/AC:L/AT:P/PR:L/UI:N/VC:H/VI:H/VA:N/SC:N/SI:N/SA:N |
| **Calculator** | [CVSS v3.1 Calculator](https://www.first.org/cvss/calculator/3.1#CVSS:3.1/AV:L/AC:H/PR:L/UI:N/S:U/C:H/I:H/A:N) | [CVSS v4.0 Calculator](https://www.first.org/cvss/calculator/4.0#CVSS:4.0/AV:L/AC:L/AT:P/PR:L/UI:N/VC:H/VI:H/VA:N/SC:N/SI:N/SA:N) |

The attacker requires a low-privileged identity on the same host. CVSS v3.1
represents the non-default permissive inherited ACL as `AC:H`; CVSS v4.0
represents it more directly as `AT:P`. Once that deployment condition and the
generated key exist, reading the key requires no mitigation bypass or
specialized exploit, so v4.0 remains `AC:L`.

`UI:N` is appropriate in both versions. The operator's earlier legitimate use
of `--attest` creates the vulnerable state, but the operator does not interact
with the attacker or an attacker-supplied payload during the later key read or
forgery. Confidentiality and integrity are High because the complete private
key is disclosed and the protection supplied by that signing identity is fully
defeated. Availability and subsequent-system impacts are None.

The v4.0 High result does not mean default Windows installations are exposed:
`AT:P` records the required vulnerable ACL condition. See the
[CVSS v3.1 specification](https://www.first.org/cvss/v3.1/specification-document),
[CVSS v4.0 specification](https://www.first.org/cvss/v4.0/specification-document),
and [CVSS v4.0 FAQ](https://www.first.org/cvss/v4.0/faq).

## Validation basis and limitation

I reviewed the exact affected revision and traced the managed key from path
selection through directory creation, temporary-file creation, hard-link
publication, loading, signing, and verification. I also ran the focused
attestation tests successfully:

```text
$ go test ./internal/cli -run 'Test(WriteRunReceipt|Verify|EnsureAttest|Attest)' -count=1
ok github.com/openclaw/crabbox/internal/cli
```

Those tests validate key creation/reuse and the signing/verification consequence
on the available macOS host. I did not have a usable Windows runtime, so I did
not perform an empirical `Get-Acl`, `icacls`, or effective-access readback and
do not claim a Windows runtime PoC. The Windows permission conclusion follows
from the exact creation path plus the official Go and Microsoft semantics cited
above.

The current test suite itself exposes the platform gap:
[`internal/cli/attest_test.go:514-547`](https://github.com/openclaw/crabbox/blob/0171ae2ee19fbf18ef0c2d5a2d04d8fe56a437f2/internal/cli/attest_test.go#L514-L547)
checks `0600`/`0700` only when `runtime.GOOS != "windows"`; there is no Windows
DACL assertion for the managed attestation key.

## Recommended remediation

The security invariant should be: **before any private-key byte is written, the
key directory and file must have a protected DACL owned by the current user and
granting access only to that user; every existing managed key must satisfy the
same invariant before use.**

An implementation-ready approach is:

1. Extract a shared current-user-only Windows security-descriptor builder and
   validator from the existing WebVNC/SSH implementations. Its descriptor can
   retain the established shape `O:<SID>D:P(A;;GA;;;<SID>)`.
2. Create every missing managed-key directory component with
   `windows.CreateDirectory` and those `SecurityAttributes`. Validate existing
   components before proceeding instead of relying on `MkdirAll(..., 0700)`.
3. Create the same-directory temporary key file with `windows.CreateFile`,
   `CREATE_NEW`, and the protected security attributes. Do this at creation
   time—applying a restrictive DACL only after `CreateTemp` would leave a
   creation-to-repair exposure window.
4. Read back the owner and DACL before writing the PKCS#8 bytes. Require the
   current user as owner, require `SE_DACL_PROTECTED`, and reject unexpected
   allow trustees or rights.
5. Preserve the current atomic hard-link publication behavior, then validate
   the final path as defense in depth. A hard link should retain the already
   protected descriptor.
6. Before loading an existing automatically managed key, validate the same
   invariant. Fail closed with an actionable error, or repair the DACL safely
   before reading/signing. If exposure is suspected, direct the operator to
   rotate the key and replace any trusted fingerprint.
7. Keep this credential-specific boundary distinct from intentionally shared,
   operator-selected output paths; do not silently make all output locations
   current-user-only if sharing is a supported use case.

## Regression tests

Add Windows-native tests that create a disposable parent directory with an
inheritable read ACE for a second test SID, then exercise the real managed-key
creation path and assert:

- the created `crabbox` and `attest` directories are owned by the current user
  and have protected DACLs;
- the temporary and final key file are born with, and retain, a protected
  current-user-only DACL;
- the second SID has no effective read access to the final key;
- private bytes are never written before ACL validation succeeds;
- an existing key with a permissive or unprotected DACL is rejected or safely
  repaired before signing;
- concurrent first-use key minting still converges on one key;
- hard-link publication preserves the protected descriptor; and
- normal key creation, override-key use, receipt signing, and verification
  continue to work.

The Windows test should inspect the actual owner, DACL, control flags, inherited
ACEs, and effective access. It should not substitute POSIX mode-bit assertions
for Windows authorization checks.

## References

- Go: [`os.Chmod`](https://pkg.go.dev/os#Chmod)
- Go: [`os.UserConfigDir`](https://pkg.go.dev/os#UserConfigDir)
- Go source: [`os.CreateTemp`](https://go.dev/src/os/tempfile.go#L26)
- Go source: Windows [`syscall.Open`](https://go.dev/src/syscall/syscall_windows.go#L364) and [`syscall.Mkdir`](https://go.dev/src/syscall/syscall_windows.go#L623)
- Microsoft: [File Security and Access Rights](https://learn.microsoft.com/en-us/windows/win32/fileio/file-security-and-access-rights)
- Microsoft: [ACE Inheritance Rules](https://learn.microsoft.com/en-us/windows/win32/secauthz/ace-inheritance-rules)
- Microsoft: [Automatic Propagation of Inheritable ACEs](https://learn.microsoft.com/en-us/windows/win32/secauthz/automatic-propagation-of-inheritable-aces)
- Microsoft: [`SECURITY_DESCRIPTOR_CONTROL`](https://learn.microsoft.com/en-us/windows/win32/secauthz/security-descriptor-control)
- Microsoft: [`CreateHardLinkW`](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createhardlinkw)
- Microsoft: [About User Profiles](https://learn.microsoft.com/en-us/windows/win32/shell/about-user-profiles)
- Microsoft: [Folder Redirection and Roaming User Profiles](https://learn.microsoft.com/en-us/windows-server/storage/folder-redirection/folder-redirection-rup-overview)

## Conclusion

Crabbox correctly intends the automatically generated attestation key to be a
per-user secret, but on Windows it currently relies on POSIX mode values that do
not implement that boundary. Under a permissive inheritable AppData DACL, the
temporary key file and its final hard link can retain cross-user read access,
allowing disclosure of the private key and forgery under an already trusted
signer fingerprint.

Default Windows profile ACLs are an important countercontrol, so the report does
not claim universal or default exploitability. Creating and validating an
explicit protected current-user-only DACL at file-creation time restores the
intended invariant without changing the attestation format or trust model.

### LINKED ISSUE #1304: Artifact manifest carrying presigned URLs is written world-readable
## Problem

`artifact-manifest.json` embeds each file's `URL`. Under the default
`signed-url` access policy that URL is a presigned S3/R2 link — a time-limited
bearer capability. The manifest code knows this: `artifactAccessPolicy`
(`internal/cli/artifacts_manifest.go`) classifies a URL as `signed-url`
specifically by detecting `X-Amz-Signature` / `X-Amz-Credential` in its query.

It is nevertheless written with mode `0o644`:

```go
if err := writeArtifactBundleFile(root, artifactManifestFilename, data, 0o644); err != nil {
```

The repository already has the right primitive — `ensurePrivateRunOutputDir` /
`openPrivateRunOutputFile` (`internal/cli/run_local_output_unix.go`) apply
`0o700` / `0o600` — and `ensureArtifactBundleDir` uses it for the **default**
bundle directory. But when the user passes an explicit `--output`, it falls back
to `os.MkdirAll(path, 0o755)`, leaving a `0644` manifest readable by every local
user.

There is a second path to the same outcome. `writeArtifactBundleFile` preserves
an existing file's mode on rewrite:

```go
if info.Mode().IsRegular() {
    createPerm = 0o600
    existingMode = info.Mode().Perm()
    preserveExistingMode = true
}
```

so republishing over a manifest that is already `0644` keeps it `0644`, even into
a directory that is otherwise private.

## Impact

On a shared build host or multi-tenant CI runner, any local user can read the
presigned URLs and use them until expiry — `defaultURLExpiresSeconds` is 7 days.

Severity is capped by requiring local filesystem access and by URL expiry, and
one can argue an explicit `--output` signals intent to share the *artifacts*. But
the manifest is metadata, not a published artifact, and it is the one file in the
bundle that carries credentials. The default path is already private, which shows
the intent; `--output` should not silently drop it.

## Expected behaviour

The manifest should be owner-only regardless of the directory mode, and a
republish should never inherit a wider mode from an earlier bundle.

More generally, the `perm` argument to `writeArtifactBundleFile` expresses the
maximum permission the caller intends for that content. Preserving a user's
narrower mode is reasonable; widening past what the caller asked for is not.

## Sibling instance

`published-artifacts.md` has the same problem. `artifactTemplateMarkdown` embeds
each file's URL (`internal/cli/artifacts_publish.go:896`,
`location := firstNonBlank(file.URL, file.Path)`), so under the `signed-url`
policy it carries the same presigned capabilities as the manifest — and it is
written `0o644` through the same helper (`internal/cli/artifacts.go:685`).

Both should be fixed together; they are one bug, not two.

## Scope note

The remaining `0o644` writers in `internal/cli` were checked and carry no bearer
credential: doctor output, the `writeJSONFile` payloads (`metadata.json`, desktop
proof metadata, WebVNC status), the temp-dir PowerShell capture script, and the
generated config template.

One case is deliberately left as a follow-up: `artifacts.go:472`
(`artifacts template --output`) can embed user-supplied before/after URLs, which
may be presigned. The user names that output path explicitly and it sits outside
the bundle writer, so it is a different owner boundary.

dwin-gharibi:
Closes https://github.com/openclaw/crabbox/issues/1304.
Closes https://github.com/openclaw/crabbox/issues/1240.

## Summary

Sensitive generated files were not governed by one cross-platform privacy boundary. Artifact manifests and publication summaries could contain signed URLs, run captures could contain command output, and managed attestation keys are credentials, but individual writers relied on POSIX mode literals or inherited filesystem behavior. That left wider existing modes intact, did not remove inherited macOS ACLs, and did not provide an equivalent protected-DACL contract on Windows.

This change introduces one shared private-output path for those generated files:

- POSIX files are created or replaced as `0600` and verified after writing.
- macOS additionally removes inherited extended ACLs before accepting the file as private.
- Windows creates files with a protected, current-user-owned DACL, validates owner and access entries through the file handle, rejects reparse points, and uses handle-relative creation beneath `os.Root` where required.
- Existing Crabbox-managed attestation keys are tightened before use. User-supplied `--attest-key` files are never modified.
- Artifact manifests and Markdown summaries use the private writer only when they contain signed-URL bearer capabilities. Public/local artifact output retains its intended shareability.
- Republishing sensitive output cannot restore a broad POSIX mode or Windows DACL.

The shared primitive also replaces duplicated per-feature permission logic, so future sensitive generated output has one ownership boundary rather than another platform-specific mode check.

## Verification

Local and cross-platform proof:

- focused `internal/cli` regression tests passed
- `go vet ./...` passed
- Windows and Linux test cross-builds passed
- focused affected race suites passed before the final rebase
- structured branch security review passed with no actionable findings after explicitly adding `FILE_READ_ATTRIBUTES` to handles that are subsequently validated

Native Windows proof ran these exact tests successfully:

- `TestPrivateRunOutputWindowsDoesNotInheritPermissiveDACL`
- `TestManagedAttestKeyWindowsRepairsPermissiveDACL`
- `TestArtifactOutputWindowsPrivacyFollowsSignedURLs`

The native proof confirmed private temporary creation, repair of permissive managed-key DACLs, private signed-URL manifests/summaries, and shareable public manifests/summaries. Crabbox workspace-owner acquisition on the clean Windows lease twice stopped earlier with `ambiguous remote state`; to isolate that unrelated wrapper state, the already cross-built focused test executable was uploaded and executed through the owned direct SSH transport. It passed, the executable was removed, and the lease was released.

## Compatibility

Public artifact bundles remain public. The privacy change is conditional on sensitive content or an explicitly private output surface, and existing user-supplied attestation-key permissions remain under user control.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - Reported explicit stdout and stderr capture paths and byte counts in emitted run proofs without reading or embedding captured content. Thanks @coygeek.
 - Routed implicit `status` and `inspect` lease identifiers through the provider recorded in local claims before initializing the configured provider, while preserving explicit-provider precedence and missing-claim fallback. Thanks @coygeek.
 - Made POSIX workspace ownership independent of the remote account's login shell by transporting owner scripts through a private `/bin/sh` launcher, fixing static macOS sync under zsh, Bash, and Fish. Thanks @osouthgate and @hosmelq.
+- Restricted sensitive generated local files, including managed attestation keys and signed-URL artifact outputs, to the current OS user on POSIX and Windows. Thanks @dwin-gharibi.
 
 ## 0.41.5 - 2026-08-12
 
--- a/docs/commands/artifacts.md
+++ b/docs/commands/artifacts.md
@@ -289,6 +289,15 @@ published file:
 }
 ```
 
+When a generated manifest contains a signed or presigned bearer URL, Crabbox
+keeps the local manifest owner-only: mode `0600` on POSIX, with inherited
+extended ACLs removed on macOS, and a protected DACL owned by and limited to
+the current user on Windows. The generated
+`published-artifacts.md` summary uses the same protection only when it embeds a
+signed URL. Republishing tightens an older broad file instead of preserving its
+permissions. Manifests and summaries containing only local paths or public URLs
+keep their existing shared-output `0644` and mode-preservation behavior.
+
 Inspect or fetch that manifest later:
 
 ```sh
--- a/docs/security.md
+++ b/docs/security.md
@@ -429,7 +429,16 @@ S3, R2, or Cloudflare uploads, so later bundle path replacement cannot change
 uploaded bytes. Local and dry-run manifests hash through rooted validated file
 handles without duplicating the bundle. Generated manifest and Markdown files
 replace reserved outputs through root-confined temporary files without
-following symlinks. Required artifact paths must resolve to regular files.
+following symlinks. Manifests and summaries that contain signed bearer URLs are
+restricted to the current OS user before their sensitive bytes are written and
+are verified before publication: mode `0600` on POSIX, with inherited macOS
+ACLs removed, and a protected, current-user-owned DACL with no unrelated grants
+on Windows. Public-URL and local-path artifact outputs retain their shareable
+permissions. The same local
+private-file boundary protects Crabbox-generated run outputs and the managed
+attestation signing key; replacement or reuse tightens an older broad mode or
+DACL instead of preserving it. Required artifact paths must resolve to regular
+files.
 Automatic remote failure bundles confine member names and link targets to their
 generated subtree and omit
 escaping, rooted, empty, or special-file entries. These filesystem checks do
--- a/internal/cli/artifacts.go
+++ b/internal/cli/artifacts.go
@@ -682,9 +682,14 @@ func (a App) publishArtifactDirectory(ctx context.Context, opts artifactPublishO
 	}
 	body := artifactTemplateMarkdown(opts.Template, summary, "", "", published)
 	bodyPath := filepath.Join(opts.Directory, "published-artifacts.md")
-	// Same reason as the manifest: artifactTemplateMarkdown embeds each file's URL, which
-	// under the signed-url access policy is a presigned bearer capability.
-	if err := writeArtifactBundleFile(bundleRoot, "published-artifacts.md", []byte(body), privateRunOutputFileMode); err != nil {
+	// artifactTemplateMarkdown embeds each file's URL. Restrict the summary only when one of
+	// those URLs is a presigned bearer capability; a public or local summary is meant to be
+	// shared, so it keeps the readable mode it has always had.
+	bodyMode := os.FileMode(0o644)
+	if artifactFilesContainSignedURL(published) {
+		bodyMode = privateRunOutputFileMode
+	}
+	if err := writeArtifactBundleFile(bundleRoot, "published-artifacts.md", []byte(body), bodyMode); err != nil {
 		return nil, "", "", err
 	}
 	if opts.PR > 0 && !opts.NoComment {
--- a/internal/cli/artifacts.go
+++ b/internal/cli/artifacts.go
@@ -682,7 +682,9 @@ func (a App) publishArtifactDirectory(ctx context.Context, opts artifactPublishO
 	}
 	body := artifactTemplateMarkdown(opts.Template, summary, "", "", published)
 	bodyPath := filepath.Join(opts.Directory, "published-artifacts.md")
-	if err := writeArtifactBundleFile(bundleRoot, "published-artifacts.md", []byte(body), 0o644); err != nil {
+	// Same reason as the manifest: artifactTemplateMarkdown embeds each file's URL, which
+	// under the signed-url access policy is a presigned bearer capability.
+	if err := writeArtifactBundleFile(bundleRoot, "published-artifacts.md", []byte(body), privateRunOutputFileMode); err != nil {
 		return nil, "", "", err
 	}
 	if opts.PR > 0 && !opts.NoComment {
--- a/internal/cli/artifacts.go
+++ b/internal/cli/artifacts.go
@@ -685,11 +685,12 @@ func (a App) publishArtifactDirectory(ctx context.Context, opts artifactPublishO
 	// artifactTemplateMarkdown embeds each file's URL. Restrict the summary only when one of
 	// those URLs is a presigned bearer capability; a public or local summary is meant to be
 	// shared, so it keeps the readable mode it has always had.
-	bodyMode := os.FileMode(0o644)
 	if artifactFilesContainSignedURL(published) {
-		bodyMode = privateRunOutputFileMode
+		err = writePrivateArtifactBundleFile(bundleRoot, "published-artifacts.md", []byte(body))
+	} else {
+		err = writeArtifactBundleFile(bundleRoot, "published-artifacts.md", []byte(body), 0o644)
 	}
-	if err := writeArtifactBundleFile(bundleRoot, "published-artifacts.md", []byte(body), bodyMode); err != nil {
+	if err != nil {
 		return nil, "", "", err
 	}
 	if opts.PR > 0 && !opts.NoComment {
--- a/internal/cli/artifacts_manifest.go
+++ b/internal/cli/artifacts_manifest.go
@@ -137,11 +137,21 @@ func writeArtifactManifest(root *os.Root, opts artifactPublishOptions, files []a
 		return "", nil, exit(2, "encode artifact manifest: %v", err)
 	}
 	data = append(data, '\n')
-	// The manifest embeds each file's URL, which for the signed-url access policy is a
-	// presigned bearer capability. Keep it owner-only even when --output selects a
-	// directory that is not itself private.
-	if err := writeArtifactBundleFile(root, artifactManifestFilename, data, privateRunOutputFileMode); err != nil {
-		return "", nil, err
+	private := false
+	for _, file := range manifest.Files {
+		if file.AccessPolicy == "signed-url" {
+			private = true
+			break
+		}
+	}
+	var writeErr error
+	if private {
+		writeErr = writePrivateArtifactBundleFile(root, artifactManifestFilename, data)
+	} else {
+		writeErr = writeArtifactBundleFile(root, artifactManifestFilename, data, 0o644)
+	}
+	if writeErr != nil {
+		return "", nil, writeErr
 	}
 	return path, data, nil
 }
--- a/internal/cli/artifacts_manifest.go
+++ b/internal/cli/artifacts_manifest.go
@@ -137,7 +137,10 @@ func writeArtifactManifest(root *os.Root, opts artifactPublishOptions, files []a
 		return "", nil, exit(2, "encode artifact manifest: %v", err)
 	}
 	data = append(data, '\n')
-	if err := writeArtifactBundleFile(root, artifactManifestFilename, data, 0o644); err != nil {
+	// The manifest embeds each file's URL, which for the signed-url access policy is a
+	// presigned bearer capability. Keep it owner-only even when --output selects a
+	// directory that is not itself private.
+	if err := writeArtifactBundleFile(root, artifactManifestFilename, data, privateRunOutputFileMode); err != nil {
 		return "", nil, err
 	}
 	return path, data, nil
--- a/internal/cli/artifacts_manifest.go
+++ b/internal/cli/artifacts_manifest.go
@@ -491,16 +491,37 @@ func rejectSymlinkedArtifactOutputParents(root, name string) error {
 	return nil
 }
 
+// artifactURLIsSigned reports whether a URL carries a presigned bearer capability. It is the
+// single definition of that boundary, shared by the access policy and by output file modes.
+func artifactURLIsSigned(rawURL string) bool {
+	if strings.TrimSpace(rawURL) == "" {
+		return false
+	}
+	parsed, err := url.Parse(rawURL)
+	if err != nil {
+		return false
+	}
+	query := parsed.Query()
+	return query.Get("X-Amz-Signature") != "" || query.Get("X-Amz-Credential") != ""
+}
+
+// artifactFilesContainSignedURL reports whether any published file exposes a presigned URL,
+// so generated output that embeds those URLs can be written owner-only.
+func artifactFilesContainSignedURL(files []artifactFile) bool {
+	for _, file := range files {
+		if artifactURLIsSigned(file.URL) {
+			return true
+		}
+	}
+	return false
+}
+
 func artifactAccessPolicy(rawURL, storage string) string {
 	if strings.TrimSpace(rawURL) == "" {
 		return "local"
 	}
-	parsed, err := url.Parse(rawURL)
-	if err == nil {
-		query := parsed.Query()
-		if query.Get("X-Amz-Signature") != "" || query.Get("X-Amz-Credential") != "" {
-			return "signed-url"
-		}
+	if artifactURLIsSigned(rawURL) {
+		return "signed-url"
 	}
 	if strings.HasPrefix(strings.ToLower(strings.TrimSpace(rawURL)), "http://") || strings.HasPrefix(strings.ToLower(strings.TrimSpace(rawURL)), "https://") {
 		return "public"
--- a/internal/cli/artifacts_private_other.go
+++ b/internal/cli/artifacts_private_other.go
@@ -1,0 +1,21 @@
+//go:build !windows
+
+package cli
+
+import (
+	"os"
+)
+
+func openArtifactBundleTemp(root *os.Root, name string, perm os.FileMode, private bool) (*os.File, error) {
+	file, err := root.OpenFile(name, os.O_WRONLY|os.O_CREATE|os.O_EXCL, perm)
+	if err != nil {
+		return nil, err
+	}
+	if private {
+		if err := securePrivateFile(file); err != nil {
+			_ = file.Close()
+			return nil, err
+		}
+	}
+	return file, nil
+}
--- a/internal/cli/artifacts_private_windows.go
+++ b/internal/cli/artifacts_private_windows.go
@@ -1,0 +1,21 @@
+//go:build windows
+
+package cli
+
+import (
+	"os"
+
+	"golang.org/x/sys/windows"
+)
+
+func openArtifactBundleTemp(root *os.Root, name string, perm os.FileMode, private bool) (*os.File, error) {
+	if !private {
+		return root.OpenFile(name, os.O_WRONLY|os.O_CREATE|os.O_EXCL, perm)
+	}
+	directory, err := root.Open(".")
+	if err != nil {
+		return nil, err
+	}
+	defer directory.Close()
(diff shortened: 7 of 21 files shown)


steipete (approved):
Maintainer review complete: root-cause fix, cross-platform permission regressions, native Windows proof, exact-head CI, and structured P1 security review are clean.

steipete:
Maintainer decision:

- **Fail closed is intentional.** Managed attestation keys and generated files containing command output or signed-URL bearer capabilities must not be used when Crabbox cannot establish or verify the documented current-user-only Windows DACL. User-supplied `--attest-key` files remain outside this repair path and are not modified.
- **The completed native Windows proof is sufficient for this change.** The focused executable ran natively on a clean Windows lease and passed private inherited-DACL creation, permissive managed-key DACL repair, private signed-URL manifest/summary output, and the public-output shareability control. Two ordinary `crabbox run` attempts stopped earlier at workspace-owner acquisition with `ambiguous remote state`; that was isolated as an unrelated wrapper boundary, so the same reviewed executable was uploaded and run through the owned SSH transport, then removed and the lease released.

The exact head also has green CI and connector/security checks, zero unresolved review threads, and a clean structured P1 security review. I am accepting that evidence and waiving an additional CLI transcript rather than conflating the unrelated workspace-owner acquisition defect with the Windows DACL behavior under review.

Issue #1309, opened 2026-08-12
GitHub team allowlist is not scoped to the org being authorized


dwin-gharibi:
## Problem

`CRABBOX_GITHUB_ALLOWED_TEAMS` exists to narrow access *within* an allowed org:
everyone may be in `acme`, but only `acme/engineers` should be able to spend
cloud budget. That narrowing does not currently hold when more than one org is
configured.

`requireAllowedTeamMembership` (`worker/src/github-membership.ts`) receives the
org it is validating, but never filters the configured allowlist down to it:

```ts
const allowed = allowedGitHubTeams(env, org);          // ALL configured teams, every org
if (allowed.length === 0) return;
const allowedKeys = new Set(allowed.map((team) => teamKey(team.org, team.slug)));
for (const team of await userGitHubTeams(accessToken)) {   // /user/teams spans every org
  const teamOrg = team.organization?.login?.toLowerCase() ?? "";
  const teamSlug = team.slug?.toLowerCase() ?? "";
  if (teamOrg && teamSlug && allowedKeys.has(teamKey(teamOrg, teamSlug))) return;  // pass
}
throw new GitHubAuthorizationError(
  `GitHub user ${login} is not a member of an allowed team in ${org}.`,   // claims org scope
);
```

`GET /user/teams` returns the caller's teams across **every** GitHub org they
belong to. Because the allowlist is not narrowed, a membership in org **B**
satisfies the team gate for org **A**.

The error message on the failure path already states the intended semantics
(*"an allowed team in `${org}`"*), and `docs/features/broker-auth-routing.md`
documents them:

> If `CRABBOX_GITHUB_ALLOWED_TEAMS` is set, the user must also belong to at
> least one listed team after org membership passes. Entries are team slugs: use
> `team-slug` for the resolved org, or `org/team-slug` **to qualify the org**.

Qualifying the org currently has no effect on which org an entry can authorize.

## Reproduction

Coordinator config:

```text
CRABBOX_GITHUB_ALLOWED_ORGS  = acme,partner
CRABBOX_GITHUB_ALLOWED_TEAMS = acme/engineers,partner/contractors
```

Actor: an **active member of `acme`** whose only allowed-team membership is
`partner/contractors` — they are deliberately *not* in `acme/engineers`.

Driving the real module with a stubbed `fetch`:

```text
RESULT:  ALLOWED  <- user in partner/contractors was authorized for org=acme
[...]

PR #1315, opened 2026-08-12
fix(worker): reject malformed allowed-team entries instead of widening them

### LINKED ISSUE #1308: A malformed `CRABBOX_GITHUB_ALLOWED_TEAMS` entry silently widens access
## Problem

`parseAllowedGitHubTeam` (`worker/src/github-membership.ts`) splits an allowed-team
entry with a JavaScript `split` limit:

```ts
const [org, slug] = value.includes("/") ? value.split("/", 2) : [defaultOrg, value];
if (!org || !slug) return undefined;
return { org, slug };
```

`String.prototype.split(sep, limit)` **truncates** the result array — unlike Go's
`strings.SplitN` or Python's `split(maxsplit=)`, it does not put the remainder in
the final element. So `"acme/engineers/leads".split("/", 2)` is
`["acme", "engineers"]`, and a nested entry is silently reinterpreted as the
**parent** team.

Separately, entries with an empty segment (`"/operators"`, `"acme/"`) return
`undefined` and are dropped by the `.filter()` in `allowedGitHubTeams`.

## Reproduction

Config `CRABBOX_GITHUB_ALLOWED_TEAMS="acme/engineers/leads"`, actor is a member of
`acme/engineers` but of no team named `leads`:

```text
ALLOWED: config "acme/engineers/leads" silently matched team "acme/engineers"
```

## Impact

Two failure modes, both opening access wider than written, both silent:

1. **Widening.** A nested or typo'd entry grants the parent team.
   `acme/engineers` is typically a far larger group than the intended
   `acme/engineers/leads`.
2. **Disabling the gate entirely.** Dropped entries do not just lose themselves —
   if *every* configured entry is malformed, `allowedGitHubTeams` returns an
   empty array and `requireAllowedTeamMembership`'s
   `if (allowed.length === 0) return` early-out skips the team check for all
   callers. A config consisting only of `"/operators"` silently degrades to
   "org membership is enough".

Neither produces a log line, a startup error, or any other signal. An operator
has no way to notice from the outside.

## Expected behaviour

An allowed-team entry must be exactly `team-slug` or `org/team-slug`. Any other
shape is a configuration error and should fail closed — loudly and immediately —
rather than being reinterpreted or dropped.

This is the posture the same file already takes for
`CRABBOX_GITHUB_REVOKED_USERS`: `requireSafeGitHubRevocationConfig` throws on an
invalid selector so that, as `docs/security.md` puts it, invalid selectors *"fail
all GitHub auth closed until replaced"*. Team entries should be no different — a
malformed allowlist is exactly the case where guessing is most dangerous.

dwin-gharibi:
Closes #1308.

## Summary

`parseAllowedGitHubTeam` used `value.split("/", 2)`. JavaScript's `split` limit
**truncates** rather than keeping the remainder, so `acme/engineers/leads` was
silently reinterpreted as the parent team `acme/engineers`. Entries with an empty
segment were dropped, and if every entry was malformed the team gate switched off
altogether.

Malformed entries now throw, so a bad allowlist fails closed instead of granting
more than it says.

## Motivation

Both failure modes open access **wider** than the configuration reads, and both
are silent — no log line, no startup error. The worst case is not the widening
but the disappearance: a config of only `"/operators"` left
`allowedGitHubTeams` empty, and `requireAllowedTeamMembership`'s
`if (allowed.length === 0) return` early-out then skipped the team check for
every caller, quietly degrading to "org membership is enough".

`worker/src/github-membership.ts` already takes the fail-closed posture for
`CRABBOX_GITHUB_REVOKED_USERS`, where `docs/security.md` states that invalid
selectors "fail all GitHub auth closed until replaced". This applies the same
rule to team entries.

## What changed

| File | Change |
| --- | --- |
| `worker/src/github-membership.ts` | `parseAllowedGitHubTeam` returns `AllowedGitHubTeam` (no longer optional) and throws `GitHubAuthorizationError` on any entry that is not exactly `team-slug` or `org/team-slug`. `allowedGitHubTeams` drops its now-unreachable `.filter()`. |
| `worker/test/github-membership.test.ts` | Three malformed shapes — nested path, missing org, missing slug. |
| `docs/features/broker-auth-routing.md` | Documents that any other shape is rejected and fails auth closed. |

```
 docs/features/broker-auth-routing.md  |  4 +++-
 worker/src/github-membership.ts       | 20 ++++++++++++++------
 worker/test/github-membership.test.ts | 32 ++++++++++++++++++++++++++++++++
 3 files changed, 49 insertions(+), 7 deletions(-)
```

## Design decisions

- **Throw rather than drop.** Returning `undefined` was the root of the second
  failure mode: a dropped entry can silently empty the allowlist and disable the
  gate. Throwing makes a malformed allowlist impossible to ignore, and reuses
  `GitHubAuthorizationError` so it lands on the existing deny path rather than
  surfacing as a 500.
- **Explicit segment count.** The check is `segments.length !== 2 || !org || !slug`,
  which covers the nested path, both empty-segment shapes, and (via the
  `defaultOrg` branch) an empty resolved org — one rule instead of three.
- **Error text names the variable, not the value.** The message is
  `"CRABBOX_GITHUB_ALLOWED_TEAMS contains an invalid entry. Use team-slug or
  org/team-slug."` — enough for an operator to find and fix it, without echoing
  configuration back into an auth error response.

## Testing

New cases in `worker/test/github-membership.test.ts`, driving
`requireFreshGitHubMembership` against a stubbed GitHub API, with the caller
deliberately a member of `example-org/operators` so a widened match would
otherwise succeed:

- nested path `example-org/operators/leads`
- missing org `/operators`
- missing slug `example-org/`

Each must reject with a message naming `CRABBOX_GITHUB_ALLOWED_TEAMS`.

**Proof that the tests observe the defect.** Reverting *only*
`worker/src/github-membership.ts` (keeping the new tests):

```text
FAIL  ... fails closed on an allowed-team entry with a nested path
FAIL  ... fails closed on an allowed-team entry with a missing org
FAIL  ... fails closed on an allowed-team entry with a missing slug
      Tests  3 failed | 17 passed | 1 skipped (21)
```

With the fix restored: `20 passed | 1 skipped (21)`.

**Verified locally** (Node 24 toolchain, worker CI gate in order):

```sh
npm ci --prefix worker
npm run lint --prefix worker
npm run check --prefix worker
npm test --prefix worker          # 1258 passed | 3 skipped (1261)
scripts/check-docs.sh
```

All green. `../verify-fixes.sh gh-team-entry-parse` re-runs the red/green proof
and the full gate.

## Non-goals

- Does not validate that a team actually exists on GitHub; that stays a runtime
  membership question.
- Does not change how a well-formed unqualified `team-slug` resolves.

### DIFF
--- a/docs/features/broker-auth-routing.md
+++ b/docs/features/broker-auth-routing.md
@@ -177,7 +177,9 @@ Login is gated by GitHub org membership before a user token is minted:
 - The user must be an **active** member of an allowed org.
 - If `CRABBOX_GITHUB_ALLOWED_TEAMS` (or `CRABBOX_GITHUB_ALLOWED_TEAM`) is set, the user must
   also belong to at least one listed team after org membership passes. Entries are team
-  slugs: use `team-slug` for the resolved org, or `org/team-slug` to qualify the org.
+  slugs: use `team-slug` for the resolved org, or `org/team-slug` to qualify the org. Any other
+  shape (a nested `org/team/extra` path, or a missing org or slug) is rejected and fails all
+  GitHub auth closed until corrected, so a typo cannot silently widen or disable the team gate.
 
 ### Coordinator secrets for login
 
--- a/worker/src/github-membership.ts
+++ b/worker/src/github-membership.ts
@@ -238,14 +238,22 @@ function allowedGitHubOrgs(env: GitHubMembershipEnv): string[] {
 
 function allowedGitHubTeams(env: GitHubMembershipEnv, defaultOrg: string): AllowedGitHubTeam[] {
   const raw = env.CRABBOX_GITHUB_ALLOWED_TEAMS || env.CRABBOX_GITHUB_ALLOWED_TEAM;
-  return envList(raw)
-    .map((value) => parseAllowedGitHubTeam(value, defaultOrg))
-    .filter((team): team is AllowedGitHubTeam => team !== undefined);
+  return envList(raw).map((value) => parseAllowedGitHubTeam(value, defaultOrg));
 }
 
-function parseAllowedGitHubTeam(value: string, defaultOrg: string): AllowedGitHubTeam | undefined {
-  const [org, slug] = value.includes("/") ? value.split("/", 2) : [defaultOrg, value];
-  if (!org || !slug) return undefined;
+/**
+ * Accepts only `team-slug` or `org/team-slug`. A malformed entry throws rather than
+ * being dropped: silently skipping it would either widen the entry to a different team
+ * or, if every entry is malformed, disable the team gate entirely.
+ */
+function parseAllowedGitHubTeam(value: string, defaultOrg: string): AllowedGitHubTeam {
+  const segments = value.includes("/") ? value.split("/") : [defaultOrg, value];
+  const [org, slug] = segments;
+  if (segments.length !== 2 || !org || !slug) {
+    throw new GitHubAuthorizationError(
+      "CRABBOX_GITHUB_ALLOWED_TEAMS contains an invalid entry. Use team-slug or org/team-slug.",
+    );
+  }
   return { org, slug };
 }
 
--- a/worker/test/github-membership.test.ts
+++ b/worker/test/github-membership.test.ts
@@ -197,6 +197,38 @@ describe("GitHub user-token membership", () => {
     await expect(authenticateRequest(tokenRequest(token), env)).resolves.toBeUndefined();
   });
 
+  it.each([
+    ["a nested path", "example-org/operators/leads"],
+    ["a missing org", "/operators"],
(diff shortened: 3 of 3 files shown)


steipete:
Thank you for identifying and proving the malformed-team-selector fail-open path. The strict parsing work was incorporated into the broader authorization repair in https://github.com/openclaw/crabbox/pull/1316, which also scopes team admission to the selected organization and invalidates warm membership/device proofs when policy changes. That PR merged as `b52245ba52701112992ec247d18037b4c9960be6` with full CI and closes https://github.com/openclaw/crabbox/issues/1308 and https://github.com/openclaw/crabbox/issues/1309, so this narrower proposal is now superseded.

PR #1320, opened 2026-08-13
fix(cli): derive macOS work root from effective SSH user

### LINKED ISSUE #1302: run: default work-root for ssh/static provider uses "ec2-user" instead of -static-user
## Environment
- crabbox version: 0.41.5 (Windows 11 client)
- Provider: `ssh` (static target), macOS remote

## Command
```
crabbox run \
  -provider ssh \
  -target macos \
  -static-host <host> \
  -static-user <user> \
  -static-port 22 \
  --shell 'pwd'
```
(no `-static-work-root` given, so the provider default should apply)

## Observed
SSH itself correctly authenticates as `<user>@<host>:22` (confirmed in the log), but the computed work directory is:

```
workdir=/Users/ec2-user/crabbox/static_100-92-90-71/loam-web
```

...instead of deriving from `-static-user`. Since `/Users/ec2-user` does not exist (and is not writable) on this host, sync fails immediately:

```
mkdir: /Users/ec2-user: Permission denied
write sync manifests: exit status 1
```

Exit code 7.

## Reproducibility
100% reproducible, multiple runs.

## Workaround
Passing `-static-work-root` explicitly (e.g. `-static-work-root /Users/<user>/crabbox-runs`) avoids the bad default and lets sync proceed to the next stage (see #TBD, a separate finalize-step bug hit after working around this one).

## Guess at cause
The default work-root template for this provider looks hardcoded to `ec2-user` (plausibly copied from a cloud-VM provider default) rather than derived from `-static-user`.

SebTardif:
## What problem this solves

For `provider=ssh` / static macOS targets, `-static-user` correctly became the SSH login, but the default work root stayed hardcoded as `/Users/ec2-user/crabbox`. On non-EC2 hosts that path often does not exist or is not writable, so sync fails immediately with `mkdir: /Users/ec2-user: Permission denied` (exit 7).

Closes #1302

## Change

- Derive the macOS default work root as `/Users/<effective-ssh-user>/crabbox` (same shape as the worker default).
- Apply derivation after static/AWS user defaults so `-static-user` is visible.
- Treat any `/Users/<user>/crabbox` path as a platform default so re-derivation stays correct when the user changes.
- Keep AWS EC2 Mac default as `ec2-user` → `/Users/ec2-user/crabbox`.

### Red (main tree + new regression test)

```text
--- FAIL: TestNormalizeTargetConfigStaticMacOSWorkRootFollowsStaticUser (0.00s)
    target_test.go:487: WorkRoot="/Users/ec2-user/crabbox" want "/Users/alice/crabbox"
FAIL
```

### Green (this branch)

```text
=== RUN   TestNormalizeTargetConfigStaticMacOSWorkRootFollowsStaticUser
--- PASS: TestNormalizeTargetConfigStaticMacOSWorkRootFollowsStaticUser (0.00s)
PASS
ok  github.com/openclaw/crabbox/internal/cli
```

### Real behavior proof

**Environment:** local Go test on macOS arm64, package `github.com/openclaw/crabbox/internal/cli`, head of this branch vs `upstream/main` for the product file.

**Command / scenario:**

```bash
# Config shape from the issue: static ssh + macos + -static-user alice, no -static-work-root
cfg.Provider = "ssh"           # static provider
cfg.TargetOS = "macos"
cfg.Static.User = "alice"
cfg.WorkRoot = ""
normalizeTargetConfig(&cfg)
```

**Before (main `defaultWorkRootForTarget` ignores SSH user):**

```text
SSHUser=alice
WorkRoot=/Users/ec2-user/crabbox   # wrong: would mkdir under missing/unwritable home
```

**After (this PR):**

```text
SSHUser=alice
WorkRoot=/Users/alice/crabbox      # matches -static-user; sync can mkdir under the real home
```

AWS macOS managed path still defaults to `ec2-user` + `/Users/ec2-user/crabbox`. Explicit `-static-work-root` still wins.

## Tests

```bash
go test ./internal/cli/ -count=1 -run 'TestNormalizeTargetConfigStaticMacOS|TestNormalizeTargetConfigAWSMacOS|TestIsDefaultWorkRoot|TestNormalizeTargetConfig'
```

### DIFF
--- a/internal/cli/target.go
+++ b/internal/cli/target.go
@@ -51,7 +51,10 @@ func normalizeTargetConfig(cfg *Config) {
 		}
 	}
 	if shouldDeriveTargetWorkRoot(cfg) {
-		cfg.WorkRoot = defaultWorkRootForTarget(cfg.TargetOS, cfg.WindowsMode)
+		// Derive after static/aws user defaults so macOS work roots follow
+		// the effective SSH user (e.g. -static-user), not a hardcoded
+		// ec2-user path that may not exist on the host.
+		cfg.WorkRoot = defaultWorkRootForTarget(cfg.TargetOS, cfg.WindowsMode, cfg.SSHUser)
 	}
 	if isStaticProvider(cfg.Provider) {
 		if cfg.Static.Port != "" && cfg.SSHPort == baseConfig().SSHPort {
@@ -85,11 +88,13 @@ func shouldDeriveTargetWorkRoot(cfg *Config) bool {
 }
 
 func isDefaultWorkRoot(value string) bool {
+	value = strings.TrimSpace(value)
 	switch value {
-	case "", defaultPOSIXWorkRoot, defaultMacOSWorkRoot, defaultWindowsWorkRoot:
+	case "", defaultPOSIXWorkRoot, defaultWindowsWorkRoot:
 		return true
 	default:
-		return false
+		// macOS defaults are user-specific (/Users/<user>/crabbox).
+		return isDefaultMacOSWorkRoot(value)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


steipete:
Thanks @SebTardif — this correctly identified the static macOS work-root bug. It was independently fixed and merged in https://github.com/openclaw/crabbox/pull/1321, which derives the implicit Static SSH macOS root from the effective SSH user while preserving explicit generic/static work roots and the AWS EC2-Mac default.

Closing this pull request as superseded by that merged fix.

PR #1332, opened 2026-08-14
feat(cp): add validated archive fallback for old rsync

### LINKED ISSUE #1280: feat(cp): add a secure fallback when local rsync is too old
## Summary

Make SSH-backed `crabbox cp` usable on supported operator hosts that do not
provide rsync 3.4.3 or newer, including a stock macOS installation, by selecting
a safe alternate transfer path without weakening the existing lease trust
boundary.

## Problem to solve

`crabbox cp` preserves provider-native copy support and otherwise transfers over
the resolved SSH lease. The SSH path currently rejects local rsync versions
older than 3.4.3 before connecting because known sender and receiver
vulnerabilities cross the lease trust boundary.

Stock macOS exposes `/usr/bin/rsync` as OpenRsync with a 2.6.9-compatible version.
In a clean local-container verification, the lease was healthy and reachable,
but `crabbox cp` could not read back a proof artifact because the operator's
rsync did not meet the minimum. The artifact had to be read through the same
task-owned SSH channel outside the `cp` command instead.

This leaves an otherwise supported SSH-backed workflow dependent on a separate
rsync installation and makes automation choose between provisioning a newer
host tool or bypassing Crabbox's file-copy contract.

## Proposed behavior

When a backend exposes a managed POSIX SSH lease and no provider-native copy
capability, `crabbox cp` should select a secure alternate transfer path if local
rsync is absent or below 3.4.3. Upload and download should retain the command's
current path, credential, metadata, and remote-file safety guarantees.

The existing rsync path can remain preferred when both sides satisfy its safety
requirements. If Crabbox cannot establish any safe transfer path, it should
retain an actionable pre-transfer error and should not invoke a rejected rsync
version.

## Acceptance criteria

- On stock macOS with the bundled 2.6.9-compatible OpenRsync and a managed POSIX
  SSH lease, `crabbox cp` can upload and download regular files and directories
  without requiring the operator to install a newer rsync.
- The fallback does not execute an rsync version rejected by the existing 3.4.3
  safety boundary.
- Provider-native copy remains preferred, and the existing rsync path remains
  available when its local and remote requirements are satisfied.
- Resolved usernames, keys, certificates, agents, host-key policy, proxy routes,
  and other credential material retain the current private transport boundary
  and do not appear in command arguments or normal output.
- Downloads continue to reject lease-provided symlinks and special files,
  discard remote ownership and group metadata, and apply normalized host modes.
  Uploads preserve the existing explicit `-L` behavior for host-side symlinks.
- An environment with neither the normal rsync path nor a safe fallback fails
  before transfer with an actionable diagnostic.
- Automated coverage exercises upload, download, file and directory handling,
  the stock-macOS version signal, credential redaction, rejected remote entry
  types, and fallback cleanup.

## Affected area

- `crabbox cp`
- provider-neutral resolved SSH lease transport
- macOS operator environments using the system-provided OpenRsync

## Non-goals

- Lowering or bypassing the rsync 3.4.3 security floor.
- Changing Git workspace synchronization, which has a different manifest and
  transfer contract.
- Expanding `cp` to native Windows SSH targets that do not already expose a
  compatible copy backend.
- Changing provider-native copy behavior.

## Alternatives considered

- **Require operators to install a newer rsync.** This preserves the current
  implementation but leaves the stock configuration of a supported operator
  platform unable to use SSH-backed `cp`.
- **Tell callers to invoke SSH transfer tools directly.** This can recover an
  artifact, but it moves resolved transport and file-materialization behavior
  outside the provider-neutral `crabbox cp` contract.
- **Allow the bundled older rsync.** This conflicts with the documented lease
  trust boundary and is not an acceptable fallback.

## Supporting context

- Current source at `7f450bfa950fee1a5eb119296dd9b5508b11298f`
  documents the 3.4.3 minimum in `docs/features/ssh-transport.md` and
  `docs/commands/cp.md`; `internal/cli/cp.go` enforces it before connection.
- Focused current-source tests pass and explicitly classify the
  2.6.9-compatible OpenRsync banner as below the safe boundary.
- The failure was reproduced end to end across two separate Macs. A MacBook
  running macOS 27.0 arm64, Crabbox 0.40.0, and the system OpenRsync
  2.6.9-compatible client acted as the operator. A Mac Studio running macOS
  26.5.2 arm64 was the SSH-backed peer. Direct batch SSH reached and hashed the
  peer fixture, while both `crabbox cp` download and upload exited 2 with
  `SSH cp requires rsync 3.4.3 or newer for secure transfers; found 2.6.9`;
  neither destination was created.
- The currently installed Crabbox 0.41.2 and current source retain the same
  version gate.
- https://github.com/openclaw/crabbox/issues/1067 requested the original
  provider-neutral SSH copy capability and was completed by
  https://github.com/openclaw/crabbox/pull/1154. That closure explicitly shipped
  the local rsync 3.4.3 requirement; this request is the narrower follow-up for
  operator hosts that cannot use that transport path as installed.

steipete:
## Summary

- prefer the existing secure rsync transport when local rsync is 3.4.3 or newer
- fall back on POSIX operator hosts to a Go-created/validated tar+gzip stream over the resolved private SSH session for native Linux and macOS leases
- bound compressed, expanded, per-file, and entry counts; reject traversal, duplicate filesystem aliases, links, hard links, and special files on download
- make publication durable and target-scoped with staged validation, authenticated transaction ownership, crash recovery, and fail-closed tar diagnostics
- preserve upload permissions/mtimes and normalize downloaded modes while preserving mtimes
- document platform scope, replacement semantics, and pre/post-publication recovery behavior

## Verification

- `go test ./...`
- `go test -race ./internal/cli -run 'Test(CopyArchive|CreateCopyArchive|ExtractValidatedCopyArchive|CopyOverResolvedSSH|RemoteCopyArchive|LocalCopyArchive|ValidatedCopyArchive|RecoverCopyArchive)' -count=1`
- `go vet ./...`
- `GOOS=windows GOARCH=amd64 go test -c -o /tmp/crabbox-cli-1280.test.exe ./internal/cli`
- structured autoreview: clean, no accepted/actionable findings

## Live proof

Built the branch CLI on macOS with stock `/usr/bin/rsync` reporting OpenRsync 2.6.9, then used a task-owned local-container SSH lease to upload and download a regular file and a trailing-slash directory. The file SHA-256 matched across host, lease, and downloaded copy; file counts matched 4/4; mode was normalized to `0644`; mtime was preserved. The lease was deleted after proof.

No configuration or secret changes.

Closes https://github.com/openclaw/crabbox/issues/1280

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 0.41.7 - Unreleased
 
+### Added
+
+- Added a checksummed, validated archive fallback for SSH-backed `cp` from POSIX operator hosts to native Linux or macOS leases (not WSL2) when local rsync is missing or older than 3.4.3, including stock macOS OpenRsync. Thanks @coygeek.
+
 ### Fixed
 
 - Rejected native Jujutsu workspaces before Git-manifest sync can fall through to an outer checkout, while preserving colocated Git workspaces and `--no-sync`. Thanks @atimmer.
--- a/docs/commands/cp.md
+++ b/docs/commands/cp.md
@@ -3,7 +3,7 @@
 `crabbox cp` copies files or directories between the host and a Crabbox-owned
 lease. It resolves the lease id or slug first. Providers with a native copy
 backend keep using that backend; SSH-backed leases otherwise transfer through
-the resolved SSH transport with rsync.
+the resolved SSH transport with rsync or a validated archive stream.
 
 ```sh
 crabbox cp --provider docker-sandbox --id blue-box ./coverage.xml SANDBOX:/tmp/coverage.xml
@@ -43,8 +43,11 @@ is a path on the resolved remote host.
 
 Provider-native copy remains the first choice when the backend implements it.
 Otherwise, Crabbox accepts any provider that resolves a managed SSH lease and
-uses rsync over the same `SSHTarget` used by `run`, sync, and interactive SSH.
-Delegated providers with neither transport fail clearly.
+uses the same `SSHTarget` used by `run`, sync, and interactive SSH. Crabbox
+prefers rsync 3.4.3 or newer. On a POSIX operator host, if the local client is missing or older, it uses a
+Go-owned tar+gzip archive stream over SSH instead; it never invokes the rejected
+rsync client, scp, or another copy command. Delegated providers with neither
+transport fail clearly.
 
 The resolved SSH user, key/certificate paths, host-key policy, and ProxyCommand
 are rendered into a mode-`0600` temporary OpenSSH config. The Crabbox-launched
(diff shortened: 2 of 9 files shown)


steipete:
Merging over the CodeQL check: its flagged high alert (js/clear-text-logging, worker/src/fleet.ts:14894, alert #59) was created on main on 2026-07-31 and this PR touches only Go archive code + docs — the 'new alert' attribution is an artifact of the scan's own config mismatch (missing /language:swift configuration). The pre-existing alert should be triaged separately.

PR #1349, opened 2026-08-15
feat(images): add explicit catalog variant selectors

obviyus:
## Summary

- Add generic AWS catalog-only image variants without replacing the scoped default image.
- Separate an image's full SDK/runtime capability inventory from the small set of explicit selectors allowed to activate a specialized variant.
- Keep catalog-only promotion fail-closed through a dedicated coordinator route and a rollback-safe variant namespace that older coordinators never enumerate.
- Make promotion, retirement, and cleanup atomic across Cloudflare Durable Object and Node/PostgreSQL coordinators, with serializable retries for shared PostgreSQL state.
- Add safe external-variant retirement through `image delete --catalog-only`, preserve ordinary promotion/delete behavior, and clear variant metadata when an image is promoted normally.
- Remove the Telegram Desktop installer, workflow, bake mode, and consumer-specific documentation from Crabbox; those assets belong in the downstream consumer repository.

## Root cause and design

The original proposal treated every declared SDK/runtime capability as an activation key. A specialized image that contained both its distinguishing toolkit and a common runtime such as Node could therefore be selected by a Node-only request.

Catalog-only promotion now requires at least one explicit `--variant-sdk` or `--variant-runtime`. These flags also declare the capability, but ordinary `--sdk` and `--runtime` entries remain inventory only. The coordinator persists `variantSelectors` separately from `capabilities`, and a catalog-only image is eligible only when the lease names one matching selector. Once activated, it must still satisfy every requested capability.

The dedicated `POST /v1/images/<id>/promote-catalog` route remains authoritative: older coordinators return 404 and the CLI never falls back to ordinary promotion. AWS stores variants in a dedicated namespace, so rolling coordinator code back cannot make common inventory activate a specialized image. Ordinary and variant roles remain distinct even when they share an AMI ID.

Existing-image mutations are serialized around provider validation, then committed atomically in coordinator storage. Cloudflare uses its native Durable Object transaction. Node/PostgreSQL uses `SERIALIZABLE` transactions with bounded retries on serialization failures and deadlocks. Ordinary re-promotion preserves capability inventory and removes the variant role atomically; `image delete --catalog-only` retires external variants without deleting provider-owned AMIs. Azure rejects catalog-only publication and retirement.

## Verification

- `go test -race ./internal/cli -run '^TestImage(Promote|Delete)' -count=1`
- `npm test --prefix worker -- test/fleet.test.ts test/coordinator-runtime.test.ts test/server-support.test.ts test/postgres-storage.test.ts test/oauth-browser-binding.test.ts test/pairing.test.ts test/private-aws-workspace.test.ts` — 697 passing
- `npm run check:node --prefix worker`
- `git diff --check`
- Source-blind CLI retirement contract: 5/5 passing
- Codex autoreview plus independent storage/lifecycle review: clean
- Exact-head CI: https://github.com/openclaw/crabbox/actions/runs/32002935281 (rebased head `06e60308`; the prior run exposed the Node-only typecheck gap fixed in the stack)

Focused regressions cover namespace isolation from old readers, explicit activation and incidental-capability non-activation, same-AMI ordinary/variant roles, atomic ordinary/variant transitions, scope and Region isolation, external-variant retirement, invalid inputs, concurrent mutation ordering, Cloudflare transaction rollback, PostgreSQL serializable retries/rollback failure, old-coordinator failure, Azure rejection, response/output compatibility, and implicit Region routing without widening provider-delete ownership.

## Live proof boundary

An isolated changed-path Cloudflare-plus-AWS run was attempted on August 17, 2026. The repository has no staging coordinator or PR-head deployment workflow: its deploy workflow targets the fixed production Worker, while a separate preview Worker does not inherit the production AWS secrets. A short-lived brokered AWS control lease successfully selected the current promoted AMI and reached SSH readiness, but—correctly for the security boundary—exposed no instance role that could authorize an isolated coordinator. Using a production version preview would share production Durable Object state, so it was rejected as non-isolated.

The changed path is therefore proven by real Cloudflare/PostgreSQL transaction implementations, source-blind CLI HTTP behavior, injected persistence/restart/rollback/interleaving regressions, exact-head CI, and a separate real-AWS baseline—but not by a combined isolated coordinator+AMI run. This is the concrete infrastructure gap; no production coordinator or durable state was mutated.

### DIFF
--- a/docs/commands/image.md
+++ b/docs/commands/image.md
@@ -118,9 +118,12 @@ Flags:
 --os-version <version>   numeric OS version present in the image
 --sdk <name=version>     SDK present in the image (repeatable)
 --runtime <name=version> runtime present in the image (repeatable)
+--variant-sdk <name=version>     SDK that explicitly activates a catalog-only image (repeatable)
+--variant-runtime <name=version> runtime that explicitly activates a catalog-only image (repeatable)
 --browser                image includes browser support
 --webview2               image includes Microsoft WebView2
 --desktop                image includes desktop support
+--catalog-only           publish an AWS capability variant without changing the default image
 --fast-snapshot-restore  enable AWS Fast Snapshot Restore for the backing snapshots
 --fsr-az <az>            availability zone for Fast Snapshot Restore (repeatable)
 --json                   print the promoted image record as JSON
@@ -143,6 +146,17 @@ Capability declarations make the AMI eligible for capability-aware selection.
 Versions use numeric dot notation, for example `--os-version 15.5`,
 `--sdk xcode=16.4`, or `--runtime node=24.2`. Omitted capabilities remain
 unknown and do not satisfy an explicit image requirement.
+
+AWS catalog-only images keep specialized variants out of the scoped default.
+Each promotion requires at least one `--variant-sdk` or `--variant-runtime`;
+the variant flag both declares the capability and records it as an activation
+selector. Ordinary `--sdk` and `--runtime` flags remain capability inventory
+only. For example, an image promoted with `--catalog-only --variant-sdk
+toolkit=2.0 --runtime node=24` is activated by an exactly matching `--image-sdk
+toolkit=2.0` request, not by a Node-only request. After activation, every
+requested capability must still match. Catalog-only promotion uses the
+dedicated `POST /v1/images/<id>/promote-catalog` route and fails closed against
+older coordinators.
 
 Add `--fast-snapshot-restore` plus one or more `--fsr-az` values when the
 promoted image backs hot lanes that need immediate EBS snapshot reads:
--- a/docs/commands/image.md
+++ b/docs/commands/image.md
@@ -14,6 +14,7 @@ crabbox image promote ami-1234567890abcdef0 --target macos --region us-east-1 --
 crabbox image promote snapshot-devtools --provider azure --target linux --region westeurope
 crabbox image fsr-status ami-1234567890abcdef0 --region us-west-2 --fsr-az us-west-2a
 crabbox image delete ami-1234567890abcdef0 --region eu-west-1
+crabbox image delete ami-external --catalog-only
 crabbox image delete my-managed-image --provider azure --region westeurope
 crabbox image delete my-machine-image --provider gcp --region europe-west1-b --project example-project
 crabbox image delete 123456789 --provider hetzner --region fsn1
@@ -241,6 +242,7 @@ Delete a Crabbox-created provider image.
 
 ```sh
 crabbox image delete ami-1234567890abcdef0 --region eu-west-1
+crabbox image delete ami-external --catalog-only
 crabbox image delete my-managed-image --provider azure --region westeurope
 crabbox image delete my-machine-image --provider gcp --region europe-west1-b --project example-project
 crabbox image delete 123456789 --provider hetzner --region fsn1
@@ -252,11 +254,30 @@ Flags:
 --provider <name>   image provider: aws, azure, gcp, or hetzner (default aws)
 --region <name>     region, location, or zone containing the image
 --project <name>    GCP project containing the image
+--catalog-only      unpublish every AWS catalog-only role without deleting the AMI
 ```
 
 AWS deletion deregisters the AMI and then deletes the EBS snapshots referenced by
 its block-device mappings. Azure deletion removes the managed image or disk
 snapshot. GCP deletion removes the machine image or disk snapshot.
+
+For an externally sourced AWS variant, use `--catalog-only` to remove every
+catalog-only role for the AMI while leaving the provider-owned AMI untouched.
+The command uses the dedicated `DELETE /v1/images/<id>/promote-catalog` route,
+never falls back to provider deletion, and fails with an upgrade error when the
+coordinator does not support safe retirement.
+
+Because AWS AMI IDs are regional, `--catalog-only --region <region>` retires
+only variant roles in that Region. Omitting `--region` is the explicit
+convenience form that retires every regional variant role with the same AMI ID.
+Its text output is:
+
+```text
+retired catalog-only image=ami-external provider=aws variants=1
+```
+
+Without `--catalog-only`, `image delete` keeps the existing provider deletion
+and ownership checks.
 
 Hetzner deletion runs directly without coordinator admin auth. It rejects
 `--project`, requires any supplied `--region` to match the recorded source
--- a/internal/cli/coordinator.go
+++ b/internal/cli/coordinator.go
@@ -286,6 +286,8 @@ type CoordinatorImage struct {
 	PromotedAt           string                           `json:"promotedAt,omitempty"`
 	FastSnapshotRestores []CoordinatorFastSnapshotRestore `json:"fastSnapshotRestores,omitempty"`
 	Capabilities         *imageCapabilities               `json:"capabilities,omitempty"`
+	CatalogOnly          bool                             `json:"catalogOnly,omitempty"`
+	VariantSelectors     *imageVariantSelectors           `json:"variantSelectors,omitempty"`
 }
 
 type CoordinatorFastSnapshotRestore struct {
@@ -357,6 +359,8 @@ type CoordinatorImageRef struct {
 	FastSnapshotRestore    bool
 	FastSnapshotRestoreAZs []string
 	Capabilities           imageCapabilities
+	CatalogOnly            bool
+	VariantSelectors       imageVariantSelectors
 }
 
 type CoordinatorGitHubLoginStart struct {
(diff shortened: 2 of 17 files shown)


obviyus:
@clawsweeper re-review

obviyus:
@clawsweeper re-review

obviyus:
@clawsweeper re-review

obviyus:
@clawsweeper re-review

obviyus:
@clawsweeper re-review

obviyus:
@clawsweeper re-review

steipete (changes requested):
This adds explicit selector-gated AWS catalog variants without replacing the ordinary scoped default. Separating activation selectors from capability inventory is the right abstraction, and the focused test coverage is strong, but the current persisted representation has two blocking role/rollback failures.

Production LOC: +285/-40 (net +245) | Tests: +500/-0 | Docs: +14/-0

## Findings

### [P1] Store selector-gated variants outside the legacy catalog namespace

Catalog-only promotion writes the new role into the existing `image:aws:catalog:` namespace at https://github.com/openclaw/crabbox/blob/9e3c4f357e787bfb1dca65288d6fcfd616a5b0dc/worker/src/fleet.ts#L24425-L24431. A rollback to current `main` still scans that namespace and selects solely by general capability satisfaction, with no knowledge of `catalogOnly` or `variantSelectors`: https://github.com/openclaw/crabbox/blob/bba654d60c65137eac8783b1e110861092e82d16/worker/src/fleet.ts#L24379-L24403.

Cloudflare Worker rollback does not roll back connected Durable Object data. I reproduced the resulting failure with a preserved default plus a newer catalog-only image containing both `toolkit=2.0` and ordinary `node=24` inventory: the rolled-back reader selected the variant for a Node-only request. That restores the exact incidental-activation bug this PR is intended to prevent.

The dedicated `promote-catalog` route correctly fails closed when a new CLI talks to an old coordinator, but it cannot protect old code reading already-persisted new records. The best fix is a new variant-only namespace that legacy readers cannot enumerate; new readers should union the ordinary and variant namespaces.

That namespace also needs a complete lifecycle owner. Ordinary re-promotion currently overwrites only the newly derived scope key, so changing OS/server-type scope leaves the old selector-gated record active. External AMIs can be catalog-promoted, but `storedImageMetadata()` only searches ordinary promoted/created records at https://github.com/openclaw/crabbox/blob/9e3c4f357e787bfb1dca65288d6fcfd616a5b0dc/worker/src/fleet.ts#L24124-L24129, so `image delete` rejects them as unowned while manual provider deletion leaves selectable durable state. On ordinary promotion, remove every prior variant entry for that AMI across scopes, and add an explicit unpublish/retirement path that does not require deleting an externally owned provider image. Please add old-reader rollback/re-upgrade, scope-changing cleanup, and external-variant retirement regressions.

Cloudflare contract: https://developers.cloudflare.com/workers/versions-and-deployments/rollbacks/

### [P1] Preserve default and variant as distinct roles when they share an AMI ID

Selection currently builds `[selected, ...catalog.values()]` and deduplicates by `image.id` before applying the catalog-only gate: https://github.com/openclaw/crabbox/blob/9e3c4f357e787bfb1dca65288d6fcfd616a5b0dc/worker/src/fleet.ts#L24483-L24504. If the current default AMI is promoted catalog-only under the same ID, the catalog record overwrites the ordinary default record. I reproduced a Node-only capability request then failing with `no promoted AWS linux image satisfies`, even though the unchanged scoped default still satisfies it.

The same role collision also loses capability inventory on the reverse transition. `known?.capabilities` wins over the newer catalog record at https://github.com/openclaw/crabbox/blob/9e3c4f357e787bfb1dca65288d6fcfd616a5b0dc/worker/src/fleet.ts#L24223-L24231, so ordinary re-promotion of that same AMI can drop capabilities declared by the variant. The existing cleanup test passes `known === undefined`, so it misses the real coordinator path.

Treat ordinary-default and selector-gated-variant records as separate policy roles rather than deduplicating solely by provider image ID, and make capability preservation/cleanup role-aware. Add a route-level regression for ordinary → catalog-only → ordinary promotion with the same AMI ID, covering both an inventory-only request and the explicit selector.

### [P3] Keep promotion validation closed at both boundaries

The CLI caps `--sdk` and `--variant-sdk` independently at 32 entries, then merges them without rechecking the total at https://github.com/openclaw/crabbox/blob/9e3c4f357e787bfb1dca65288d6fcfd616a5b0dc/internal/cli/image_capabilities.go#L32-L69. I reproduced 32 ordinary SDK declarations plus one distinct variant SDK passing local validation and producing a 33-entry map; the coordinator rejects that payload because its contract caps the combined map at 32. Runtimes have the same mismatch. Revalidate the merged map locally so the CLI and coordinator expose one contract.

The changed Worker query parser also accumulates versions in a normal `{}` at https://github.com/openclaw/crabbox/blob/9e3c4f357e787bfb1dca65288d6fcfd616a5b0dc/worker/src/fleet.ts#L17004-L17015. Assigning an invalid `__proto__` capability invokes the inherited setter instead of creating an own property, so the declaration disappears before name validation rather than being rejected. Use a null-prototype object or `Map` and add a malformed-name regression.

## Verdict and proof

Best-fix verdict: **wrong storage boundary on this head**. A separate variant namespace plus role-aware selection/cleanup resolves both blocking failures; merely rejecting same-ID promotion would not solve rollback safety.

Code read: CLI promotion/client path, Worker promotion/storage/selection path, capability normalization, current-main reader behavior, docs, focused Go/Vitest coverage, and capability-selection history from https://github.com/openclaw/crabbox/pull/1036.

Proof run against the clean current-main merge:

- `go test -race ./internal/cli -run '^TestImagePromote' -count=1` — passed
- `npm test --prefix worker -- test/fleet.test.ts` — 584 passed
- same-ID default/variant regression — failed on the PR head; the candidate-retention simplification fixes that selection failure and keeps all 584 fleet tests green
- current-main rollback regression — failed by selecting `ami-variant` instead of `ami-default`
- combined 32+1 SDK limit regression — failed by merging 33 entries without a local error

GitHub reports the head mergeable and CI is green. After these code fixes, the PR's already-declared deployed coordinator plus real AWS AMI proof remains a separate before-merge requirement for persistence, selected-image identity, explicit activation, incidental non-activation, cleanup, and rollback/re-upgrade behavior.

steipete:
Maintainer repair and exact-head verification are complete on rebased head `06e60308c809b4b38cbe419f3c3e325987b4c0f1`.

The original selector separation remains, but the repaired head now owns the full lifecycle boundary: variants use a rollback-safe namespace, ordinary and variant roles remain distinct for the same AMI, image mutations serialize across provider waits, catalog transitions are atomic on both Durable Object and PostgreSQL storage, PostgreSQL retries `SERIALIZABLE` conflicts on fresh clients, external variants have a metadata-only retirement path, and Region routing remains separate from provider-delete ownership.

Focused local proof:

- `go test -race ./internal/cli -run '^TestImage(Promote|Delete)' -count=1`
- `npm test --prefix worker -- test/fleet.test.ts test/coordinator-runtime.test.ts test/server-support.test.ts test/postgres-storage.test.ts test/oauth-browser-binding.test.ts test/pairing.test.ts test/private-aws-workspace.test.ts` — 703 passed after rebasing onto current main
- `npm run check:node --prefix worker`
- `git diff --check`
- Source-blind CLI retirement contract — 5/5 passed
- Fresh Codex autoreview and independent storage/lifecycle review — clean

Rebased exact-head CI is green: https://github.com/openclaw/crabbox/actions/runs/32002935281

Live-proof gap and explicit maintainer disposition: I attempted to establish an isolated changed-path Cloudflare coordinator with real AWS execution. The repository has no staging/PR-head coordinator deployment; a separate preview cannot inherit the production AWS secrets, while a production version preview would share production Durable Object state. A short-lived brokered AWS control lease did select the current promoted AMI and reach SSH readiness, but intentionally had no instance role that could authorize a separate coordinator. I rejected mutating the production Worker or durable state. The PR body now records this concrete infrastructure limitation. The deterministic owner-boundary, persistence, rollback, interleaving, source-blind HTTP, exact-head CI, and separate real-AWS baseline evidence are accepted for this landing; combined isolated coordinator+AMI proof was not misrepresented as completed.

Review metric against rebased current main: production +767/-123 (net +644), tests net +1,912 after discounting conflict-resolved moves (raw +3,116/-1,204), docs +35/-0; no generated or lockfile churn is included. The production growth implements the new selector capability plus its cross-runtime transactional ownership boundary and safe retirement contract. Current main's resumable AWS image-deletion state machine was preserved during the rebase rather than duplicated in this PR.

steipete:
@clawsweeper re-review

steipete (dismissed):
Approved after the catalog lifecycle rewrite and exact-head verification. The blocking rollback, same-AMI role collision, retirement, Region, mutation-ordering, and cross-runtime transaction findings are resolved. Exact-head CI is green; the documented combined Cloudflare+AWS live-proof gap is accepted under the concrete isolation/credential infeasibility recorded in the PR body and maintainer proof comment.

steipete:
@clawsweeper re-review

steipete (approved):
Approved after the catalog lifecycle rewrite and exact-head verification. The blocking rollback, same-AMI role collision, retirement, Region, mutation-ordering, and cross-runtime transaction findings are resolved. Exact-head CI is green; the documented combined Cloudflare+AWS live-proof gap is accepted under the concrete isolation/credential infeasibility recorded in the PR body and maintainer proof comment.

PR #1350, opened 2026-08-15
feat: catalog-only image promotion for capability variants

obviyus:
## What Problem This Solves

Resolves a problem where operators who bake a capability *variant* image (a desktop image with an extra application only some lanes need) had no safe way to publish it: `crabbox image promote` always rewrote the scoped default alongside the capability catalog, so a variant silently became every matching lease's image, and the only workaround was to re-promote the generic image afterwards.

## Why This Change Was Made

`crabbox image promote --catalog-only` (AWS) writes only the capability-catalog record. Ordinary leases keep the existing scoped default; leases that request the capability (`warmup --image-sdk name=version`, `--image-runtime`, `--image-require-desktop`, …) select the newest catalog image satisfying every requirement, which already covers catalog-only entries. A catalog-only promotion without any capability declaration is rejected (CLI and Worker) because nothing could ever select it. Azure rejects the flag. Existing `image promote` calls are unchanged.

## User Impact

Operators can publish variant images (`--desktop --sdk <app>=<version> --catalog-only`) without touching the generic default; consumers opt in per lease with `--image-sdk <app>=<version>` and Crabbox fails before leasing when no matching image exists. Follow-up: https://github.com/openclaw/crabbox/pull/1349 uses this to ship a Telegram Desktop variant while keeping the generic desktop image unchanged.

## Evidence

- `npm test --prefix worker` → 37 files passed, 2 skipped; 1357 tests passed, 3 skipped — new cases: catalog-only writes the catalog key only; default selection unchanged after a catalog-only promotion; requirement-based selection returns the catalog-only image; catalog-only without capabilities → 400 `catalog_only_capabilities_required`.
- `npm run format:check`, `npm run lint`, `npm run check` (worker) → clean.
- `go test ./internal/cli/...` → ok (flag parsing/serialization `catalogOnly=true`, `--catalog-only` without capabilities exits 2, Azure rejects).
- `go build ./cmd/crabbox` → built; `crabbox image promote --help` shows `--catalog-only`.
- `node scripts/build-docs-site.mjs` → built.

### DIFF
--- a/docs/commands/image.md
+++ b/docs/commands/image.md
@@ -119,6 +119,7 @@ Flags:
 --browser                image includes browser support
 --webview2               image includes Microsoft WebView2
 --desktop                image includes desktop support
+--catalog-only           publish an AWS capability variant without changing the default image
 --fast-snapshot-restore  enable AWS Fast Snapshot Restore for the backing snapshots
 --fsr-az <az>            availability zone for Fast Snapshot Restore (repeatable)
 --json                   print the promoted image record as JSON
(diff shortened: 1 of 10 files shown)


obviyus:
Folded into https://github.com/openclaw/crabbox/pull/1349 so the primitive and its first consumer land as one review.

PR #1362, opened 2026-08-16
docs: clarify portable work root

### LINKED ISSUE #1269: feat(crabbox): add a unified --workdir flag across providers
feat(crabbox): add a unified --workdir flag across providers

## Summary

Target repository: `openclaw/crabbox`

`crabbox run` accepts provider-specific workdir flags such as `--local-container-work-root`, `--apple-vm-work-root`, `--apple-container-work-root`, `--agent-sandbox-workdir`, `--anthropic-sandbox-runtime-workdir`, and `--ascii-box-workdir`, but it has no general `--workdir` flag.

As a result, this command fails:

```sh
crabbox run --workdir /tmp/foo -- pnpm test
```

```text
flag provided but not defined: -workdir
```

Scripts and skills that operate against an abstract provider must repeat the workdir override for every provider they support. When a provider is added later, those scripts silently omit the override unless they are also updated.

## Problem to solve

The current provider-specific workdir model creates a portability tax:

- Portable skills and recipe scripts must enumerate providers in advance. Adding a provider with its own workdir setting requires updating every script that needs a workdir override.
- The locally verified `crabbox run --help` output lists more than twenty provider-specific workdir flags but no general equivalent, making portable workdir configuration appear unsupported.
- The upstream `docs/providers/local-container.md` documentation demonstrates provider-specific configuration through `localContainer.workRoot`, while sibling providers expose their own namespaced workdir settings. That model works in `.crabbox.yaml`, where configuration is already provider-namespaced, but it does not provide a portable CLI equivalent.

## Proposed behavior

`crabbox run`, `crabbox warmup`, and `crabbox prewarm` accept a unified `--workdir <path>` flag. When set, the flag overrides the selected provider's default workdir through that provider's existing setting:

- `local-container` → `--local-container-work-root <path>`
- `apple-container` → `--apple-container-work-root <path>`
- `apple-vm` → `--apple-vm-work-root <path>`
- `apple-vz` → `--apple-vz-work-root <path>`
- `apple-machine` → `--apple-machine-workdir <path>`
- `agent-sandbox` → `--agent-sandbox-workdir <path>`
- `anthropic-sandbox-runtime` → `--anthropic-sandbox-runtime-workdir <path>`
- `ascii-box` → `--ascii-box-workdir <path>`
- `static` (SSH) → `static.workRoot` for that lease

If the selected provider does not expose a configurable workdir, Crabbox should reject `--workdir` with an explicit error that identifies the provider and explains that it does not support the override.

An explicit provider-specific flag takes precedence over `--workdir`. The unified flag therefore acts as a portable default without changing existing provider-specific behavior.

## Acceptance criteria

- `crabbox run --provider local-container --workdir /tmp/x --shell pwd` runs `pwd` inside `/tmp/x` on the lease.
- `crabbox run --provider local-container --local-container-work-root /tmp/override --workdir /tmp/x --shell pwd` resolves to `/tmp/override`, confirming that the provider-specific flag takes precedence.
- `crabbox run --provider hetzner --workdir /tmp/x -- uname` does not silently ignore the override. If the provider does not expose a configurable workdir, Crabbox exits with an explicit error identifying that provider.
- `crabbox run --provider does-not-exist --workdir /tmp/x -- true` reports the unknown provider before evaluating the workdir override.
- The same `--workdir` flag is accepted by `crabbox warmup` and `crabbox prewarm`, with the same mapping and precedence behavior as `crabbox run`.
- Existing commands that omit `--workdir` retain their current behavior and provider defaults.

## Affected area

- CLI flag surface for `crabbox run`, `crabbox warmup`, and `crabbox prewarm`.
- Existing flag-set helpers and per-provider configuration wiring. The relevant CLI paths were verified in `internal/cli/run.go` and `internal/cli/prewarm.go`; `warmup` is implemented in a sibling file under `internal/cli/` and dispatched from `internal/cli/app.go`.
- Documentation in `docs/cli.md`, `docs/commands/run.md`, `docs/commands/warmup.md`, `docs/commands/prewarm.md`, and `docs/features/configuration.md`.

## Non-goals

- Renaming or changing the behavior of provider-specific flags. They remain authoritative when both forms are supplied.
- Adding `--workdir` to commands other than `run`, `warmup`, and `prewarm`.
- Adding a generic `workRoot` key to `.crabbox.yaml`. Configuration remains provider-namespaced.
- Changing any provider's default workdir when `--workdir` is omitted.

## Alternatives considered

- **Continue using provider-specific configuration only.** This keeps the CLI surface smaller but requires every portable script to enumerate providers. The maintenance cost falls on skills and recipes instead of being handled once by Crabbox.
- **Replace the provider-specific flags with `--workdir` and an optional provider name.** This would create a cleaner surface, but it would be breaking: existing flags, scripts, and configuration keys would stop working.
- **Add only a short alias for the unified flag.** This would require the same provider mapping while being less discoverable in `crabbox run --help`. A single-letter alias would also need to avoid collisions across the existing flag surface.

## Supporting context

Verified with the locally installed `crabbox 0.40.0`:

```console
$ crabbox run --workdir /tmp/foo -- true
flag provided but not defined: -workdir
```

Provider-specific flags exist for the most common local providers:

```console
$ crabbox run --help 2>&1 | grep -E 'work(root|dir)' | head
  -agent-sandbox-workdir string
  -apple-container-work-root string
  -apple-machine-workdir string
  -apple-vm-work-root string
  -apple-vz-work-root string
  -ascii-box-workdir string
  -local-container-work-root string
```

A portable skill or recipe should be able to use one command regardless of the selected provider:

```sh
crabbox run --provider "$PROVIDER" --workdir "$REMOTE_WORKDIR" -- pnpm test
```

Today, that command must be rewritten for each provider. If a script omits a provider-specific branch, the command silently runs in the provider's default workroot instead of the requested one.

The closure signal is a portable `--workdir` flag accepted by `run`, `warmup`, and `prewarm`, with explicit precedence for provider-specific flags and a clear error when the selected provider does not expose a configurable workdir.

steipete:
## Summary

- document `CRABBOX_WORK_ROOT` as the provider-neutral one-run base-root override
- clarify that ordinary SSH-backed runs derive `<root>/<lease>/<repository>` rather than treating the value as an exact command directory
- document provider-specific root precedence and Actions hydration's canonical workspace precedence

Crabbox intentionally does not add a unified `--workdir` alias: that name would imply an exact remote `chdir` while workspace placement remains owned by provider adapters and hydration.

Closes https://github.com/openclaw/crabbox/issues/1269

## Verification

- `scripts/check-docs.sh`
- focused work-root, provider-precedence, and Actions-hydration Go tests
- Go formatting check
- `git diff --check`
- P1 autoreview clean

### DIFF
--- a/docs/commands/run.md
+++ b/docs/commands/run.md
@@ -33,6 +33,29 @@ The trailing command after `--` is sent to the box verbatim as argv. Use
 `--shell` to run it through the remote shell instead, for multi-statement
 snippets, pipes, or shell expansion.
 
+## Remote workspace root
+
+Use `CRABBOX_WORK_ROOT` to change the portable base root for one run without
+selecting a provider-specific flag:
+
+```sh
+CRABBOX_WORK_ROOT=/srv/crabbox \
+  crabbox run --provider "$PROVIDER" -- pnpm test
+```
+
+This changes the base root, not the command's exact working directory. A normal
+SSH-backed run derives `<root>/<lease>/<repository>`, syncs the checkout there,
+and uses that derived repository workspace as the command PWD. There is no
+provider-neutral exact-chdir flag; provider adapters and hydration own their
+workspace semantics.
+
+An explicitly configured provider-specific work root or workdir takes
+precedence over the generic root. If the lease has a valid GitHub Actions
+hydration marker, its canonical `WORKSPACE` takes precedence over both roots
+for sync and command execution. See [Configuration](../features/configuration.md#work-roots)
+for configuration precedence and [Sync](../features/sync.md#remote-workspace-path)
+for path resolution.
+
 ## Leasing model
 
 If `--id` is omitted, Crabbox creates a fresh, non-kept lease and releases it
--- a/docs/features/configuration.md
+++ b/docs/features/configuration.md
@@ -108,6 +108,7 @@ class: beast             # standard | fast | large | beast
 serverType: c7a.48xlarge # explicit provider type; overrides class fallback
 network: auto            # auto | tailscale | public
 hostId: h-0123456789abcdef0 # brokered use requires admin authentication
+workRoot: /srv/crabbox   # portable base root; not an exact command PWD
 
 lease:
   idleTimeout: 30m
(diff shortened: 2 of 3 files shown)


steipete:
Maintainer decision: retain the closing reference for https://github.com/openclaw/crabbox/issues/1269.

Crabbox intentionally defines the portable workspace override as a **base root**, not an exact remote command directory. Ordinary SSH-backed runs derive the lease/repository workspace under that root; provider adapters and Actions hydration remain authoritative where they own workspace placement. A generic `--workdir` flag would imply an exact `chdir` contract that Crabbox does not want to promise across those ownership boundaries.

This PR documents the supported product direction, so the broader exact-workdir proposal is being resolved as not wanted rather than left open for a second, conflicting interface.

@clawsweeper re-review

PR #1370, opened 2026-08-16
fix(lambda-microvm): bound runner response headers

SebTardif:
## What Problem This Solves

The Lambda MicroVM runner fallback previously had no response-header deadline. A stalled runner endpoint could therefore wait forever. The original 60-second `http.Client.Timeout` proposal bounded that wait, but it also bounded uploads and streamed command bodies, breaking valid long-running work.

This rewrite scopes the 60-second bound to response headers only.

## Correctness and Security Boundary

- A nil source uses a clone of the default transport with `ResponseHeaderTimeout` set to 60 seconds and `http.Client.Timeout` left at zero.
- Uploads and response streams remain bounded by the caller context after headers arrive.
- An injected `http.Client` is shallow-cloned rather than mutated. Its transport, proxy/TLS settings, cookie jar, timeout, and other fields remain intact.
- The clone wraps the injected redirect callback. Cross-origin scheme or host changes are rejected before caller policy; same-origin redirects delegate to a stricter caller callback unchanged.
- A nil injected redirect callback retains Go's normal ten-redirect limit for same-origin redirects.
- `X-aws-proxy-auth` and `X-aws-proxy-port` never reach a cross-origin redirect target. Go does not classify these custom headers as sensitive, so the runner client must enforce this boundary itself.

## Evidence

Focused security, timeout, upload, and stream coverage under the race detector:

```console
$ go test -race ./internal/providers/awslambdamicrovm -count=1 -run 'Test(NewRunnerClient|DefaultRunnerHTTPClient|RunnerClientCrossOriginRedirect)' -v
PASS
ok  github.com/openclaw/crabbox/internal/providers/awslambdamicrovm  2.175s
```

Full provider package under the race detector:

```console
$ go test -race ./internal/providers/awslambdamicrovm -count=1
ok  github.com/openclaw/crabbox/internal/providers/awslambdamicrovm  8.592s
```

Static analysis:

```console
$ go vet ./internal/providers/awslambdamicrovm
```

The same package race suite passed on a clean hydrated Blacksmith Testbox:

```console
$ go test -race ./internal/providers/awslambdamicrovm -count=1
ok  github.com/openclaw/crabbox/internal/providers/awslambdamicrovm  1.493s
```

Run: https://github.com/openclaw/crabbox/actions/runs/31994992845

The loopback redirect regression test exercises Go's real redirect machinery: a simulated runner response points at a second HTTP server, the permissive injected callback is not called, and the target observes zero requests.

## Review

A P1-focused structured autoreview of the exact branch diff found no P1 security or correctness issue. It raised one P2 concern about replacing Go's process-global `http.DefaultTransport` with a non-`*http.Transport` implementation. That was reviewed and not accepted: the requested fallback contract explicitly clones the standard default transport, and Crabbox uses the same typed-clone convention across its provider clients. Supporting arbitrary replacement round-trippers while timing only response headers would require a separate adapter and is outside this fix.

## Not Tested

No live AWS Lambda MicroVM control-plane or runner endpoint was available in this checkout. The network behavior is covered with loopback HTTP servers locally and on the clean Testbox.

## Credit

The rewritten commit preserves Sebastien Tardif's contribution with a `Co-authored-by` trailer, and the changelog thanks `@SebTardif`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ### Fixed
 
+- Bounded Lambda MicroVM runner response-header waits without limiting uploads or streamed executions, while preserving injected HTTP clients. Thanks @SebTardif.
 - Rejected native Jujutsu and other unsupported local sync sources before delegated archive providers can provision or execute a remote sandbox.
 - Accepted explicit local-container architecture assertions only when the selected Docker or Podman daemon reports a matching native architecture, without enabling emulation. Thanks @coygeek.
 - Omitted coordinator-only history commands from failure digests when run history is unavailable, while preserving direct lease recovery guidance.
--- a/internal/providers/awslambdamicrovm/backend_test.go
+++ b/internal/providers/awslambdamicrovm/backend_test.go
@@ -3,16 +3,22 @@ package awslambdamicrovm
 import (
 	"bytes"
 	"context"
+	"crypto/tls"
 	"encoding/json"
 	"errors"
 	"flag"
+	"fmt"
 	"io"
 	"net/http"
+	"net/http/cookiejar"
+	"net/http/httptest"
+	"net/url"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"slices"
 	"strings"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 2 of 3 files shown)


steipete:
Rewritten and force-pushed with an exact lease from `06f42e57` to `58534d83`. The PR body now documents the corrected response-header-only timeout, injected-client redirect guard, credential non-forwarding proof, focused and clean-runner race tests, review result, and the live-AWS limitation.

@clawsweeper re-review

PR #1380, opened 2026-08-17
fix(claims): bound inventory file reads

### LINKED ISSUE #1373: fix(claims): bound malformed claim file reads
## Summary

`crabbox claims list` reads each regular `.json` claim file completely before it decides that the record is malformed. A single oversized local claim therefore makes memory use grow with the file size, even though the command emits only a small bounded problem entry.

On exact current `main`, a sparse 128 MiB malformed claim increased maximum resident memory from 48,840,704 bytes for an empty claims directory to 330,842,112 bytes. The command eventually returned the documented partial JSON and exit code 2, but only after allocating enough memory to decode the entire file.

This is an ordinary local-state reliability bug, not an access-control or remote-input report.

## Steps to reproduce

1. Fetch and build the exact live default-branch revision:

   ```sh
   git fetch origin main --tags
   git checkout 3251e056596fe5368e7aa718feda7e9f4d5cb926
   go build -trimpath -o /tmp/crabbox-main ./cmd/crabbox
   ```

2. Create isolated state containing one small valid claim and one sparse 128 MiB malformed claim:

   ```sh
   state_root="$(mktemp -d)"
   mkdir -p "$state_root/crabbox/claims"
   printf '%s\n' '{"leaseID":"cbx_valid","provider":"local-container","repoRoot":"/tmp/repro"}' \
     > "$state_root/crabbox/claims/cbx_valid.json"
   truncate -s 134217728 "$state_root/crabbox/claims/cbx_oversized.json"
   ```

3. Run the local inventory under the macOS resource meter:

   ```sh
   XDG_STATE_HOME="$state_root" /usr/bin/time -l \
     /tmp/crabbox-main claims list --json
   ```

4. For a baseline, repeat the command with an isolated empty claims directory.

## Expected behavior

The inventory should reject a claim file above a small, documented implementation limit without reading or allocating memory proportional to its full size. It should preserve valid partial output, add a bounded structured problem for the oversized record, and return exit code 2.

The closure signal is that the 128 MiB fixture produces the same valid claim plus a structured problem while peak memory remains within a small fixed envelope over the empty-state baseline. Increasing the malformed file size should not increase process memory proportionally.

## Actual behavior

The command returns the valid record and a compact error as intended:

```json
{"version":1,"source":"local-claims","claims":[{"leaseId":"cbx_valid","slug":"","provider":"local-container","repoRoot":"/tmp/repro","pond":"","targetOS":"","windowsMode":"","claimedAt":"","lastUsedAt":"","idleTimeoutSeconds":0}],"problems":[{"file":"cbx_oversized.json","code":"invalid_json","message":"claim file is not valid JSON"}]}
```

It exits 2, but `/usr/bin/time -l` reports:

```text
330842112  maximum resident set size
```

The same exact binary against an empty isolated claims directory reports:

```text
48840704  maximum resident set size
```

That is approximately 6.8 times the baseline RSS for one 128 MiB malformed sparse file. Larger regular files remain unbounded by the current reader.

## Affected area

- `crabbox claims list` in both human and JSON modes, because both use the same read-only local-claim snapshot.
- Malformed or corrupted regular `.json` files under the local Crabbox claims directory.
- Valid partial results are delayed and can be lost if the process is killed for memory pressure before decoding completes.

## Runtime or environment

- Repository: `openclaw/crabbox`
- Latest release baseline: [`v0.43.0`](https://github.com/openclaw/crabbox/releases/tag/v0.43.0), source commit `7e03cb13a9ff89c894d2960e2c2c6024f2fa7b48`
- Live default branch: `main`
- Exact reproduced current-main revision: [`3251e056596fe5368e7aa718feda7e9f4d5cb926`](https://github.com/openclaw/crabbox/commit/3251e056596fe5368e7aa718feda7e9f4d5cb926)
- Go toolchain: `go1.26.5 darwin/arm64`
- Host: macOS, arm64
- Provider credentials: none; the reproduction uses only isolated local state

## Evidence

The read-only snapshot validates that the path is a stable regular file, then calls [`io.ReadAll(file)`](https://github.com/openclaw/crabbox/blob/3251e056596fe5368e7aa718feda7e9f4d5cb926/internal/cli/claim.go#L1969-L2004) before JSON validation. There is no per-file byte limit in that path. The separate `maxLocalClaimProblems` and diagnostic-filename limits bound emitted problem records, not the bytes read from each claim.

This path was introduced by [pull request #1364](https://github.com/openclaw/crabbox/pull/1364), which completed the local claim-inventory request in [issue #1250](https://github.com/openclaw/crabbox/issues/1250). Neither item covers a maximum claim-file size or the proportional-memory behavior above.

The complete range from the `v0.43.0` source commit through the exact reproduced `main` revision was inspected, together with the complete live open-issue and open-pull-request corpora and relevant closed issues and merged pull requests. No duplicate, active fix, or superseding report was found for oversized local claim reads.

Focused exact-main tests for claims, heartbeat, delegated archive sync, Windows workspace-owner transport, failure-digest routing, and local-container architecture assertions passed. The oversized-file result above was reproduced with the exact source-built binary.

## Impact

Local claim inventory is most useful when state is stale or malformed. A corrupted, accidentally enlarged, or otherwise abnormal claim file can instead make the diagnostic command consume hundreds of megabytes or more, cause severe memory pressure, or be terminated before it reports the other valid claims. Because the read is unbounded, the worst case is determined by available file size and process memory rather than by the small claim schema.

## Additional context

A size check should preserve the current read-only and race-safe invariants: inspect the opened regular file, reject an oversized record through a stable structured problem, and keep the existing post-read identity checks for accepted files. The exact byte threshold and problem-code spelling are product decisions; the required invariant is that a local claim cannot make inventory memory scale with arbitrary file size.

steipete:
## Summary

- cap `crabbox claims list` at an inclusive 1 MiB per claim file
- reject known oversized files before allocation and bound concurrent growth with a limit-plus-one reader
- emit a stable `claim_too_large` problem while preserving valid partial output and exit code 2
- leave runtime and mutating claim readers unchanged

Closes https://github.com/openclaw/crabbox/issues/1373

## Verification

- exact 1 MiB acceptance and 1 MiB + 1 rejection tests
- malformed, valid, concurrent-growth, deterministic-output, and lock-free inventory race tests
- source-blind RSS proof: empty 49,381,376 B; 128 MiB sparse file 49,397,760 B; 1 GiB sparse file 49,414,144 B
- `go vet ./...`, docs validation, `gofmt`, and `git diff --check`
- P1 autoreview clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ### Fixed
 
+- Bounded `crabbox claims list` inventory reads to 1 MiB per local claim while preserving valid partial output for oversized files. Thanks @coygeek.
 - Made local-container heartbeat authorize recorded dynamic runtime scopes and durably compare-and-swap exact claim lifecycle state without recreating or overwriting changed claims. Thanks @coygeek.
 - Made AWS image deletion resume from owner-level durable snapshot claims after AMI deregistration or catalog cleanup failures, preventing stale ordinary and capability-variant records from remaining selectable.
 - Made static SSH heartbeats persist touched timestamps and explicit idle-timeout replacements across fresh CLI processes, while omitted overrides preserve the stored timeout. Thanks @coygeek.
--- a/docs/commands/claims.md
+++ b/docs/commands/claims.md
@@ -50,12 +50,17 @@ details, labels, login and registration URLs or IDs, credentials, cache
 metadata, revisions, and fixed-create internals are never included. Claims are
 sorted by lease ID, provider, and slug.
 
-Malformed files do not hide valid claims. Each problem has stable `file`, `code`,
-and `message` fields. Supported codes are `invalid_filename`, `invalid_json`,
-`empty_lease_id`, `lease_id_mismatch`, `read_error`, and `invalid_claim`. Unsafe
-or overlong filenames are represented by a short SHA-256 fingerprint instead of
-their contents. Non-regular claim paths use `non_regular_file`. At most 100
-problem entries are emitted; the final
+Malformed files do not hide valid claims. For this read-only inventory, each
+local claim file has an inclusive 1 MiB implementation limit. Files that exceed
+the limit while being read use `claim_too_large` and are never read into memory
+in full; other concurrent changes remain `read_error`. This inventory limit does
+not change the runtime claim loader. Each problem has stable `file`, `code`, and
+`message` fields. Supported codes are
+`invalid_filename`, `invalid_json`, `claim_too_large`, `empty_lease_id`,
+`lease_id_mismatch`, `read_error`, and `invalid_claim`. Unsafe or overlong
+filenames are represented by a short SHA-256 fingerprint instead of their
+contents. Non-regular claim paths use `non_regular_file`. At most 100 problem
+entries are emitted; the final
 `problems_truncated` entry reports when additional files were omitted. Problems
 are sorted by file reference and code.
 
--- a/internal/cli/claim.go
+++ b/internal/cli/claim.go
@@ -76,6 +76,8 @@ type FixedCreateIntent struct {
 
 const FixedAWSClaimProvider = "aws-fixed-v1"
 
+const maxLocalClaimInventoryFileBytes int64 = 1 * 1024 * 1024
+
 var claimMutationMutexes sync.Map
 
 type invalidLeaseClaimIDError struct{ id string }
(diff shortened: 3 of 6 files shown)


steipete:
Maintainer decision: adopt the inclusive 1 MiB per-file limit and stable `claim_too_large` problem code for the read-only `crabbox claims list` inventory contract. Runtime and mutating claim readers remain unchanged; bounding diagnostic memory use is the correct tradeoff for oversized local records. The Unreleased changelog entry remains because repository policy requires coverage for user-visible changes.

PR #1387, opened 2026-08-17
feat(preflight): add raw-socket capability probe

### LINKED ISSUE #1258: feat(preflight): add an opt-in raw-socket capability probe
## Summary

Target repository: `openclaw/crabbox`

Add an explicit, opt-in raw-socket capability probe to `crabbox run --preflight`, so packet-oriented test jobs can determine whether the remote execution user may open a raw socket before the user command starts.

## Problem to solve

The current preflight is described as a target-specific capability snapshot, but it cannot report whether the execution user has the privilege required by raw-socket workloads. Its configurable registry covers command availability and a few OS-specific probes. On POSIX targets, the existing `sudo` probe only reports whether passwordless `sudo` is usable; that does not establish whether the command can open raw sockets without escalation.

This surfaced while validating a Scapy skill on Crabbox 0.40.0 with the `local-container` provider on Ubuntu 26.04 AArch64:

1. `crabbox run --preflight` reported `user=crabbox` and `sudo=yes`, but no raw-socket status.
2. The unprivileged workload reached `socket(AF_INET, SOCK_RAW, IPPROTO_RAW)` and failed with `PermissionError: [Errno 1] Operation not permitted`.
3. The same bounded localhost packet proof succeeded when explicitly run through passwordless `sudo`.

The user therefore learns about the missing direct-execution capability only after provisioning, synchronization, environment setup, and workload startup. `--preflight-tools` cannot currently express this requirement: `raw_socket`, `raw-socket`, and equivalent names are not registered probes.

## Proposed behavior

- Provide one documented, opt-in preflight probe for opening a raw socket on supported POSIX targets.
- Run the direct-execution check as the same remote user and from the same workdir/environment that the user command will use.
- Report a stable distinction between:
  - available to the execution user;
  - unavailable directly but available through Crabbox's already-detected non-interactive `sudo` path; and
  - unavailable.
- Keep the probe diagnostic only, consistent with current preflight behavior. Requesting it must not silently run the workload as root or alter the target's capabilities.
- Make the probe bounded and side-effect free: opening and closing the minimum socket needed to establish permission must not transmit a packet.
- Skip unsupported target types through the existing target-filtering behavior.

The exact probe name and implementation are intentionally left to the maintainers; the user-visible requirement is an explicit raw-socket capability result before command execution.

## Acceptance criteria

- A user can explicitly request the raw-socket probe through the same per-run and repository configuration surfaces used by existing preflight probes.
- On Linux and WSL2, a user with direct raw-socket permission is reported as directly capable.
- On Linux and WSL2, a user without direct permission but with a verified non-interactive escalation path is reported distinctly from a user with no usable path.
- The probe does not send packets, persist a socket, grant capabilities, modify the image, or implicitly elevate the user command.
- Unsupported target types are skipped consistently rather than producing a misleading capability result.
- Existing default preflight output remains unchanged unless the new probe is requested.
- Automated tests cover name validation, target filtering, all supported result states, and the no-packet/no-implicit-escalation contract.
- Documentation explains the probe's scope and clarifies that raw-socket permission is separate from interpreter or packet-capture-tool availability.

## Affected area

- `internal/cli/run_observability.go`: preflight registration, target filtering, and POSIX probe generation.
- `internal/cli/run_test.go`: capability-preflight generation and result-state coverage.
- `docs/commands/run.md` and `docs/observability.md`: supported opt-in probes and semantics.

## Non-goals

- Granting `CAP_NET_RAW`, changing container or VM configuration, or installing packet tools.
- Automatically running packet workloads through `sudo`.
- Sending test traffic or checking reachability to any network target.
- Detecting libpcap/BPF capture permissions, which are a separate capability from opening the raw socket observed here.
- Adding the probe to the default preflight set.

## Alternatives considered

- **Always run packet jobs with `sudo`:** this changes workload semantics and hides whether least-privileged execution is possible.
- **Treat `sudo=yes` as sufficient:** passwordless escalation availability does not prove direct raw-socket permission and does not say whether escalation is necessary.
- **Check for Python, Scapy, tcpdump, or another executable:** tool presence does not establish the kernel permission needed to open a raw socket.
- **Discover the requirement inside the workload:** this preserves the current late failure after the remote environment has already been prepared.

## Supporting context

The behavior above was reproduced with Crabbox 0.40.0. Current `main` was inspected at `ce9487f552054326d9209d3ba9606d01a36f5666`: `remoteCapabilityPreflightCommand`, `preflightToolRegistry`, and `posixPreflightProbe` contain no raw-socket capability probe.

Open issue #1243 is related only at the preflight-registry layer: it requests `python` and `python3` interpreter availability probes. It does not cover execution privileges or raw sockets, so this request has a separate closure signal.

steipete:
## Summary

- add the opt-in `raw_socket` preflight capability for Linux and WSL2
- report stable `direct`, `sudo`, `unavailable`, or `probe_missing` states using `python3` then `python`
- keep the probe diagnostic-only: it opens and immediately closes one raw socket, never binds, connects, sends, receives, installs software, grants capabilities, or elevates the workload
- isolate interpreter imports, suppress bytecode writes, reject ambiguous protocol output, and run sudo fallback only through trusted executable locations with a fixed root-side interpreter path

## Verification

- `go test -race ./internal/cli -run 'TestRawSocket|TestPreflight|TestPythonPreflight|TestRemoteCapabilityPreflight' -count=1 -timeout=10m`
- `go vet ./internal/cli`
- `git diff --check origin/main...HEAD`
- clean P1 maintainer autoreview after addressing portability, protocol-noise, bytecode, and multi-sudo-candidate findings

A broad local CLI race run encountered unrelated host-sensitive controller/listener timing failures and timed out under concurrent test load; every reported failing test passed when rerun independently. Exact-head CI is the authoritative broad gate.

## Exact-head live proof

Built from exact head `3d44f9c95428b16ab377c6ca92a2a771609d3d9b`:

```text
binary_sha256=72874b482cfc66399109ae5668b38cf56f90c5d845aa84eb620a615437f635df
provider=local-container runtime=docker server=29.4.0 image=python:3.13-slim

$ crabbox run --provider local-container --local-container-runtime docker \
    --local-container-image python:3.13-slim --no-sync \
    --preflight --preflight-tools raw_socket -- true
provisioning provider=local-container lease=<redacted> keep=false
provisioned lease=<redacted> container=<redacted> state=ready
remote preflight workspace=raw hydrate_supported=true
remote preflight raw_socket=sudo
running on 127.0.0.1 true
command complete exit=0
releasing <redacted>
lease cleanup stopped=true policy=auto lease=<redacted>

$ docker ps -a --filter label=com.crabbox.lease=<redacted>
<no output>
$ crabbox claims list --json | jq '[.claims[] | select(.leaseId == "<redacted>")] | length'
0
```

The execution user could not open the raw socket directly; the identical bounded probe succeeded through non-interactive sudo. The ordinary workload remained unprivileged, the user command completed normally, and cleanup left no proof-owned container or claim. Separate Docker controls observed permission denial without `NET_RAW`, direct success with `NET_RAW`, and sudo-assisted success from an unprivileged user.

Maintainer decision: retain the Unreleased changelog entry. Repository policy requires changelog coverage for user-visible features; release preparation consumes that maintained section rather than replacing normal feature entries.

Closes https://github.com/openclaw/crabbox/issues/1258

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
+- Added an opt-in Linux/WSL2 `raw_socket` preflight probe that distinguishes direct, non-interactive-sudo, unavailable, and missing-interpreter states without sending packets or elevating workloads. Thanks @coygeek.
 - Added checkpoint last-use tracking and composable `checkpoint prune --unused-for` cleanup for inactive local records and provider artifacts.
 - Added provider-native create, verify, delete, and fork lifecycle for direct Hetzner project-snapshot checkpoints, including exact local-claim image deletion.
 - Added `crabbox heartbeat` so external SSH drivers can refresh owned lease idle deadlines and optionally update the idle timeout.
--- a/docs/commands/run.md
+++ b/docs/commands/run.md
@@ -401,8 +401,9 @@ or the command/script you run.
 By default it probes common language and infrastructure tools plus OS-specific
 basics. Default generic probes are `git`, `tar`, `node`, `npm`, `corepack`,
 `pnpm`, `yarn`, `bun`, and `docker`; `uv`, `python`, and `python3` are available
-as additional opt-in built-ins. POSIX/Linux/WSL probes also include `sudo`,
-`apt`, and `bubblewrap`; native Windows probes include `powershell`,
+as additional opt-in built-ins. Linux and WSL2 also support the opt-in
+`raw_socket` capability probe. POSIX/Linux/WSL probes include `sudo`, `apt`,
+and `bubblewrap`; native Windows probes include `powershell`,
 `execution_policy`, `longpaths`, `temp`, and `pwsh`.
 
 Use `--preflight-tools` to replace the default tool list for one run:
@@ -411,6 +412,7 @@ Use `--preflight-tools` to replace the default tool list for one run:
 crabbox run --preflight --preflight-tools node,bun,docker -- bun test
 crabbox run --preflight --preflight-tools default,uv -- node --test
 crabbox run --preflight --preflight-tools python,python3 -- python3 -m pytest
+crabbox run --preflight --preflight-tools raw_socket -- ./packet-tests
 crabbox run --preflight --preflight-tools none -- ./smoke.sh
 ```
 
@@ -422,13 +424,21 @@ including `python` and `python3` on native Windows; Crabbox does not map either
 name to `py`. An unavailable literal command prints `<name>=missing` and the run
 continues.
 
+`raw_socket` uses `python3`, then `python`, to open and immediately close
+`socket(AF_INET, SOCK_RAW, IPPROTO_RAW)` without binding, connecting, sending,
+or receiving. Its stable result is `direct`, `sudo`, `unavailable`, or
+`probe_missing`. `sudo` means only that the same bounded probe succeeded through
+non-interactive `sudo -n`; Crabbox never elevates the workload, grants
+capabilities, installs software, or changes an image or container. The probe is
+separate from Python, Scapy, tcpdump, libpcap, and packet-capture availability.
+Unsupported targets skip it through normal preflight target filtering.
+
 Configure the default per repo:
 
 ```yaml
 run:
   preflightTools:
-    - python
-    - python3
+    - raw_socket
 ```
 
 ## Profiles, presets, and proof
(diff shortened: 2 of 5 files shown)


steipete:
@clawsweeper re-review

steipete:
Maintainer decision: accept the opt-in `raw_socket` preflight boundary and its stable `direct`, `sudo`, `unavailable`, and `probe_missing` states. The `sudo` path runs only the fixed, bytecode-free socket-open/close probe through `sudo -n`; it never elevates the workload or performs packet I/O. The Unreleased changelog entry remains under repository policy for user-visible features.

PR #1403, opened 2026-08-19
feat(providers): add tiny and small machine classes

steipete:
## Summary

The class ladder started at `standard`, which is a 32-vCPU machine on AWS, Azure and GCP — heavy
and expensive for smoke checks or small repos. This adds two classes below it.

| class | aws | azure | gcp | hetzner | namespace-instance |
|---|---|---|---|---|---|
| **tiny** | `m7a.large` 2/8 | `Standard_D2ads_v6` 2/8 | `c4-standard-4` 4/15 | `ccx13` 2/8 | `1x2` |
| **small** | `c7a.2xlarge` 8/16 | `Standard_D8ads_v6` 8/32 | `c4-standard-8` 8/30 | `ccx23` 4/16 | `2x4` |
| standard *(unchanged)* | `c7a.8xlarge` 32/64 | `Standard_D32ads_v6` 32/128 | `c4-standard-32` 32/120 | `ccx33` 8/32 | `4x8` |

`standard`, `fast`, `large` and `beast` keep their exact mappings and fallback ordering. The
default class stays `beast`.

## Notes on the sizing

- **GCP stays at 4 vCPU minimum.** `c4-standard-2` is 7.5 GB, and C4 standard is 3.75 GB/vCPU, so
  smaller sizes carry fractional GB. Sizes are chosen so memory lands on whole gigabytes.
- **Providers without a distinct smaller shape reuse their existing smallest type** rather than
  naming one that does not exist: Phala maps `tiny`/`small` to `tdx.small`, Namespace Devbox to
  `S`, and Cloudflare Containers keeps its single `standard-4`.
- **Every class switch was updated, not just the primary one.** Each switch ends in a `default:`
  that returns the class string as a literal provider type, so a partially-covered class would try
  to launch an instance type named "tiny" and fail at the provider. Coverage includes AWS
  amd64/ARM64/Windows/WSL2, Azure amd64/ARM64/Windows, GCP, Hetzner, both Namespace variants,
  Cloudflare, Phala, and the AWS quota recommendation path.

Vendor sources for the ten new sizes:
- AWS [M7a](https://aws.amazon.com/ec2/instance-types/m7a/) (4 GiB/vCPU) and
  [C7a](https://aws.amazon.com/ec2/instance-types/c7a/) (2 GiB/vCPU)
- Azure [Dadsv6](https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/general-purpose/dadsv6-series) (4 GiB/vCPU)
- GCP [C4 machine types](https://docs.cloud.google.com/compute/docs/machine-resource) (c4-standard = 3.75 GB/vCPU)
- Hetzner [dedicated vCPU](https://www.hetzner.com/cloud/dedicated-vcpu/) CCX line (4 GB/vCPU)

## Verification

```
gofmt -l $(git ls-files '*.go')          # clean
go vet ./...                              # clean
go test ./internal/cli/... ./internal/providers/{aws,azure,gcp,hetzner,namespaceinstance,phala}
                                          # all ok (internal/cli 711s)
bash scripts/check-docs.sh                # passed, no generated drift
```

`internal/providers/freestyle` has a timing-bound test that can fail on a loaded machine
(`TestFreestyleSyncHonorsConfiguredTimeout`). It passes on clean `main` in under a second and this
branch touches no files in that package.

Note `internal/cli` needs ~12 minutes here, so a full `go test ./...` can exceed Go's 10-minute
default; pass `-timeout 40m` when running the whole suite locally.

## Config / secret implications

None. Class names are additive; no existing mapping, default, or fallback order changes. Nothing
contacts a provider API at report time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
+- Added `tiny` and `small` machine classes for lower-cost smoke checks and small repositories.
 - Added an opt-in `cmake --version` preflight probe for POSIX, WSL2, and native Windows targets. Thanks @coygeek.
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -411,53 +411,71 @@ The providers below fall back across ordered instance-type lists unless `--type`
 pins a specific provider-native size.
 
 ```text
-Hetzner    standard  ccx33, cpx62, cx53
+Hetzner    tiny      ccx13, cpx22, cx23
+           small     ccx23, cpx32, cx33
+           standard  ccx33, cpx62, cx53
            fast      ccx43, cpx62, cx53
            large     ccx53, ccx43, cpx62, cx53
            beast     ccx63, ccx53, ccx43, cpx62, cx53
 
-AWS Linux  standard  c7a/c7i/m7a/m7i.8xlarge family
+AWS Linux  tiny      m7a.large, m7i.large, c7a.xlarge, c7i.xlarge
+           small     c7a.2xlarge, c7i.2xlarge, m7a.xlarge, m7i.xlarge, c7a.xlarge
+           standard  c7a/c7i/m7a/m7i.8xlarge family
            fast      …16xlarge family
            large     …24xlarge family
            beast     …48xlarge family, falling back to 32x/24x/16x
            arm64     c7g/m7g/r7g families with --arch arm64
 
-AWS Win    standard  m7i.large, m7a.large, t3.large
+AWS Win    tiny      m7a.large, m7i.large, t3.large
+           small     c7a.2xlarge, c7i.2xlarge, m7a.xlarge, m7i.xlarge, t3.xlarge
+           standard  m7i.large, m7a.large, t3.large
            fast      m7i.xlarge, m7a.xlarge, t3.xlarge
            large     m7i.2xlarge, m7a.2xlarge, t3.2xlarge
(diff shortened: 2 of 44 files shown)


steipete:
Landing with an admin override at the maintainer's explicit instruction, recorded here rather than
done silently. Branch protection reports `REVIEW_REQUIRED` and GitHub does not let an author
approve their own PR.

State at merge: 22 checks passing, 3 skipped, `mergeable: MERGEABLE`, head `f816d6bc`.

Rebasing onto the now-merged #1396 surfaced three things worth noting, all fixed here:

- Each `ClassSpecs()` implementation carried its own copy of the class order, so `tiny` and `small`
  existed in the type tables and worked as CLI flags but never reached `providers --json`. The
  order now lives in one exported `MachineClassOrder` that every provider iterates.
- Hetzner's shape table predated the new classes, so `ccx13` and `ccx23` resolved to a type with no
  vCPU or memory and would have rendered bare while every other provider showed numbers.
- Two tests still asserted exactly four classes. Both now assert against `MachineClassOrder`.

A new registered-provider test asserts every provider reporting classes covers the full order with
a complete shape; it fails naming `hetzner/tiny` and `hetzner/small` if that fix is reverted.

Local proof on this head: gofmt clean, `go vet` clean, `internal/cli` green (559s), all five touched
provider packages plus `internal/providers/all` green, docs gate passing with no generated drift.

PR #1407, opened 2026-08-19
feat: add experimental HTTPS Boxd provider

MichielMAnalytics:
## Summary

Add an experimental Boxd provider, rewritten around the HTTPS console API and authenticated WSS bootstrap instead of the vendor CLI or plaintext gRPC. The provider adapter owns authentication, immutable VM claims, isolation verification, SSH bootstrap, lifecycle operations, and recovery; core remains provider-neutral.

Interactive console sessions come from an explicit HTTPS device-login helper and are supplied through environment variables. API-key sessions cannot perform console mutations and are rejected rather than exchanged or silently substituted. Credentials are not stored in Crabbox configuration, placed on command lines, or sent over plaintext transport.

Ownership is bound to the API origin, organization, authenticated user, and immutable VM ID. Inventory must independently confirm isolation before any guest access. The intended guest path installs a per-lease key, obtains the host key over authenticated WSS, and uses a dedicated restricted SSH daemon. Cleanup verifies absence and retains recovery claims when the outcome is uncertain. Smoke cleanup is armed only after acquisition succeeds and targets the returned lease ID, preserving existing same-name leases.

## Accepted experimental limitation

The maintainer has explicitly accepted landing with this limitation documented; this is **not a claim of production guest readiness**.

On 2026-08-27, the production HTTPS API accepted `isolated: true` but inventory reported `isolated: false`. Direct creation, fork, and snapshot restore all exhibited this behavior. Crabbox therefore refuses guest access. There is no isolation bypass or plaintext fallback. `doctor` can succeed while `warmup` or `run` fails this check.

The real CLI run stopped before guest commands, SSH-key installation, port exposure, or repository sync. With `--keep=false`, it automatically deleted only the returned immutable VM ID. Independent inventory confirmed no new remaining VMs, preservation of the preexisting VM, and no local recovery claim. Additional probe VMs and the test snapshot were also removed with preexisting resources preserved.

Successful guest execution/reuse, organization semantics, terminal completion, raw forwarding, and guest sudo/sshd prerequisites remain unverified in production. Boxd also publicly proxies guest port 8000 without authentication, including for isolated VMs; the provider documentation warns against exposing private services there.

Public CLI/SDK source and examples did not supply a verified secure alternative: the documented production gRPC endpoint is plaintext, and credential-free probes found no working TLS gRPC listener at the tested production endpoints. The [vendor protobuf reference](https://docs.boxd.sh/reference/grpc-proto) supports isolation requests for creation/fork/restore, but those HTTPS paths did not honor the request in our live tests. See [the provider documentation](https://github.com/MichielMAnalytics/crabbox/blob/add-boxd-provider/docs/providers/boxd.md) for the full contract and verification limits.

## Verification

- TLS/WSS fake-server regression coverage for authentication, origin restrictions, ownership, isolation rejection, pinned SSH bootstrap, ambiguous creation, cleanup, reuse, and keep/stop semantics.
- Provider race tests, registration/configuration/CLI tests, provider vet, and targeted login/inventory/smoke script tests.
- Generated provider matrix, command docs, Markdown links, shell syntax, and diff checks.
- Independent Codex autoreview and source-blind CLI validation. Source-blind checks prove discovery and safe rejection; they do not establish successful production guest execution.
- Live HTTPS device authorization, identity/inventory reads, interactive creation, non-isolated rejection, and independent cleanup verification. Fork/restore and credential-free transport probes expanded the investigation; no credentials were sent over plaintext gRPC.

Contributor credit preserved: thanks @MichielMAnalytics.


MichielMAnalytics:
@clawsweeper re-review

The P1 local-claim release fence is fixed in `0176890` with regression tests, and the PR body now carries a redacted live transcript (create/use/destroy, fence refusal on an unclaimed convention-named machine, zero-residue check) against production boxd.sh.

MichielMAnalytics:
@clawsweeper re-review

Both P1s are fixed: `0176890` fences direct release behind the matching local claim, and `8e4f69d` makes resolve refuse unclaimed identities so a claim can never be minted for a machine this install did not create. The PR body transcript is regenerated at head `8e4f69d` against production boxd.sh and now demonstrates both refusals (exit 4, machine untouched) plus the full cycle and zero-residue check.

MichielMAnalytics:
@clawsweeper re-review

The machine-id binding P1 is fixed in `9a40daa`: every destructive path (resolve, release, cleanup) corroborates the claim's recorded immutable machine id against the live machine, so a stale claim can never touch a replacement that reused the name. The PR body transcript is regenerated at head `9a40daa` and includes a real name-reuse scenario against production boxd.sh: out-of-band destroy, name re-taken by a replacement, stale claim refused (exit 4), replacement untouched, stale claim reaped by cleanup.

MichielMAnalytics:
@clawsweeper re-review

The piped-installer P2 is fixed in `b9c5e3c`: the docs link the vendor's installation instructions and advise reviewing any installer script before executing it, instead of embedding `curl | sh`. `b9c5e3c` is documentation-only; the live transcript in the PR body was captured at `9a40daa` with an identical binary.

MichielMAnalytics:
@clawsweeper re-review

Both P1s are fixed and the branch is rebased onto current main (machine0 included, registry counts and help baseline reconciled). `2adec75`: acquire refuses an id-less create response and verifies the ready machine still carries the created id before any claim is persisted (identity change fails without destroying by name); delete-mode release holds the claim through the same 60s visibility grace acquire uses before treating not-found as provably gone. The PR body transcript is regenerated at head `2adec75` against production boxd.sh.

MichielMAnalytics:
@clawsweeper re-review

Both P1s are fixed in `2b3fbbf`: the acquire rollback is id-fenced (never removes by reusable name; verifies the live machine matches the created id, leaves replacements untouched, retains a recovery claim when identity is unverifiable), a blank ready-read id fails closed, and release/cleanup refuse unverifiable identities. The PR body transcript is regenerated at the current head `2b3fbbf` against production boxd.sh; the failure-injection paths (rollback, blank id) are pinned by regression tests since production always returns machine ids.

MichielMAnalytics:
@clawsweeper re-review

The unbound-claim P1 is fixed in `7fc422e`: resolve now fails closed on an unbound claim or blank live read before any start, SSH work, or claim refresh — same identity rule as release and cleanup — with a regression test pinning the refusal and the claim retained unchanged. The changelog hunk is dropped per the P3 (release flow owns CHANGELOG.md). The PR body transcript is regenerated at the current head `7fc422e` against production boxd.sh and shows the name-reuse release now reaping the stale claim with a truthful message while the replacement stays untouched.

MichielMAnalytics:
@clawsweeper re-review

The transient-cleanup-miss P1 is fixed in `49d9316`: the orphan pass corroborates absence with a direct read and reaps a claim only when the name stays absent through the same bounded visibility grace release uses, or when the live machine carries a different immutable id (definitive replacement); a matching or unverifiable id retains the claim. Regression tests pin transient-miss retention, sustained-absence reaping, and the definitive replacement case. The PR body transcript is regenerated at the current head `49d9316` against production boxd.sh; the transient-inventory edge itself is failure-injection-only and pinned by the tests.

MichielMAnalytics:
@clawsweeper re-review

Both findings from the last round were stale-merge-base artifacts: this PR has never touched `internal/cli/ssh_transport_session.go` or `internal/cli/provider_coordinator.go` — the review compared the branch against a main that had since landed #1409/#1410, making the branch appear to delete that new shared code. The branch is now rebased onto current main (`1ba708c`) with zero code delta in the rebase, full suites re-run green, and the PR body transcript regenerated at the current head against production boxd.sh.

MichielMAnalytics:
@clawsweeper re-review

The host-trust P1 is fixed in `4d9cd92`: boxd SSH targets now carry the vendor-managed known_hosts file (honoring a UserKnownHostsFile directive when the managed entry has one) and surface the CLI's exact `[host]:port` pin as the target host key, so the shared transport enforces `StrictHostKeyChecking=yes` — a mismatched edge key is rejected, never trusted on first use. Regression tests pin the pinned, pin-less, and exact-host-port matching cases. The PR body transcript is regenerated at the current head against production boxd.sh; every SSH connection in it ran under the enforced pin.

MichielMAnalytics:
@clawsweeper re-review

The pin is now REQUIRED, fixed in `94cc469`: a missing `[host]:port` pin in the vendor-managed known_hosts fails closed instead of falling back to accept-new — the resolve loop retries behind CLI re-syncs (which rewrite the pin) within the readiness window, then errors naming the missing pin, so a first public-edge connection is never trusted blind. The no-pin fallback test is replaced with fail-closed regression coverage (missing pin, absent known_hosts file), and the acquire path asserts the pinned target end to end. The PR body transcript is regenerated at the current head against production boxd.sh with the required pin active on every connection.

MichielMAnalytics:
@clawsweeper re-review

The post-check mutation race is fixed in `9e7e161`: every mutating vendor call — release destroy, stop, cleanup destroy, both rollback destroys, and resolve's start — is now addressed by the verified immutable machine id instead of the reusable name, so a replacement created between the identity read and the mutation can never be hit (an id-addressed mutation on a gone machine is simply not-found). Regression tests pin the id-addressed argv on each path. The PR body transcript is regenerated at the current head against production boxd.sh.

MichielMAnalytics:
@clawsweeper re-review

Both identity-handoff P1s are fixed in `201fbca`: acquire revalidates the immutable id after SSH endpoint resolution, and resolve compares the readiness wait's returned id against the claim binding instead of discarding it — a replacement racing into the name after the first identity check can neither pass readiness nor supply the SSH endpoint. Regression tests pin both swap races end to end with no mutation issued. The PR body transcript is regenerated at the current head against production boxd.sh.

MichielMAnalytics:
@clawsweeper re-review

The last finding is fixed in `a5ff65b`: the returned machine name is validated before any lifecycle work — on mismatch the machine is destroyed by its immutable id, or (with --keep, or no id) a recovery claim is retained under the RETURNED name, so a misnamed create can never end up live and unmanaged. Regression test pins the id-addressed destroy. Transcript regenerated at the current head against production boxd.sh. That clears every actionable finding raised across all review rounds.

MichielMAnalytics:
@clawsweeper re-review

The live-smoke trust-boundary P1 is fixed in `fa408dc`: the smoke reads `boxd.cli` through `trusted_config_value` (environment override or trusted config only), matching the Go adapter's gate, so a repository checkout can never select the binary that runs with the operator's boxd credentials. A focused regression in `live-smoke.test.js` pins the trusted source and rejects the untrusted one. `fa408dc` changes only the smoke script and its test — the provider binary is identical to the transcript's head `a5ff65b`.

steipete:
Maintainer review at head `fa408dc80d1b00b8f1316ad121b47bcf1045167d`: this is much closer, and the production create/use/destroy plus ownership-fence proof is substantial. I am not sponsoring or merging the built-in provider yet because one credential-execution boundary remains open.

The live-smoke fix still treats every explicit `CRABBOX_CONFIG` path as trusted. In `scripts/live-smoke.sh`, lines 31-37 assign `trusted_config_path` directly from the environment, including a relative path inside the checkout, an absolute path inside the checkout, or a symlink into it. `boxd_smoke` then executes `boxd.cli` from that file with the operator's stored boxd login or `BOXD_TOKEN`.

That differs from Crabbox's canonical contract in `internal/cli/config.go`: an explicit path is trusted only when it is outside the active repository, including after symlink resolution. `docs/features/configuration.md` documents the same boundary, and `TestExplicitConfigSymlinkIntoRepoRemainsUntrusted` pins it.

Please make the smoke's executable selection environment-only:

```bash
local boxd_cli="${CRABBOX_BOXD_CLI:-boxd}"
```

That is the smallest robust fix; porting the full Go repository-root and symlink classifier into shell would be a much larger duplicate policy. Replace the current regex-only test with behavior proving repository config and a symlinked explicit config cannot select the executable.

After that, please refresh onto current `main` (`e44c686cf8110facfcc84334a0c2e735b5640bd2`) and adopt its new provider conventions before the final proof:

- use `shared.Poll` for the repeated lifecycle reads instead of the four hand-written polling loops;
- expose the provider's hermetic lifecycle suite through a convention-named `*lifecycle*_test.go` file so `scripts/test-provider-lifecycles.sh` includes Boxd;
- rerun the complete local gates and the production create/use/destroy plus failure-cleanup/zero-residue proof with `CRABBOX_BOXD_CLI` set explicitly.

The project-level decision to maintain Boxd's CLI, SSH-config, credential, and host-key contracts remains separate. I will revisit sponsorship after the trust boundary, current-main integration, exact-head CI, and live destructive proof are all final.

MichielMAnalytics:
Thanks for the review — you are right, and the trusted-path assumption in the smoke was the real hole: `trusted_config_path` takes `CRABBOX_CONFIG` verbatim, so an explicit path or symlink into the checkout passed as trusted, unlike the Go classifier. All four points are addressed at head `7c5e3b6`:

- **Environment-only executable selection**: `local boxd_cli="${CRABBOX_BOXD_CLI:-boxd}"`, exactly as prescribed — no config file, trusted or not, can choose the binary. The regex test is replaced with two behavior tests: a repository `crabbox.yaml` nominating an in-repo binary, and an explicit `CRABBOX_CONFIG` symlinked into the checkout — both prove the nominated binary never executes and the PATH default is the probed one.
- **`shared.Poll`**: all four hand-written polling loops (machine readiness, ssh-endpoint resolution, release visibility grace, cleanup absence grace) now go through `shared.Poll`/`shared.SleepContext`, with the existing regression suite pinning identical grace/timeout semantics.
- **Lifecycle convention**: the hermetic suite (28 tests) moved to `backend_lifecycle_test.go`; `scripts/test-provider-lifecycles.sh` now reports 6 provider lifecycle packages including boxd, and the provider matrix records the coverage.
- **Gates + proof**: `go build`, `go vet`, `gofmt`, `go test -race` (provider, registry, `internal/cli`), `scripts/test-provider-lifecycles.sh`, `scripts/check-docs.sh`, and the live-smoke behavior tests are green on current main. The PR body transcript is regenerated at `7c5e3b6` against production boxd.sh with `CRABBOX_BOXD_CLI` set explicitly on every invocation: create/use/destroy, both ownership-fence refusals, the failure-cleanup name-reuse scenario (out-of-band destroy, name re-taken, stale claim reaped, replacement untouched), and zero residue.

On the ongoing support boundary: understood that it is a separate decision. We are committed to keeping the CLI lifecycle, ssh-config, credential, and host-key-pin contracts stable — the `--json` surface is a documented compatibility contract on our side — and we will maintain this provider.

MichielMAnalytics:
Hi @steipete let me know whether it's sufficient or that other changes are required to get it merged. Can't wait to get this going

steipete:
Live validation update (2026-08-27): I built and tested an HTTPS-only rewrite locally, but this provider is still blocked on production behavior.

The vendor's documented gRPC endpoint is cleartext; no credentials were sent to it. The HTTPS console works with an interactive device-login session. API-key sessions can read inventory but are rejected for console mutations.

The decisive failure is isolation: `POST /api/v1/vms` accepted `isolated: true`, but a separate inventory read reported the created VM as `isolated: false`. The create response also uses `vm_id`, while inventory uses `id`; the local rewrite and fixtures now handle those shapes separately. [Boxd documents isolation as creation-time and immutable](https://docs.boxd.sh/use-cases/sandboxes), so it cannot be repaired afterward.

I reproduced the failure with the rebuilt Crabbox CLI: `doctor --provider boxd --json` passed, and a fresh no-sync run exited 5 with `boxd machine is not isolated; refusing guest access`. The adapter stopped before guest bootstrap, key installation, port exposure, or repository sync, then deleted only the returned immutable VM. Independent inventory confirmed zero new VMs, preservation of the pre-existing VM, and zero remaining local claims.

The local rewrite removes vendor CLI dependence, uses HTTPS/WSS with no plaintext fallback, fences lifecycle actions to authenticated ownership and immutable IDs, and adds regression coverage for this rejection and cleanup. Focused provider race tests, vet, CLI/config/registration tests, eight auth/smoke tests, and docs/matrix checks pass. Those checks are not successful production end-to-end proof; guest execution and reuse remain unvalidated.

No rewrite commits have been pushed and this PR has not been merged. Landing needs a supported path that provides both encrypted control-plane transport and isolated creation. Silently disabling the isolation check would change the provider's safety contract rather than fix the integration. Thanks @MichielMAnalytics for the initial provider work.

steipete (approved):
Reviewed the HTTPS-only rewrite and its ownership/isolation boundaries. Independent Codex autoreview reports no actionable findings. The documented production limitation is accepted for this experimental provider: live tests prove rejection and cleanup of non-isolated VMs, not successful guest execution. No isolation bypass or plaintext fallback is introduced.

The rewritten head `59e1aaab8ec21f445a0d070c777ad48741cae84c` passed [full CI](https://github.com/openclaw/crabbox/actions/runs/33139401371), [release snapshot checks](https://github.com/openclaw/crabbox/actions/runs/33139401372), [connector E2E smokes](https://github.com/openclaw/crabbox/actions/runs/33139401395), and [docs visual proof](https://github.com/openclaw/crabbox/actions/runs/33139401376). All required workflows were approved explicitly for this fork revision; no failed tests needed rerunning.

Provider race, registration, configuration, claim-scope, and routing checks were also rerun against the combined tree with the latest provider-routing refactor on main. Contributor credit is preserved. Approved to land with the experimental caveats in the provider guide and changelog.

steipete:
Landed as [160ea4ba](https://github.com/openclaw/crabbox/commit/160ea4ba4990346e600768571bb9d354209bd09b), with contributor credit preserved. This supersedes the earlier hold: the production isolation limitation is explicitly accepted and documented for an experimental provider, without weakening the isolation check or introducing plaintext fallback.

The rewritten head passed [full CI](https://github.com/openclaw/crabbox/actions/runs/33139401371), [release snapshot checks](https://github.com/openclaw/crabbox/actions/runs/33139401372), [connector E2E smokes](https://github.com/openclaw/crabbox/actions/runs/33139401395), and [docs visual proof](https://github.com/openclaw/crabbox/actions/runs/33139401376). Focused provider race, registration, configuration, routing, and claim-scope checks also passed on the combined tree with the latest main refactor. No failed test reruns were needed.

The normal merge was held by the last-pusher approval rule, despite the recorded approval. The explicitly requested maintainer landing used the merge override after all CI passed; no branch-protection settings were changed.

Live proof remains deliberately narrow: HTTPS authentication, refusal of non-isolated production VMs before guest access, and immutable-ID cleanup with independent inventory confirmation. Guest execution/reuse is still unverified; all task-created VMs and the test snapshot were removed. See the [provider guide](https://github.com/openclaw/crabbox/blob/main/docs/providers/boxd.md) for the known limitations and public-port warning. Thanks @MichielMAnalytics.

MichielMAnalytics:
Thanks Peter @steipete

PR #1413, opened 2026-08-19
refactor(providers): share JSON subprocess exchange

steipete:
## Architecture and scope

This introduces a minimal shared internal/providers/shared/procjson transport for subprocess protocols that accept one JSON request and return exactly one JSON response. The helper owns request encoding, mandatory per-stream capture bounds, cancellation grace, the runner call, nonzero-exit detection, and strict single-document response decoding.

This PR migrates only:

- the CodeSandbox control bridge;
- the CUA control bridge; and
- External provider versioned protocol mode.

Provider adapters continue to own SDK setup, environment construction, operation timeouts, response envelopes, identity and version validation, redaction, and provider-specific error classification. No streaming, NDJSON, noisy CLI, legacy External mode, provider lifecycle, configuration, or protocol contract moves into the shared layer.

## Safety and behavior preservation

The shared exchange requires a positive capture bound and enabled output capture. It rechecks fake-runner results against that bound, treats runner errors and nonzero exits as command failures, and rejects empty, malformed, or trailing JSON. Shared failure strings never include captured subprocess output; adapters retain control of safe diagnostics and redaction.

CodeSandbox and CUA keep their existing setup, environment, timeout, envelope, and redaction behavior. External keeps its versioned request/response validation, error mapping, and newline-terminated request record for compatibility with line-oriented adapters. Production code remains 125 additions and 126 deletions: net -1 LOC.

## Verification

- go test -race ./internal/providers/external -run TestDoctorExecutesLineOrientedProviderAsChildProcess -count=1 -v
  - PASS: a real child-process External helper reads exactly one newline-delimited request through the production invoke path and returns a valid response.
- go test -race ./internal/providers/shared/procjson ./internal/providers/external ./internal/providers/codesandbox ./internal/providers/cua
- CommandRunner output-bound and cancellation coverage passed.
- Full provider and full Go race suites passed during implementation; a later unrelated coordinator timing flake passed alone.
- go vet ./...
- docs build and git diff --check
- fresh autoreview after the delimiter compatibility repair: clean, with no accepted/actionable findings

ClawSweeper's line-termination finding was accepted and fixed in a separate follow-up commit, with the real child-process regression above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- Bounded strict single-request provider JSON subprocess exchanges through a shared transport while keeping provider validation and diagnostics adapter-owned.
 - Enforced coordinator-provided SSH host keys before first transport and removed per-lease local SSH credentials after confirmed brokered release.
 - Shared lifecycle observation across Machine0, Tenki, DigitalOcean, Linode, Vultr, and Morph while keeping provider state handling adapter-owned, and made canceled Tenki readiness waits report cancellation instead of timeout.
 
--- a/docs/features/provider-authoring.md
+++ b/docs/features/provider-authoring.md
@@ -493,6 +493,10 @@ Rules:
 - Use `rt.Exec.Run(ctx, core.LocalCommandRequest{...})` for every subprocess.
   Never call `exec.CommandContext` directly. Tests pass a fake `CommandRunner`
   to assert on argv without spawning real processes.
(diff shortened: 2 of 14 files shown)


steipete:
@clawsweeper re-review

The External newline-termination finding is fixed on the new head. A real line-oriented child-process helper now verifies the production invoke path, and the focused race suite plus fresh autoreview are clean. The PR body contains the exact proof.

PR #1424, opened 2026-08-20
fix(provider): fence Vultr cleanup with exact claims

steipete:
## Summary

- fence Vultr cleanup with one exact revisioned claim snapshot as the transaction carrier
- persist ambiguous identities before locking, then revalidate provider account, instance, and key inside the claim CAS
- delete the instance before the exact key while retaining the claim and local key across partial failures

## Testing

- focused Vultr race tests
- `go test -timeout 20m ./...`
- `go vet ./...`
- `git diff --check`
- Codex autoreview

## Live proof

No live Vultr proof was run because no approved paid Vultr fixture is configured.


steipete:
The P1 is fixed: cleanup preparation is now observational, while actual deletion still durably persists recovered claim state immediately before provider mutation. Regression coverage exercises both ambiguous-create and ambiguous-key dry-run recovery, asserting byte-identical claims and zero provider mutations.

Proof is clean: focused race regressions, the full Vultr provider suite, the full repository race suite, `go vet ./...`, docs generation, diff checks, and an independent structured review. Live Vultr proof remains unavailable because no approved paid fixture is configured.

@clawsweeper re-review

PR #1444, opened 2026-08-20
feat: define versioned Linux runner readiness manifests

vincentkoc:
## What Problem This Solves

Crabbox's Go and Worker Linux bootstrap paths currently maintain separate package expectations, while prebaked images expose only a legacy marker that cannot establish whether the required tools still work. Image builders can therefore claim readiness without closing the producer/consumer contract, and otherwise reusable images can unexpectedly fall back to package installation.

## Why This Change Was Made

This rewrite adds a standalone, provider-neutral `crabbox-linux-readiness/v1` contract directly on `main`, independent from runner telemetry and typed ready pools.

The authoritative manifest is canonical JSON with exactly three fields:

```json
{"profile":"linux-minimal","recipeDigest":"sha256:<digest>","schema":"crabbox-linux-readiness/v1"}
```

- `linux-minimal` defines the exact baseline package and functional-probe contract shared by Go and Worker bootstrap.
- `linux-builder` is a generic verified superset. Its Python capability creates and checks an actual temporary pip-enabled virtual environment rather than merely importing `venv`.
- `recipeDigest` covers the deterministic executable capability contract. The original ambiguous `inventoryDigest` is intentionally omitted.
- Canonical recipes generate the Go bootstrap, Worker bootstrap, and standalone image producer. CI verifies generated freshness and runs both `.test.js` and `.test.mjs` script suites.
- The refactor removes duplicate package inventories, custom TypeScript serialization, redundant probe execution, repeated producer uploads, and unnecessary filesystem-wide flushes while preserving one final pre-capture proof.

The authoritative manifest lives under root-controlled `/var/lib/crabbox-readiness/linux.json`. Existing `/var/lib/crabbox` directories are intentionally runtime-user-owned, so `/var/lib/crabbox/image-ready` remains only a compatibility hint. Marker-only migration requires exact root-owned marker bytes and every minimal functional probe under a sanitized system path; an invalid authoritative manifest is never rescued by the marker. Cold bootstrap safely creates an absent legacy marker directory rather than assuming an image already contains it. Locale-sensitive filesystem metadata is parsed in the C locale without changing the locale seen by readiness probes.

Direct AWS Linux launches now gzip cloud-init before base64 encoding, matching the existing Worker path and enforcing EC2's 16 KiB raw user-data limit. Windows and macOS payloads retain their existing transport behavior.

## User Impact

[...]

PR #1452, opened 2026-08-21
feat: publish protected AWS runner image candidates

vincentkoc:
Stack dependency: https://github.com/openclaw/crabbox/pull/1444

## What Problem This Solves

AWS runner image creation lacked a protected, repeatable candidate pipeline for Linux and Windows. Operators could not produce reviewable image evidence without coupling candidate creation to promotion or changing the active default image.

## Why This Change Was Made

This adds a protected manual candidate-only workflow and versioned recipes for Linux and Windows AWS images. It mints an AMI, validates source and candidate boots, scrubs source identity and credentials, and emits an immutable `CandidateRecordV1` evidence bundle.

The hardened capture path prepares the source before credential scrubbing, then resolves the durable AWS lease identity without SSH, Tailscale, or login probes. `--source-prepared` is restricted to native AWS AMI capture and fails closed when the durable claim or instance ID is missing.

Linux and Windows scrub gates remove private SSH key families as well as authorized keys. Linked home/profile roots, unsafe profile links, dangling reparse points, and multi-hop compatibility links fail closed without traversing outside the image root. The two standard Windows compatibility links are accepted only when their literal targets and every destination component match trusted in-root directories.

Windows smoke commands fail on native nonzero exits and prove Windows Server 2022 build `20348` with Desktop Experience / Full installation. Candidate evidence also requires checkpoint architecture to be present and exactly match the requested architecture.

Workflow tenant, organization, and publisher email values come from the protected GitHub environment's variables or secrets and fail closed when missing or invalid.

The GHCR artifact contains evidence only: candidate record, recipe, SBOM, provenance, and scrub report. AMI and snapshot bytes remain in AWS, GHCR is never a boot dependency, tags are digest-derived and immutable, and the bundle has an explicit GitHub Actions keyless-signing contract.

This PR intentionally performs no promotion, Fast Snapshot Restore, default-image mutation, release, or automatic workflow dispatch.

## User Impact

Operators can build Linux or Windows AWS image candidates from an exact base AMI and receive structured, redacted evidence suitable for a separate review and promotion decision.

## Evidence

Local validation:

- 46 Node tests across candidate, publisher, workflow, mint, and scrub suites
- focused Go checkpoint and durable-lease tests
- `actionlint` plus Bash and Node syntax checks
- `gofmt`, `git diff --check`, and privacy scans
- final structured autoreview: clean

No live AWS, GHCR, promotion, default-image, or Fast Snapshot Restore action was run. PowerShell parser/runtime validation was unavailable locally because `pwsh` is not installed; Windows behavior is covered by fake-root model tests and source-contract assertions.


steipete:

PR #1453, opened 2026-08-21
feat: add opt-in Git overlay sync

vincentkoc:
## Summary

Adds default-off `sync.gitOverlay` support for eligible Linux SSH-backed workspaces. The runner is prepared from the exact advertised Git commit and complete filtered commit/tree history, then Crabbox transfers only the authoritative local dirty overlay while retaining the normal complete file and deletion manifests.

This is a maintainer rewrite directly on current `main`. It preserves Vincent Koc's contribution credit while removing the original draft's dependency on the unmerged https://github.com/openclaw/crabbox/pull/1443, https://github.com/openclaw/crabbox/pull/1444, and https://github.com/openclaw/crabbox/pull/1450 stack and leaving ready-pool branch/tag behavior unchanged.

## User-visible contract

- Enable with `sync.gitOverlay: true` or `CRABBOX_SYNC_GIT_OVERLAY=true`; the default remains `false`.
- Eligible runs require Linux SSH synchronization, Git seeding and deletion enabled, a complete conflict-free checkout without submodules or transforms, and an anonymous HTTP(S) or remotely readable filesystem origin that supports filtered history.
- Staged, unstaged, untracked, deleted, renamed, executable-bit, and symlink changes are preserved. A clean checkout transfers no source-file payload.
- The complete ordinary sync manifest remains authoritative, including excludes, reserved paths, transfer guardrails, and deletion safety.
- Git hooks, credential helpers, prompting, SSH agents, global/system configuration, URL rewrites, external transports, filters, and shared/linked worktree metadata are excluded from the overlay transaction.
- Trusted ignored dependency caches and real `.crabbox` runtime state may survive reuse; mutable `.git/info/exclude` rules and symlinked/escaping roots cannot authorize preservation.
- Unsupported or safely classified states keep the ordinary full-manifest path. Authentication-required anonymous HTTP(S) falls back without forwarding credentials; genuine eligible-origin DNS/TLS/network failures remain fatal.
- Timing JSON adds overlay mode, transferred-file/byte counts, and sanitized fallback reason only when overlay was requested.

## Live proof

The runtime implementation at `a897fa8175d8b2a2ba8c56f0f38970345bec7223` was exercised on a retained AWS Linux SSH lease (`m7a.large`), and the lease was released afterward. Final head `a5797027` differs only by removing the release-owned changelog section requested in review. Focused race and rollback coverage separately proves the fail-closed symlink boundaries and transactional Git-recovery paths.

| Case | Result |
|---|---|
[...]

PR #1455, opened 2026-08-21
feat: add AWS cache volume lifecycle

vincentkoc:
Stacked on https://github.com/openclaw/crabbox/pull/1453.

## What Problem This Solves

AWS Linux SSH leases could not safely reuse durable build caches across
workspaces. Repeated dependency downloads and rebuilds increased startup and
feedback latency, while ad hoc volume reuse risked cross-tenant disclosure,
availability-zone mismatches, unsafe mounts, or leaked attachments.

## Why This Change Was Made

This adds the AWS cache-volume lifecycle for both direct and brokered Linux SSH
leases. It reserves one encrypted `gp3` EBS member per logical cache member and
availability zone, binds reuse to the authenticated tenant, opaque repository
scope, cache key, and explicit ABI, and durably reconciles reservation,
attachment, quarantine, release, purge, and TTL garbage collection states.

The broker protocol requires cache-volume protocol v1 and resolved bindings.
Cache-bearing fixed create intent uses v3 and includes the opaque repository
scope, while no-cache fixed intent remains stable at v2. AWS tags contain only
opaque random identifiers, generation, and ABI digest.

The generated IAM policy scopes volume mutation to exact Crabbox-owned cache
volumes and attach/detach to existing Crabbox-owned instances. Bootstrap
resolves the exact NVMe serial, rejects unsafe or symlinked mount paths, verifies
the ABI sentinel, uses `nodev,nosuid`, and gates readiness. Release and garbage
collection require exact detach, ownership, property, and zero-attachment
proofs.

## User Impact

Operators can configure persistent AWS cache volumes for faster repeated Linux
SSH workloads in either direct or coordinator-backed mode. Required caches fail
closed on unsupported or old coordinator paths; optional caches may only be
ignored with explicit evidence.

## Evidence

- Focused Go fixed-intent, config, coordinator capability, IAM, and AWS cache
  lifecycle tests passed.
- Focused Worker cache, config, AWS adapter, coordinator runtime, and fixed
  intent tests passed: 151 lifecycle/runtime tests plus the v2/v3 identity
  regression.
- Worker formatting, lint, and TypeScript checks passed.
- Documentation generation and `git diff --check` passed.

No AWS mutation occurred while developing or validating this change.

Residual risk: validation used fake cloud adapters only. Live IAM evaluation,
EBS attach/detach behavior, and Nitro NVMe serial discovery still require a
controlled AWS canary before production rollout.


steipete:

PR #1456, opened 2026-08-21
feat(images): qualify AWS builder candidates

vincentkoc:
## What Problem This Solves

AWS Linux builder image candidates lacked a protected, reproducible qualification gate binding the immutable Q1 candidate evidence to live typed-pool, overlay, cache, cleanup, and publication proof.

This PR is stacked on https://github.com/openclaw/crabbox/pull/1454. Review and merge that dependency first.

## Why This Change Was Made

This adds a protected manual AWS builder qualification workflow, strict public receipt and Q1 input contracts, fail-closed live qualification gates, exact input-byte preservation, and a separately authorized signing and publication job. The qualification job has read-only package access; only the publication job receives package write and OIDC permissions.

## User Impact

Operators can qualify one immutable AWS Linux builder AMI and publish a signed, privacy-safe receipt without promoting the image or changing provider defaults.

The only configuration implication is the required `image-qualifier` GitHub environment and its documented scoped coordinator and AWS secrets. No secret values are added to the repository.

## Evidence

Focused local verification on exact head `8a7b51ecd2cfd2a8df11f8002ba7bc8d36478032`:

- `node --test scripts/aws-image-qualification.test.js scripts/qualify-aws-image-candidate.test.js scripts/publish-aws-image-qualification.test.js scripts/publish-aws-image-qualification.integration.test.js scripts/devtools-image-qualification-workflow.test.js`
  - 24 tests: 23 passed, 1 skipped.
  - The real ORAS/Cosign localhost integration was skipped locally because those binaries are absent; the workflow contract requires and CI covers that integration.
- `bash -n scripts/qualify-aws-image-candidate.sh scripts/publish-aws-image-qualification.sh`
  - Passed.
- `node --check scripts/aws-image-qualification.mjs`
  - Passed.
- `node --check scripts/consume-aws-image-candidate.mjs`
  - Passed.
- `actionlint .github/workflows/devtools-image-qualify.yml`
  - Passed.
- `go run ./scripts/validate-json-schema recipes/aws/v1/qualification-input.schema.json scripts/fixtures/aws-image-qualification-input.json`
  - Passed.
- `node scripts/consume-aws-image-candidate.mjs validate-qualification-input --file scripts/fixtures/aws-image-qualification-input.json`
  - Passed.
- `git diff --check fba465b477870c0a21c8ce79965bb5919f315c5d...HEAD`
  - Passed.
- Privacy scan and `git verify-commit HEAD~2 HEAD~1 HEAD`
[...]

PR #1458, opened 2026-08-21
feat(images): add protected image default promotion

vincentkoc:
## What Problem This Solves

Image-default promotion needs an auditable, race-safe path that cannot silently overwrite a concurrent default change or reuse unverified qualification output.

## Why This Change Was Made

- Adds provider-neutral versioned CAS contracts and current-default revision discovery, with atomic AWS and Azure default updates.
- Verifies immutable signed AWS/Linux qualification evidence, including source identity, normalized scope, owner account, and exact backing snapshots.
- Persists durable idempotent promotion and rollback attempts, authoritative outcomes, fresh-selection verification, and signable privacy-closed receipts.
- Adds a protected minimal-client promotion workflow with separated verify, mutate, probe, finalize, and signing permissions.
- Rejects protected legacy mutations and FSR/catalog-only options before provider side effects; protected promotion never enables FSR.

Stack dependency: Q2 qualification and canary workflow in https://github.com/openclaw/crabbox/pull/1456.

## User Impact

Operators can promote qualified AWS image defaults with revision-bound compare-and-swap semantics, recover safely from lost responses or expired claims, and publish signed receipts. Authorized clients can also use versioned CAS for unprotected provider-neutral defaults, currently Azure, without gaining legacy admin mutation access.

No live workflow dispatch, cloud mutation, GHCR publication, FSR action, or image-default mutation was performed during development. The required external `image-promoter` GitHub environment and its promotion-only token configuration were not created by this change.

## Evidence

- Worker image promotion CAS/protected-attempt suite: 27 passed.
- Node promotion, receipt, privacy, docs, and publisher suite: 33 passed.
- Exact-head workflow policy suite: 6 passed.
- Go CLI image/promotion tests passed.
- Worker typecheck and lint passed; changed files pass formatting.
- `actionlint`, diff checks, and verification of all five signed commits passed.


steipete:

PR #1459, opened 2026-08-21
chore(release): block v0.46.0 record and prepare v0.46.1

steipete:
Marks `release/records/v0.46.0.json` blocked (tag signed before its SSH signing key was registered on GitHub — verification is push-time-evaluated, so the tag is permanently unverified and can never pass publication; v0.37.0 precedent, tag preserved). Retitles the unpublished 0.46.0 changelog section to 0.46.1 and bumps `worker/package.json` + both lock entries. No Go changes since the fully green v0.46.0 gate battery and coordinator-backed live smoke on f832b8d5; quick gates (vet/build/script tests/diff-check) re-run green on this commit.


steipete:
Obsolete: reclaiming v0.46.0 by replacing the tag under temporarily lifted ruleset enforcement (explicit maintainer instruction).

PR #1463, opened 2026-08-21
test: make two timing-sensitive tests robust on loaded CI runners

steipete:
Both tests failed release-critical CI this week on busy shared runners while passing locally:

- **`TestReserveSSHLocalForwardPortIgnoresUDPUse`** (run 32436829695): asserted a fixed UDP-selected port's TCP twin was reservable; an unrelated runner process held it. Now retries across up to 32 candidate ports — the UDP-must-not-block-TCP invariant fails only if every candidate exhibits the guarded misbehavior.
- **opensandbox `TestRunReconcilesAmbiguousCreateAfterDelayedVisibility`** (run 32506126513): raced a wall-clock reconciliation window. Now drives polling through the existing cleanup-deadline seam with effectively immediate cadence, still requiring exactly two ownership lookups and one deletion. (The httptest warning in the CI log was an unrelated neighbor — this test uses an in-memory fake.)

Test-only; invariants unchanged; both pass `-race -count=20` plus full package suites; gofmt/vet/deadcode/build clean.

Codex autoreview: clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 0.46.1 - Unreleased
 
+### Fixed
+
+- Made two timing-sensitive tests robust on loaded CI runners.
+
 ## 0.46.0 - 2026-08-20
 
 ### Added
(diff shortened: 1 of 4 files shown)


steipete:
Added a third flake fix after this PR's own CI run flushed it out: the unikraft-cloud capture-limit test backgrounded `yes` before writing its PID file, so a fast byte-bound kill on a loaded runner could beat the write and the assertion read a missing file (run 32521847040). The noise now starts 200ms after the PID write via `exec` (same PID); kill/residue assertions unchanged; 10-round local soak green.

PR #1479, opened 2026-08-23
fix(machine0): fixed lease replay loses managed SSH key

steipete:
<details>
<summary>Additional instructions</summary>

**MUST:** Keep **Allow edits from maintainers** enabled for this PR so maintainers
can help update the branch when needed.

</details>

## What Problem This Solves

Fixed Machine0 lease replay can lose the managed SSH identity actually injected
into an already-owned VM. Machine0's `ls --json` summary may omit not only a
managed key's filename, but the entire `key` object; falling back to an unrelated
generic SSH identity then leaves replay stuck in authentication.

## Why This Change Was Made

Machine0 owns the injected VM key. Preserve explicitly returned filenames and
derive `machine0__<key-name>` for an existing non-public managed-key identity.
For fixed-ID replay only, the durable create attempt's selected key determines
whether sparse inventory requires one `machine0 get <vm> --json` detail read.
That detail must identify the same inventory-owned VM and the same selected
key; missing or mismatched identities and conflicting key types fail closed.
Explicit public keys retain
their existing public-key semantics, and attempts without a selected provider
key retain generic fallback without an extra detail read.

The inventory read remains authoritative for duplicate resource IDs,
deterministic VM names, existing claims, and exact fixed-lease resource
ownership. The conditional detail read supplements only the missing provider
key metadata; it does not replace inventory checks or add retries, timeout
extensions, provider-specific core logic, or compatibility shims.

The earlier claim that fixed replay never needs an additional machine-detail
read was incorrect for the live `key: null` inventory shape. A running replay
performs one catalog read, one ownership-authoritative inventory read, and
exactly one conditional detail read only when its durable selected provider key
is absent from the inventory summary. Initial acquisition retains its existing
inventory/readiness lifecycle.

## Redacted Live Failure Evidence

A real fixed-lease replay against previous PR head
`fa2ae1270db3b778581df8f92d946fe6d97b961f` observed an exactly owned, running
Machine0 VM whose `machine0 ls --json` inventory entry contained `key: null`.
Reading the same VM through `machine0 get <vm> --json` returned a full managed
key identity and filename matching the durable create attempt. Without the
detail hydration, replay instead selected an unrelated generic SSH identity;
SSH authentication repeatedly failed until the existing warmup budget expired.
The VM and session were cleaned up through supported lifecycle operations.

The new boundary regression first reproduced the same wrong-key selection on
that previous head, then passed after the provider-owned fix. Additional cases
cover complete and sparse inventory key metadata, exactly one conditional
detail read, mismatched resource identity, mismatched or missing selected keys,
explicit public-key fallback, and no-selected-key replay without a detail read.

## Verification

- `go test ./internal/providers/machine0 -run '^TestMachine0FixedAcquireReplayAdoptsExactMachine/inventory_omits_managed_key_entirely$' -count=1 -v` — fails on the previous head with the unrelated generic key, passes after the fix.
- `go test ./internal/providers/machine0 -run '^TestMachine0FixedAcquireReplayAdoptsExactMachine$' -count=1 -v`
- `go test ./internal/providers/machine0 -run 'TestPrepareLease|TestMachine0Fixed' -count=1`
- `go test -race ./internal/providers/machine0 -count=1`
- `go vet ./...`
- `go test ./...`
- Mandatory structured Codex-backed autoreview and TruffleHog secret scan.

## Remaining Merge Gate

A parent-owned real downstream fixed-lease replay must be rerun against the new
exact PR head and must prove successful managed-key SSH readiness without a
generic-key workaround. That exact-head live proof remains mandatory; do not
merge before it succeeds.

### DIFF
--- a/docs/providers/machine0.md
+++ b/docs/providers/machine0.md
@@ -97,15 +97,18 @@ home-root behavior.
 Machine0's `key` is the registered public or managed key name supplied to
 `machine0 new --key`. When the VM details return `key.fileName`, that filename
 is authoritative because it identifies the key actually injected into the VM.
-Crabbox resolves that filename under `SSH_KEY_PATH` or the default `~/.ssh`
-directory before readiness checks. It runs `machine0 ssh <vm> true` only when
-the resolved private-key file is absent, then verifies that the CLI actually
-materialized the file. Existing key files skip the Machine0 CLI entirely, so
-ordinary resume and checkpoint restart do not depend on the CLI's global
-`known_hosts`; Crabbox continues to perform its own direct SSH host-key
-handling. The provider-returned path overrides a generic `--ssh-key`, which is
-used only when Machine0 returns no filename. Set `SSH_KEY_PATH` when Machine0
-uses a custom key directory.
+When inventory omits the filename but retains a managed key name, Crabbox
+derives Machine0's canonical `machine0__<key-name>` filename instead. It
+resolves the provider-owned filename under `SSH_KEY_PATH` or the default
+`~/.ssh` directory before readiness checks. Crabbox runs
+`machine0 ssh <vm> true` only when the resolved private-key file is absent, then
+verifies that the CLI actually materialized the file. Existing key files skip
+the Machine0 CLI entirely, so ordinary resume and checkpoint restart do not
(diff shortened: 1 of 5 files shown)


steipete:
@clawsweeper re-review

PR #1480, opened 2026-08-23
feat: persist signed terminal run receipts

vincentkoc:
## What Problem This Solves

Retained remote runs could finish without a durable authenticated terminal
record. A lost finish response, post-command local failure, or mixed-version
CLI/coordinator rollout could leave execution complete while history still
appeared active or silently lacked the receipt operators expected.

## What Changed

Crabbox now creates a bounded Ed25519-signed schema-v2 terminal receipt for
SSH-backed runs and commits it atomically with terminal run state and retained
logs.

- Finish retries keep one exact signed payload.
- A lost response is recovered only when the stored receipt exactly matches.
- A successful finish response is not trusted by itself: the CLI retrieves the
  stored receipt and requires an exact match before marking the run recorded.
- Older coordinators that ignore the new field therefore fail closed instead
  of silently dropping evidence.
- Local `--attest` success and failure paths both use schema v2.
- Cleanup, download, artifact, proof, timing-record, and timing-JSON failures
  are incorporated before the terminal receipt is signed.
- If the coordinator terminal commit remains ambiguous after retries, the
  remote retry payload stays unchanged while a requested local attestation is
  re-signed with infrastructure exit 7.
- `crabbox receipt <run-id>` retrieves and verifies committed evidence.
- `--expected-signer sha256:<fingerprint>` pins campaign signer identity.

Receipts remain self-signed client evidence. They authenticate the signing key
and recorded fields, but do not provide coordinator time attestation or
cross-machine key continuity unless the caller pins the signer.

## Maintainer Decision

Sponsored as a core Crabbox protocol. Rollout is coordinator-first, and
receipt-bearing clients fail closed unless they can retrieve the exact receipt
the coordinator persisted.

## Review Findings Addressed

- **Mixed-version persistence:** a pre-receipt coordinator can return HTTP 200
  while ignoring an unknown receipt field. Receipt-bearing clients now verify
  the exact stored receipt after every successful finish and fail visibly when
  persistence is unavailable.
- **Successful local receipt format:** a live AWS canary on the earlier PR head
  exposed that successful `--attest` output still used schema v1. Success and
  non-zero SSH paths now share the schema-v2 terminal signer.
- **Late failures:** cleanup and deferred timing output now run before terminal
  finalization, so the signed receipt cannot claim exit 0 when the CLI exits
  non-zero.
- **Ambiguous coordinator commit:** exact remote retries remain immutable; the
  local artifact reports exit 7 only after terminal persistence cannot be
  proven.
- **Release-note ownership:** removed the PR-authored `CHANGELOG.md` entry;
  Crabbox release automation owns changelog updates.

## Exact-Head Evidence

Signed commit: `35775d6571dd8ba888d685a2883b4ddd35e061b2`  
Tree: `a55800a59a3292f4523504239615de29bf1ee537`

- `go test ./...` passed.
- `go test -race ./...` passed.
- `go vet ./...` and `go build ./...` passed.
- Focused receipt, cleanup, late timing failure, old-coordinator, retry, and
  lost-response tests passed under normal and race builds.
- Worker Vitest: 38 files passed, 2 skipped; 1,425 tests passed, 3 skipped.
- Worker build, TypeScript check, format check, and lint passed.
- Docs site generation passed.
- Two independent current-diff reviews returned GO with no remaining findings.
- The repository autoreview wrapper was attempted twice on the earlier exact
  head; both Sol subprocesses remained local sleepers without network or
  output, so those task-owned processes were stopped and are not represented
  as successful reviews.

Production delta: +1,141/-102 lines.  
Test delta: +1,437/-11 lines.  
Documentation delta: +123/-10 lines.

The production growth is the receipt schema/signing, coordinator transaction,
retrieval command, fail-closed persistence postcondition, and terminal
finalization capability. The test growth covers bounds, signatures,
cross-language parity, concurrency, rollback, retry, lost-response recovery,
mixed-version failure, signer pinning, and post-command failure paths.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -90,12 +90,14 @@ crabbox logs <run-id>                         print run logs
 crabbox events <run-id>                       print run events
 crabbox attach <run-id>                       follow events for an active run
 crabbox results <run-id>                      show test-result summaries
+crabbox receipt <run-id>                      retrieve and verify terminal evidence
 crabbox bench report [--json]                 aggregate local benchmark timing observations
 ```
 
 See [history](commands/history.md), [logs](commands/logs.md),
 [events](commands/events.md), [attach](commands/attach.md),
-[results](commands/results.md), [bench](commands/bench.md).
+[results](commands/results.md), [receipt](commands/receipt.md),
+[bench](commands/bench.md).
 
 ### Access and desktop
 
--- a/docs/commands/README.md
+++ b/docs/commands/README.md
@@ -30,6 +30,7 @@ same order as the CLI help.
 - [events](events.md)
 - [attach](attach.md)
 - [results](results.md)
+- [receipt](receipt.md)
 - [verify](verify.md)
 - [cache](cache.md)
 - [status](status.md)
--- a/docs/commands/receipt.md
+++ b/docs/commands/receipt.md
@@ -1,0 +1,61 @@
+# receipt
+
+`crabbox receipt <run-id>` retrieves a brokered run's committed terminal
+receipt, retained log, and terminal run record. It verifies the Ed25519
+signature and exact coordinator-observable bindings before printing JSON.
+
+```sh
+crabbox receipt run_a1b2c3d4
+crabbox receipt run_a1b2c3d4 --expected-signer sha256:<64-hex-digits>
+```
+
+Verification covers the run ID, lease ID, slug, provider, raw command digest,
+final exit code, sync and command duration, coordinator start timestamp,
+client-observed end timestamp, retained log hash, truncation state, public key,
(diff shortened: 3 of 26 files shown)


vincentkoc:
@clawsweeper re-review

The prior P1 is addressed on signed head
`5b9db22c995db04452c51132150c8488f3c05139`: receipt-bearing finish success now
requires retrieval of the exact persisted signed receipt. An older coordinator
that returns HTTP 200 while dropping the unknown field fails closed, covered by
the new legacy-coordinator regression. The same head also unifies successful
local attestations on schema v2 and incorporates late cleanup/timing failures
before terminal signing.

PR #1495, opened 2026-08-24
fix(machine0): preserve doctor failures and probe concurrently

steipete:
## Summary

- Preserve context-deadline and signal failures in Machine0 CLI diagnostics instead of allowing partially printed stdout, such as a valid version string, to mask the real error.
- Run the independent Machine0 version, live-size, and inventory doctor probes concurrently inside the provider adapter, so three slow Node CLI startups fit the existing provider-neutral ten-second doctor budget.
- Cancel sibling probes after a failure and retain the existing non-mutating doctor result and actionable authentication/install diagnostics.
- Add table-driven signal/deadline regressions and a four-second-per-probe doctor regression that succeeds within the shared timeout.

## Verification

- `gofmt -l internal/providers/machine0/backend.go internal/providers/machine0/backend_test.go internal/providers/machine0/client.go internal/providers/machine0/client_test.go` — clean.
- `go vet ./...` — passed with no findings.
- `go test ./internal/providers/machine0 -run 'TestDoctor|TestClientCommandFailure'`

  ```text
  ok  github.com/openclaw/crabbox/internal/providers/machine0  4.882s
  ```

- `go test -race -timeout 30m ./internal/providers/machine0/... ./internal/cli/...`

  ```text
  ok  github.com/openclaw/crabbox/internal/providers/machine0  (cached)
  ok  github.com/openclaw/crabbox/internal/cli                 369.112s
  ```

- Independent structured autoreview — clean, with no accepted/actionable findings.

An earlier concurrent full-suite attempt hit Go's default ten-minute package timeout; one isolated retry also exposed an unrelated Windows fake-SSH one-second-deadline flake. That test passed independently, and the complete retry above passed.

### DIFF
--- a/internal/providers/machine0/backend.go
+++ b/internal/providers/machine0/backend.go
@@ -429,17 +429,52 @@ func (b *backend) Resume(ctx context.Context, req ResumeRequest) error {
 }
 
 func (b *backend) Doctor(ctx context.Context, req DoctorRequest) (DoctorResult, error) {
-	version, err := b.api.Version(ctx)
-	if err != nil {
-		return DoctorResult{}, err
+	type probeResult struct {
+		version  string
+		sizes    []machineSize
+		machines []machine
+		err      error
 	}
-	sizes, err := b.api.Sizes(ctx)
-	if err != nil {
-		return DoctorResult{}, err
+	probeCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	results := make(chan probeResult, 3)
+	go func() {
+		version, err := b.api.Version(probeCtx)
+		results <- probeResult{version: version, err: err}
+	}()
+	go func() {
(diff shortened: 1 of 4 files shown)


steipete:
Live verification against a real Machine0 account, binary built from this branch: `doctor --provider machine0` completed in **5.9 s** (wall) within the 10 s provider budget and reported `cli=ready auth=ready control_plane=ready inventory=ready leases=1 sizes=23 version=1.0.164`. On v0.46.0 the same command failed as `machine0 --version failed: 1.0.164` — the deadline kill masked by stdout — because the three sequential Node-CLI probes could not fit the budget.

steipete:
Follow-up https://github.com/openclaw/crabbox/pull/1498 carries the required Unreleased entry and the explicit sibling-cancellation and canceled-context regression coverage that were still being prepared when this PR merged.

PR #1497, opened 2026-08-24
fix(machine0): persist recovery claims before readiness waits

steipete:
## Summary

- Persist a synchronously durable, name-scoped recovery claim immediately after Machine0 creates a VM and before readiness polling begins.
- Upgrade the recovery claim to immutable provider-ID scope as soon as Machine0 reports the matching machine, before SSH authentication can block.
- Preserve claims when kept provisioning fails or rollback fails, and remove claims only after unchanged-claim-guarded cleanup succeeds.
- Permit explicitly pending recovery claims to stop machines using only the same exact lease-derived name deletion already used by create rollback; normal ownership and release remain ID-fenced.
- Preflight explicit and default PUBLIC SSH keys from Machine0 JSON metadata, rejecting missing local private keys before a paid VM is created while preserving MANAGED-key materialization and fixed-lease replay.
- Add table-driven regression coverage for pending and ID-bound claim stages, kept failures, failed and stale rollback, failed ID binding, pending-stop ownership fencing, successful provisioning, and SSH-key selection.

## Verification

- `gofmt -l internal/providers/machine0/backend.go internal/providers/machine0/backend_test.go internal/providers/machine0/client.go internal/providers/machine0/client_test.go internal/providers/machine0/fixed.go` — clean.
- `git diff --check` — clean.
- `go vet ./...` — passed with no findings.
- `go test -race -timeout 30m ./internal/providers/machine0/... ./internal/cli/...` — the machine0 package passed; the monolithic CLI package hit pre-existing one-second fake-SSH and four-second Windows status timing assertions under host load.
- Complete provider race coverage: `go test -race ./internal/providers/machine0/...` — `ok github.com/openclaw/crabbox/internal/providers/machine0 (cached)`; its preceding uncached combined-suite run completed in `11.111s`.
- Exhaustive CLI race coverage, partitioned across every test name to avoid long-process host timing flakes:
  - `go test -race -timeout 20m ./internal/cli -run '^Test[A-H]'` — `ok github.com/openclaw/crabbox/internal/cli 133.691s`.
  - `go test -race -timeout 20m ./internal/cli -run '^Test[I-R]'` — `ok github.com/openclaw/crabbox/internal/cli 225.411s`.
  - `go test -race -timeout 20m ./internal/cli -run '^Test[S-Z]'` — `ok github.com/openclaw/crabbox/internal/cli 78.345s`.
- Independent structured review — clean; earlier findings about failed ID-binding snapshot preservation and pre-deletion claim fencing were fixed and regression-tested.

### DIFF
--- a/internal/providers/machine0/backend.go
+++ b/internal/providers/machine0/backend.go
@@ -129,6 +129,9 @@ func (b *backend) Acquire(ctx context.Context, req AcquireRequest) (LeaseTarget,
 		return b.acquireFixed(ctx, req)
 	}
 	cfg := b.configForRun()
+	if err := b.preflightSSHKey(ctx, cfg.Machine0.Key); err != nil {
+		return LeaseTarget{}, err
+	}
 	if err := b.validateCatalogSelection(ctx, cfg.Machine0.Size, cfg.Machine0.Region); err != nil {
 		return LeaseTarget{}, err
 	}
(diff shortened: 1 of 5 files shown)


steipete:
Live verification against a real Machine0 account, binary built from this branch:

**PUBLIC-key preflight** — with the account default key being a PUBLIC key whose private half is not on this host, `warmup --provider machine0 --machine0-size small` failed pre-create with:
```
Machine0 PUBLIC SSH key "mac-studio-sf" has no usable local private-key filename; select a managed key with --machine0-key <managed-key-name>
```
Account VM count was unchanged — no billable resource created. (Yesterday on v0.46.0 this exact configuration created a VM and spun 13.5 minutes in the ssh-auth wait.)

**Recovery claim under SIGKILL** — `warmup --machine0-key <managed-key> --machine0-size small`, process killed with SIGKILL 8 seconds after `machine0 new` succeeded:
- `claims list` immediately showed the pending claim (`cbx_34786c690da4`, name-scoped) — yesterday on v0.46.0 the same kill left zero claims and a RUNNING orphan only a manual `machine0 rm` could clear.
- `stop --provider machine0 --id cbx_34786c690da4 --machine0-release-policy destroy` released via the pending-claim path: `released lease=cbx_34786c690da4 machine=crabbox-tidal-lobster-7aeeefc2 action=destroy`.
- Final state: the machine gone from `machine0 ls`, machine0 claims count 0. A pre-existing unrelated VM on the account was untouched.

PR #1531, opened 2026-08-26
fix: recover local SSH multiplexing backpressure

### LINKED ISSUE #1528: crabbox run dies with `mux_client_request_session: send fds failed` under load (ControlMaster backpressure, not message size)
## Summary

On macOS, `crabbox run` intermittently dies ~1s in — before the remote command starts — with

```
mm_send_fd: sendmsg(0): Message too long
mux_client_request_session: send fds failed
```

and exits 255. The lease is provisioned and then torn down without the job ever executing.

The trigger is **backpressure on the ssh ControlMaster socket, not message size**. I originally
diagnosed it as an env/argv size limit and that was wrong in two ways; both corrections are
below, with a reproduction that needs no crabbox at all.

## It is not a size limit

`net.local.dgram.maxdgram` (2048 on macOS) does **not** apply: OpenSSH's ControlPath socket is
`SOCK_STREAM`, so the relevant buffer is `net.local.stream.sendspace` (8192 by default).

More importantly, size does not predict the failure. Observed across runs on the same host,
same week, same provider (Azure, fresh boxes each time), varying only the repository:

| forwarded env payload | outcome |
|---|---|
| ~1.9 KB | failed 6/6 on six distinct fresh boxes |
| ~3.1 KB | passed |
| ~11 KB  | passed |

The 1.9 KB case is below even the limit it was supposedly exceeding.

## Reproduction (no crabbox, no ssh, ~1 second)

`mm_send_fd` passes stdin/stdout/stderr to the master with `sendmsg` + `SCM_RIGHTS`. On macOS
that call returns `EMSGSIZE` exactly when the socket's send buffer is **full** — regardless of
how small the message is:

```python
import array, errno, os, socket

def outcome(queued: int) -> str:
    a, b = socket.socketpair(socket.AF_UNIX, socket.SOCK_STREAM)
    try:
        a.setblocking(False)
        if queued:
            a.send(b"x" * queued)
        fd = os.open(os.devnull, os.O_RDONLY)
        try:
            a.sendmsg([b"\0"], [(socket.SOL_SOCKET, socket.SCM_RIGHTS, array.array("i", [fd]))])
            return "fd passed"
        except OSError as e:
            return errno.errorcode.get(e.errno, str(e.errno))
        finally:
            os.close(fd)
    finally:
        a.close(); b.close()

for q in (0, 1024, 2048, 4096, 6144, 7168, 8000, 8192):
    print(q, "->", outcome(q))
```

Output on macOS (`net.local.stream.sendspace` = 8192):

```
0 -> fd passed        4096 -> fd passed      8000 -> fd passed
1024 -> fd passed     6144 -> fd passed      8192 -> EMSGSIZE
2048 -> fd passed     7168 -> fd passed
```

So: with any room at all the fd passes, at every size. Only a full buffer fails. Whether a given
`crabbox run` fails therefore depends on whether the mux master has drained the session-open
request before the client passes its descriptors — a race, and one that gets more likely as host
load rises (the failing runs above were on a machine running several concurrent leases).

## Why this can't be fixed in-process

`mm_send_fd` is OpenSSH's, inside the `ssh` client binary. crabbox shells out to `ssh`, so it
cannot make that call retry. Treating `EMSGSIZE` as retryable after polling for buffer space
would be the ideal fix, but that is an OpenSSH change, not one crabbox can make.

## Ask

Handle it at the invocation layer, where crabbox does have control:

1. Detect `mux_client_request_session: send fds failed` on an `ssh` exec and **retry that exec
   once**.
2. If it recurs, fall back to `-o ControlMaster=no`. Without multiplexing there is no fd pass
   over a shared socket, so the failure mode cannot occur.

The second half looks cheap from the outside: `SSH.NoControlMaster` already exists
(`internal/providers/external/protocol.go`) and is used by the asciibox, namespaceinstance and
phala backends, so this is reusing an existing knob rather than adding one.

## Why it matters beyond the retry

Callers currently cannot distinguish this from a vanished box: both surface as exit 255. Any
launcher with an infra-flake retry will therefore provision a **second** box to reproduce a
failure that never involved a box, doubling cost per attempt — and if a launcher instead
classifies exit 255 as deterministic, it will refuse launches that a retry would have won. A
distinguishable exit status or a machine-readable reason for this case would help callers even
if the retry lands.

Happy to open a PR for (1) and (2) if the approach looks right.

### LINKED ISSUE #1560: ssh mux fd-passing fails intermittently with EMSGSIZE ("mm_send_fd: sendmsg: Message too long") under load — retry once, then fall back to ControlMaster=no
## What happens

`crabbox run` intermittently dies ~1 second in, before the remote command
starts, with:

    mm_send_fd: sendmsg(0): Message too long
    mux_client_request_session: send fds failed

exit 255. Today the launcher reads any exit 255 as "box vanished mid-run" and
retries on a **fresh box** — two boxes provisioned per attempt for a failure
that never involved a box at all.

## Root cause

This is OpenSSH's mux client failing to hand stdin/stdout/stderr to the
ControlMaster over the `ControlPath` socket via `sendmsg(2)` with `SCM_RIGHTS`.

Two plausible-looking explanations turned out to be wrong, and both are worth
ruling out explicitly since the error text invites them:

- **Not `net.local.dgram.maxdgram` (2048).** The ControlPath socket is
  `SOCK_STREAM`, so the governing limit is `net.local.stream.sendspace`
  (8192 by default on macOS), not the datagram limit.
- **Not a message-size threshold at all.** Reproduced on a bare
  `AF_UNIX SOCK_STREAM` socketpair: a `sendmsg` carrying `SCM_RIGHTS` returns
  `EMSGSIZE` exactly when the peer has **not drained** the socket. With any
  room at all in the buffer, the descriptor passes at every payload size we
  swept (0/1k/2k/4k/6k/7k/8000 bytes queued) — only a completely full 8192-byte
  buffer fails. Field data agrees: a 3,140-byte and an 11,060-byte manifest
  both passed while a 1,928-byte one failed 6/6 in a row, under a supposed
  "2 KB" ceiling that doesn't hold up.

So this is backpressure/a drain race against the ControlMaster, not a size
cap — transient by nature, and load-sensitive (we've seen it recur multiple
times in a row on a busy host and never on a quiet one).

## Proposed fix

`crabbox` can't fix OpenSSH's `sendmsg` call itself, but it owns the
invocation, so it can retry once, and if the failure recurs, fall back to
`ControlMaster=no` for that connection — with no ControlPath, no descriptor
is ever passed over a shared socket, so the failure mode cannot occur at all.
This reuses `SSHTarget.NoControlMaster`, which a few backends (asciibox,
namespaceinstance, phala) already set for other reasons — no new knob.

Patch attached below, against `v0.46.0` (the tag we have pinned). Summary:

- `internal/cli/ssh.go`: on a failed SSH attempt, a `muxFailureWatcher` wraps
  stderr and remembers (without buffering the whole stream — a gate can emit
  megabytes of it) whether either failure marker appeared, matching each
  independently so a typo in one marker can't hide behind the other. If the
  markers were seen, retry once with the same settings, then once more with
  `NoControlMaster=true` before giving up.
- The watcher forwards every byte unchanged (so nothing about the log is
  lost) and keeps a bounded 128-byte carry-over so a marker split across two
  pipe writes is still detected.

## Tests

5 new Go test cases, all passing against the pinned tag:

- exact real-world stderr is recognised; each marker recognised on its own
  (not just the pair combined — a mutation flipping one marker independently
  was caught this way during our own review); near-miss strings correctly
  don't match
- the watcher forwards everything unchanged while still detecting the
  failure
- a marker split across two `Write` calls is still detected
- the watcher never buffers the whole stream (checked after 8 MiB across 8
  writes)
- the `NoControlMaster` fallback actually removes `ControlMaster=auto` /
  `ControlPath=...` from the ssh invocation rather than merely adding a flag
  alongside them

## Diff

<details>
<summary>mux-fd-retry.patch (against v0.46.0)</summary>

```diff
diff --git a/internal/cli/ssh.go b/internal/cli/ssh.go
index 8768721..477342a 100644
--- a/internal/cli/ssh.go
+++ b/internal/cli/ssh.go
@@ -850,11 +850,36 @@ func runSSHStreamResult(ctx context.Context, target SSHTarget, remote string, st
 	for _, port := range sshPortCandidates(target.Port, target.FallbackPorts) {
 		probe := target
 		probe.Port = port
+		watcher := &muxFailureWatcher{dst: stderr}
 		cmd := sshCommandContext(ctx, probe, sshArgsNoInput(probe, remote)...)
-		err := runSSHCommand(cmd, stdout, stderr)
+		err := runSSHCommand(cmd, stdout, watcher)
 		if err == nil {
 			return 0, nil
 		}
+		// The mux client could not pass its descriptors. That is backpressure on
+		// the ControlMaster socket, not a bad port and not a dead box, so the
+		// port loop below would retry the wrong thing. Try once more, and if it
+		// recurs take multiplexing out of the picture: with ControlMaster=no
+		// there is no descriptor passing over a shared socket at all.
+		if watcher.seen && !probe.NoControlMaster {
+			for _, direct := range []bool{false, true} {
+				retry := probe
+				retry.NoControlMaster = direct
+				retryWatcher := &muxFailureWatcher{dst: stderr}
+				retryCmd := sshCommandContext(ctx, retry, sshArgsNoInput(retry, remote)...)
+				retryErr := runSSHCommand(retryCmd, stdout, retryWatcher)
+				if retryErr == nil {
+					return 0, nil
+				}
+				err = retryErr
+				if !retryWatcher.seen {
+					break
+				}
+			}
+			if err == nil {
+				return 0, nil
+			}
+		}
 		lastErr = err
 		lastCode = exitCode(err)
 		if !shouldRetrySSHPort(err) {
@@ -904,6 +929,68 @@ func sshArgsNoInput(target SSHTarget, remote string) []string {
 	return sshArgsNoInputWithOptions(target, remote, "10", "3")
 }
 
+// muxFDPassFailureMarkers are the two lines OpenSSH's mux client prints when it
+// cannot hand stdin/stdout/stderr to the ControlMaster.
+//
+// The failing call is sendmsg(2) with SCM_RIGHTS. On macOS it returns EMSGSIZE
+// ("Message too long") when the ControlPath socket's send buffer is FULL — not
+// when the message is large. Reproduced on a bare AF_UNIX SOCK_STREAM
+// socketpair: with any room at all the descriptor passes at every payload size
+// tried (0/1k/2k/4k/6k/7k/8000 bytes queued); only a completely full buffer
+// (net.local.stream.sendspace, 8192 by default) fails. It is therefore a drain
+// RACE against the master, and it is load-sensitive: the same repository has
+// been observed to fail six times in a row while sessions with far larger
+// forwarded environments succeeded.
+//
+// crabbox cannot fix the sendmsg itself — that is inside the ssh client — but it
+// owns the invocation, so it can retry once and then take the connection out of
+// multiplexing entirely, where no descriptor is passed over a shared socket and
+// the failure mode cannot occur.
+var muxFDPassFailureMarkers = []string{
+	"mux_client_request_session: send fds failed",
+	"mm_send_fd: sendmsg",
+}
+
+func isMuxFDPassFailure(stderr string) bool {
+	for _, marker := range muxFDPassFailureMarkers {
+		if strings.Contains(stderr, marker) {
+			return true
+		}
+	}
+	return false
+}
+
+// muxFailureWatcher forwards everything it is given and remembers only whether
+// the mux markers appeared. It deliberately does NOT buffer the stream: a gate
+// run emits megabytes of stderr, and the markers appear within the first second
+// (the session never opens). A bounded carry-over keeps a marker split across
+// two writes detectable.
+type muxFailureWatcher struct {
+	dst   io.Writer
+	carry []byte
+	seen  bool
+}
+
+func (w *muxFailureWatcher) Write(p []byte) (int, error) {
+	if !w.seen {
+		probe := append(w.carry, p...)
+		if isMuxFDPassFailure(string(probe)) {
+			w.seen = true
+			w.carry = nil
+		} else {
+			keep := 128
+			if len(probe) < keep {
+				keep = len(probe)
+			}
+			w.carry = append(w.carry[:0], probe[len(probe)-keep:]...)
+		}
+	}
+	if w.dst == nil {
+		return len(p), nil
+	}
+	return w.dst.Write(p)
+}
+
 func shouldRetrySSHPort(err error) bool {
 	return exitCode(err) == 255
 }
diff --git a/internal/cli/ssh_mux_fd_retry_test.go b/internal/cli/ssh_mux_fd_retry_test.go
new file mode 100644
index 0000000..25910e9
--- /dev/null
+++ b/internal/cli/ssh_mux_fd_retry_test.go
@@ -0,0 +1,122 @@
+package cli
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+)
+
+// The two lines OpenSSH emits, verbatim from a failed `crabbox run` on macOS.
+const muxFailureStderr = "mm_send_fd: sendmsg(0): Message too long\n" +
+	"mux_client_request_session: send fds failed\n"
+
+func TestIsMuxFDPassFailureMatchesTheRealOutput(t *testing.T) {
+	if !isMuxFDPassFailure(muxFailureStderr) {
+		t.Fatal("the real mux failure output must be recognised")
+	}
+	// Each marker must be recognised ON ITS OWN. Keyed only on the combined
+	// stderr above, a typo in either marker is invisible -- the other one still
+	// matches and the case stays green. Verified by mutation on 2026-08-27:
+	// flipping "mm_send_fd: sendmsg" to a wrong string left this test passing
+	// until these two lines existed. The strings below are LITERAL OpenSSH
+	// output on purpose: asserting against muxFDPassFailureMarkers itself would
+	// move with the code under test and prove nothing.
+	for _, line := range []string{
+		"mux_client_request_session: send fds failed\r\n",
+		"mm_send_fd: sendmsg(0): Message too long\r\n",
+	} {
+		if !isMuxFDPassFailure(line) {
+			t.Fatalf("each real marker must be detected on its own: %q", line)
+		}
+	}
+	for _, line := range []string{
+		"", "ssh: connect to host 10.0.0.1 port 22: Connection refused\n",
+		"Warning: Permanently added '10.0.0.1' to the list of known hosts.\n",
+		// Near-misses must NOT match: this decides whether we drop multiplexing.
+		"mux_client_request_session: session request failed\n",
+		"sendmsg: Message too long\n",
+	} {
+		if isMuxFDPassFailure(line) {
+			t.Fatalf("must not match unrelated stderr: %q", line)
+		}
+	}
+}
+
+func TestMuxFailureWatcherForwardsEverythingAndDetects(t *testing.T) {
+	var sink bytes.Buffer
+	w := &muxFailureWatcher{dst: &sink}
+	if _, err := w.Write([]byte("ordinary gate output\n")); err != nil {
+		t.Fatal(err)
+	}
+	if w.seen {
+		t.Fatal("must not fire on ordinary output")
+	}
+	if _, err := w.Write([]byte(muxFailureStderr)); err != nil {
+		t.Fatal(err)
+	}
+	if !w.seen {
+		t.Fatal("must fire once the markers appear")
+	}
+	if sink.String() != "ordinary gate output\n"+muxFailureStderr {
+		t.Fatalf("the watcher must forward every byte unchanged, got %q", sink.String())
+	}
+}
+
+func TestMuxFailureWatcherDetectsAMarkerSplitAcrossWrites(t *testing.T) {
+	// ssh's stderr arrives in whatever chunks the pipe delivers; a marker split
+	// across two writes must still be detected, or the fallback silently never
+	// fires and the fix looks like it works while doing nothing.
+	var sink bytes.Buffer
+	w := &muxFailureWatcher{dst: &sink}
+	marker := "mux_client_request_session: send fds failed"
+	split := len(marker) / 2
+	if _, err := w.Write([]byte(marker[:split])); err != nil {
+		t.Fatal(err)
+	}
+	if w.seen {
+		t.Fatal("half a marker is not a match")
+	}
+	if _, err := w.Write([]byte(marker[split:] + "\n")); err != nil {
+		t.Fatal(err)
+	}
+	if !w.seen {
+		t.Fatal("a marker split across writes must still be detected")
+	}
+	if sink.String() != marker+"\n" {
+		t.Fatalf("bytes must pass through unchanged, got %q", sink.String())
+	}
+}
+
+func TestMuxFailureWatcherDoesNotBufferTheWholeStream(t *testing.T) {
+	// A gate emits megabytes of stderr. Buffering it to scan for two lines would
+	// trade one resource bug for another.
+	w := &muxFailureWatcher{dst: nil}
+	big := strings.Repeat("x", 1<<20)
+	for i := 0; i < 8; i++ {
+		if _, err := w.Write([]byte(big)); err != nil {
+			t.Fatal(err)
+		}
+	}
+	if len(w.carry) > 256 {
+		t.Fatalf("carry-over must stay bounded, got %d bytes after 8MiB", len(w.carry))
+	}
+}
+
+func TestNoControlMasterRemovesTheMultiplexingOptions(t *testing.T) {
+	// The fallback's whole premise: with NoControlMaster the ssh invocation
+	// carries no ControlPath, so no descriptor is passed over a shared socket.
+	base := SSHTarget{User: "u", Host: "h", Port: "22"}
+	muxed := strings.Join(sshBaseArgs(base), " ")
+	if !strings.Contains(muxed, "ControlMaster=auto") {
+		t.Fatalf("baseline should multiplex, got %q", muxed)
+	}
+	direct := base
+	direct.NoControlMaster = true
+	got := strings.Join(sshBaseArgs(direct), " ")
+	if !strings.Contains(got, "ControlMaster=no") || !strings.Contains(got, "ControlPath=none") {
+		t.Fatalf("fallback must disable multiplexing, got %q", got)
+	}
+	if strings.Contains(got, "ControlMaster=auto") {
+		t.Fatalf("fallback must not still request multiplexing, got %q", got)
+	}
+}
```

</details>

steipete:
Fixes https://github.com/openclaw/crabbox/issues/1528.
Fixes https://github.com/openclaw/crabbox/issues/1560.

An overloaded local OpenSSH control socket can fail while passing the session's file descriptors, before the remote command starts. This change retries that exact failure once through multiplexing, then makes one direct connection if it recurs. It preserves the lease, key, host-key policy, endpoint, proxy, command, and complete replayable input. Ordinary remote errors do not authorize this recovery.

The retry decision uses OpenSSH's separate local diagnostic channel, not remote stderr. It requires native exit 255 and the complete two-line descriptor-transfer failure record, with no extra records. Overflow, incomplete capture, and diagnostic delivery errors prevent retry. Windows, secret authentication, and explicitly disabled multiplexing keep their existing paths.

Maintainer review found and fixed two capture issues before landing: an unbounded regular-file log, and synchronous live forwarding that could stall capture shutdown. Diagnostics now drain through a private FIFO into bounded memory. Capture closes and removes its resources before forwarding up to 64 KiB per attempt, with an explicit truncation notice; the retry detector retains at most 512 bytes. Persistent SSH masters cannot prolong capture, and caller stdout/stderr retain their existing transport behavior.

Validation for head `2b1b16eee229293ecdbde52f8a460e4f328eda93`:

- Full Codex review: no actionable findings at any severity after both corrections.
- Focused SSH, mux, WSL wrapping, artifact collection, and input replay tests passed with the race detector (54.138s). A separate full-output-pipe cancellation regression passed with the race detector (4.300s).
- Replacing only the capture function with the original regular-file implementation makes the new FIFO regression fail; the corrected implementation passes.
- `go vet ./internal/cli`, documentation checks, native CLI build, and Windows/amd64 package cross-build passed.
- Independent source-blind validation used the built CLI, native macOS OpenSSH, a real loopback SSH server, and genuine native descriptor-transfer failures. One failure recovered through mux retry; two failures selected a direct connection. The original command/input and connection identity were preserved, with one remote workload execution.
- Remote stderr containing the matching diagnostic pair with exits 37 and 255 did not replay the workload. A retained native ControlMaster remained reusable after capture closed. Diagnostic volume above the native pipe capacity preserved a 32 MiB application payload, respected the per-attempt diagnostic cap, and produced no sampled disk log allocation.

The native candidate was built from the final integrated source tree before the merge commit was recorded; no production source changed afterward. Hosted checks build the committed head independently.

Candidate SHA-256: `835725bf7b9eb0bd4def244cc8000ab5d2a47dd369edef9cf93d03d7b7c29f1c`.

Maintainer scope decision: accept the strict pre-session recovery with native macOS fault proof, Linux native SSH smoke coverage, and Unix regressions. Native Linux fault injection is not claimed; OpenSSH source review confirms that descriptor handoff precedes opening the remote session.

Proof scope: native loopback SSH, not a cloud-provider allocation. Accepted runs use isolated local and remote user directories. Discarded fixture calibration initially used the account HOME and created task-owned ownership state; CLI cleanup removed those records, but preexisting directory metadata was not snapshotted, so the calibration is not claimed to have left all metadata unchanged. The unrelated static one-shot ownership issue remains tracked at https://github.com/openclaw/crabbox/issues/1584.

Exact-head validation runs: [CI](https://github.com/openclaw/crabbox/actions/runs/33224955919), [Release Check](https://github.com/openclaw/crabbox/actions/runs/33224955909), [connector smokes](https://github.com/openclaw/crabbox/actions/runs/33224955961), and [docs UI proof](https://github.com/openclaw/crabbox/actions/runs/33224955929). Release Check is the ordinary PR validation workflow; this PR does not publish a release.

<details>
<summary>Sanitized native terminal proof</summary>

```text
Native OpenSSH CLI behavior verification — final bounded-capture build

Final binary SHA-256:
835725bf7b9eb0bd4def244cc8000ab5d2a47dd369edef9cf93d03d7b7c29f1c
Retained pre-fix control SHA-256 (not asserted exact current main):
c15a6d4dda671c543eaa49d1c9f702f0adce90a591aa707658a9cf92aea8fb06
Native client: OpenSSH_10.3p1, LibreSSL 3.3.6

Actual command structure; task-owned paths/account/port consistently redacted:
env -i HOME=<fixture>/<case>/home \
  PATH=<fixture>/bin:<system-tool-path> USER=<fixture-user> LOGNAME=<fixture-user> \
  LANG=C TERM= TMPDIR=<fixture>/<case>/tmp \
  XDG_CONFIG_HOME=<fixture>/<case>/config XDG_CACHE_HOME=<fixture>/<case>/cache \
  XDG_STATE_HOME=<fixture>/<case>/state CRABBOX_CONFIG=<fixture>/empty.yaml \
  CRABBOX_SSH_KEY=<fixture>/client_key MUX_PROOF_TRACE=<fixture>/<case>/trace.jsonl \
  MUX_PROOF_FAILURES=<0|1|2> \
  <candidate> run --provider ssh --target macos --static-host 127.0.0.1 \
  --static-user <fixture-user> --static-port <loopback-port> \
  --static-work-root <fixture>/work --no-sync --no-hydrate --keep \
  --slug mux-<case> --script-stdin < <fixture>/<case>/payload.sh

Fixture: dedicated local sshd, fresh ephemeral keys, private authorized_keys,
explicit native SSH -F /dev/null, private client and forced remote HOME.
The wrapper invokes real /usr/bin/ssh for every success and failure.
An engineered mux receiver leaves the native session packet queued at FD handoff.
Only fault attempts add an inert shell comment; original command hashes remain
identical across retries, and padded-native command hashes are separately retained.
Native error/status bytes are forwarded unchanged, never synthesized.

Observed cases:
before      CLI=  7 native_upload/workload=[mux upload:255 (native FD failure)] executions=0 script_bytes=172
baseline    CLI=  0 native_upload/workload=[mux:0] executions=1 script_bytes=234
reuse       CLI=  0 native_upload/workload=[mux:0] executions=1 script_bytes=228
retry       CLI=  0 native_upload/workload=[mux:255 -> mux:0] executions=1 script_bytes=228
recover     CLI=  0 native_upload/workload=[mux:255 -> mux:255 -> direct:0] executions=1 script_bytes=232
remote-37   CLI= 37 native_upload/workload=[mux upload:0; mux workload:37] executions=1 script_bytes=259
remote-255  CLI=255 native_upload/workload=[mux upload:0; mux workload:255] executions=1 script_bytes=261
volume      CLI=  0 native_upload/workload=[mux:0] executions=1 script_bytes=320

Each successful script used unique multiline UTF-8 data and executed exactly once.
The pre-fix control executed zero workloads. Submitted script bytes/hash matched
all upload attempts; retry/fallback preserved original command hash, key, host,
port, and host-check setting. Remote errors retained37/255 and did not cause mux
recovery. No alternate-endpoint attempt was observed; a guard prevented any
alternate endpoint from authenticating or running a workload.

Actual native fault evidence (exit255, zero remote-output bytes):
mm_send_fd: sendmsg(0): Message too long
mux_client_request_session: send fds failed

FIFO volume probe: real native -vvv, RekeyLimit=1M, direct loopback transport.
Native diagnostics: total=341784 bytes; largest attempt=167008 bytes.
Measured OS pipe capacity=65536 bytes; per-attempt retained prefix=65536 bytes.
Observed notice: [local SSH diagnostics truncated after 65536 bytes]
Actual CLI stderr=66907 bytes including ordinary run metadata.
Application data:33554432 generated bytes preserved, plus unique UTF-8 marker.
Sampled diagnostic FIFO objects=24; allocated diagnostic data bytes=0; no regular-file spool observed.

ControlMaster reuse: same native PID/socket remained live and reusable across two
successful candidate runs after diagnostic capture closure; both workloads ran once.

Cleanup: actual CLI stop for every claim returned0; claims list returned[].
All owned masters, listener and descendants stopped; listener closed; no owned
processes remained. Temporary private/public keys, authorization files, runtime
state and temporary diagnostics removed. Compact redacted proof retained.

Calibration caveat: an earlier discarded fixture run inherited the account HOME.
Its task-owned runtime state was removed through the actual CLI. No operator SSH
configuration/key files were modified. Original permissions of preexisting runtime
directories were not recorded, so a broader no-metadata-change claim is not made.
Accepted final runs used the verified private remote HOME.

Scope: --keep; local macOS only. No claim of real cloud lifecycle, naturally
congested production sockets, Windows coverage, or static one-shot cleanup.
No repository source/diff/test/history/review-log inspection was used as proof.
```

</details>


steipete:
Landed in https://github.com/openclaw/crabbox/commit/139ad77cb30eb64d88c7ceeda1cfc407c7e968cf through https://github.com/openclaw/crabbox/pull/1531.

The shared SSH transport now retries only the complete local OpenSSH descriptor-transfer failure once, then falls back to a direct connection. Remote stderr and ordinary remote failures cannot authorize this recovery. Review also replaced an unbounded diagnostic file with bounded FIFO capture that ends before output forwarding, preserving cancellation and persistent-master behavior.

All exact-head checks passed: [full CI](https://github.com/openclaw/crabbox/actions/runs/33224955919), [protected Release Check](https://github.com/openclaw/crabbox/actions/runs/33224955909), [connector smokes](https://github.com/openclaw/crabbox/actions/runs/33224955961), and [docs UI proof](https://github.com/openclaw/crabbox/actions/runs/33224955929). The final Codex review found no actionable issues.

Actual CLI/native OpenSSH proof passed retry, direct fallback, exact input preservation, single workload execution, remote exits 37/255 without replay, persistent-master reuse, and diagnostic overflow with a preserved 32 MiB payload. Native fault injection was demonstrated on macOS; Linux native SSH smoke and Unix regressions passed, but Linux native fault injection is not claimed. The PR body retains the terminal evidence, candidate hash, cleanup confirmation, and discarded-calibration caveat. The unrelated static one-shot ownership issue remains https://github.com/openclaw/crabbox/issues/1584.

PR #1533, opened 2026-08-26
fix(machine0): explain unclaimed lease recovery without trusting hash matches

steipete:
## Problem

Inspecting a Machine0 lease that has no local claim fails with a name-validation error instead
of anything actionable:

```
$ crabbox inspect --provider machine0 --id cbx_… --json
machine0 get cbx_… --json failed: ✗ Name must start and end with a letter or number, and
contain only letters, numbers, and hyphens
```

`Resolve` only rewrites the lookup key when a claim exists, so unclaimed it passes the raw
Crabbox lease ID to `machine0 get`. Lease IDs contain an underscore; Machine0 names allow only
letters, numbers and hyphens, so the ID can never be a valid name. That is exactly the state
after a claim is lost or expires — precisely when an operator most wants to look at the lease.

## What this does

Discovery is now possible, but it does **not** grant adoption.

The lease ID is not recoverable from the remote resource by label (`lease`/`slug` come from
`directLeaseLabels()` and are local claim data). The one remote-discoverable link is the machine
name, which `machine0MachineName()` ends with `-%08x` of `fnv32a(leaseID)` — derived from the
lease ID alone.

That suffix is only 32 bits, so it can collide and **cannot prove full lease identity**. An
unclaimed canonical lease ID therefore fails closed with a candidate-name hint rather than
resolving into a usable target:

```
machine0 lease cbx_… has no local claim; candidate "crabbox-…" matches only a short name hash:
inspect the machine and use its explicit name with --reclaim to adopt it
```

Nothing matching gives a clean `lease/server not found`; several matches fail explicitly.
Slugs, machine names, and the claimed path keep their direct `Get` with no added list call.
Durable claims, not name hashes, establish ownership — discovery alone never authorizes reuse
or deletion.

`machine0MachineName` now shares the suffix helper instead of recomputing the hash. Generated
names are byte-identical, verified against real lease/machine pairs:

```
cbx_0a52cf884f68  ->  crabbox-amber-crayfish-7c073acc   (unchanged)
cbx_b0b0d3f1d563  ->  crabbox-openclaw-b3794-ffdbed89   (unchanged)
cbx_abcdef123456  ->  crabbox-cbx-abcdef1234-c80c2195   (unchanged)
```

### Authenticated live Machine0 run

Real `machine0` lease provisioned on the OpenClaw Team host, A/B against a binary built from
this PR's merge base. Endpoint addresses redacted; the lease was destroyed afterwards.

Lease `cbx_d19b…` → machine `crabbox-coral-lobster-f3c8f439`. The claim file was moved aside to
produce a genuine unclaimed state, then restored.

**1. Claimed, this PR — unchanged, still resolves:**

```
{"id":"cbx_d19b…","slug":"coral-lobster-58a4","provider":"machine0","state":"ready",
 "serverId":"b1c515fb-…","serverType":"large","host":"<redacted>", …}
rc=0
```

**2. Unclaimed, merge-base binary — reproduces the reported bug:**

```
machine0 get cbx_d19b… --json failed: ✗ Name must start and end with a letter or number, and
contain only letters, numbers, and hyphens
```

**3. Unclaimed, this PR, by lease ID — fails closed with the candidate hint:**

```
machine0 lease cbx_d19b… has no local claim; candidate "crabbox-coral-lobster-f3c8f439"
matches only a short name hash: inspect the machine and use its explicit name with --reclaim
to adopt it
```

**4. Unclaimed, this PR, by explicit machine name — direct `Get` path unaffected:**

```
{"id":"","slug":"unknown","provider":"machine0","state":"ready","serverId":"b1c515fb-…", …}
```

Note `id:""` / `slug:"unknown"`: the machine is visible, but no lease identity is fabricated.

**5. Unclaimed, this PR, unknown lease ID — clean not-found:**

```
lease/server not found: cbx_ffffffffffff
```

**6. The documented remedy end to end — follow the hint into `--reclaim`:**

```
run context:
  run=run_a88ed602c20e
  lease=cbx_b312… slug=crabbox-coral-lobster-f3c8f439 provider=machine0 target=linux type=large
  ssh=ubuntu@<redacted>:22
running on <redacted> echo RECLAIM_OK; hostname; uname -m
RECLAIM_OK
crabbox-coral-lobster-f3c8f439
x86_64
command complete in 1.085s total=2.226s ... exit=0
```

Adoption mints a fresh claim and runs a real command on the actual VM, so the hint in case 3
leads somewhere that works.

### Local validation

```
$ go build ./...
$ go test ./internal/providers/machine0/...
ok  	github.com/openclaw/crabbox/internal/providers/machine0	87.300s
$ gofmt -l internal/providers/machine0
$ go vet ./internal/providers/machine0/...
```

The first added test fails before the production change:

```
--- FAIL: TestResolveUnclaimedIdentifier/canonical_lease_ID
    backend_test.go:844: unexpected Get("cbx_abcdef123456"), want ""
```

## ClawSweeper items

- **Add real behavior proof / resolve merge risk** — addressed by the authenticated live run above.
- **Remove the release-owned changelog entry (P3)** — *skipped, deliberately.* crabbox's own
  `AGENTS.md` does not reserve `CHANGELOG.md` for release work (it only requires bumping
  version-carrying files during a release), and the file already carries unreleased entries from
  normal PRs, including contributor thanks. The entry here was added and then curated by the
  maintainer to avoid a parallel-PR conflict.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,6 @@
 
 ### Fixed
 
-- Added actionable Machine0 recovery hints for unclaimed lease IDs without treating short name hashes as proof of lease ownership.
 - Required exact host/project-scoped Semaphore job ownership claims and fresh provider verification before stopping jobs.
 - Required exact API-scoped Morph instance ownership claims and fresh provider verification before pause or deletion.
 - Returned machine-readable missing checkpoint verdicts and made checkpoint deletion idempotent when local records or coordinator-owned resources are already absent.
@@ -22,6 +21,7 @@
 - Removed coordinator URL credentials from Code and WebVNC browser links, opener arguments, and viewer bootstrap form actions. Thanks @coygeek.
 - Redacted configured and runtime-only credentials from coordinator-stored run failure diagnostics while preserving raw command output. Thanks @coygeek.
 - Retried temporary Machine0 read outages within the existing operation deadline while keeping provider mutations single-attempt.
+- Added actionable Machine0 recovery hints for unclaimed lease IDs without treating short name hashes as proof of lease ownership.
 
 ## 0.46.1 - 2026-08-24
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 ### Fixed
 
+- Added actionable Machine0 recovery hints for unclaimed lease IDs without treating short name hashes as proof of lease ownership.
 - Required exact host/project-scoped Semaphore job ownership claims and fresh provider verification before stopping jobs.
 - Required exact API-scoped Morph instance ownership claims and fresh provider verification before pause or deletion.
 - Returned machine-readable missing checkpoint verdicts and made checkpoint deletion idempotent when local records or coordinator-owned resources are already absent.
--- a/docs/providers/machine0.md
+++ b/docs/providers/machine0.md
@@ -184,9 +184,17 @@ Use `--keep` to keep a normal Crabbox run lease for later reuse. Adopt an
 existing unclaimed Machine0 VM only through an explicit `--reclaim` reuse;
 destructive release requires an exact local claim bound to the Machine0 ID.
 
+If a local claim is missing, a canonical Crabbox lease ID can suggest a
+Crabbox-named VM from Machine0 inventory, but the short name hash cannot prove
+the full lease identity. Even a unique match fails closed with a candidate-name
+hint. Inspect that machine and use its explicit name with `--reclaim` to adopt
+it; Crabbox fetches its full details, including its SSH key and endpoint.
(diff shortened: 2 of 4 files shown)


steipete:
@clawsweeper re-review

Added the authenticated Machine0 proof requested: a real lease on the Team host, A/B against a
binary built from this PR's merge base, covering the candidate hint, the explicit-name
`--reclaim` adoption running a real command on the VM, the unchanged claimed path, the direct
`Get` path by machine name, and the clean not-found. Endpoints redacted; lease destroyed.

The P3 changelog item is deliberately skipped, with reasoning in the PR body: crabbox's own
`AGENTS.md` does not reserve `CHANGELOG.md` for release work, and the file already carries
unreleased entries from normal PRs.

PR #1549, opened 2026-08-27
fix(daytona): preserve dependencies and enforce sandbox lifetimes

steipete:
## What Problem This Solves

Fixes issues where direct Daytona users lose installed dependencies during an ordinary resync, expose HTTP previews without opting in, or see stopped sandboxes reported as ready. Quiet commands could also be stopped for inactivity, failed startup could leak an allocation, and the requested hard TTL and heartbeat idle-timeout replacement were not enforced by Daytona.

## Why This Change Was Made

Both direct allocation paths now share one provider-owned lifecycle implementation while retaining the documented SDK/toolbox command transport. It records exact ownership before readiness polling, reconciles a lost create response by the unique sandbox name and lease ownership, and performs bounded rollback without allocating twice. Cleanup retains a recovery claim when deletion cannot be confirmed, and retries wait for an already accepted destruction.

New sandboxes use private previews, native wall-clock TTL, and provider auto-stop intervals. Active sync and execution refresh provider activity; explicit heartbeat replacements update the provider policy. Status uses live provider state. Archive sync reuses the existing safe manifest-pruning logic to remove deleted source files without deleting remote-only dependencies or caches. API, toolbox, and streamed-upload clients share the same redirect-origin protection.

## User Impact

Kept sandboxes can be reused without reinstalling dependencies on every sync. New previews require authentication, lifecycle settings are applied at the provider, and failed allocation or deletion produces an actionable recovery outcome. No new credentials or config migration is required. Streaming, artifact collection, desktop integration, and other additional Daytona capabilities remain outside this change.

## Evidence

Regression tests cover private/native-TTL create requests, delayed allocation visibility, exact ownership rejection, rollback and retained claims, deletion already in progress, stale readiness labels, heartbeat policy failures, activity-loop cancellation, real archive resync, unsafe deletion paths, and redirects. Control-plane tests retain Windows coverage; the two tests that execute Linux toolbox shell commands locally run on POSIX hosts.

Real Daytona tests used dedicated small sandboxes and verified:

- A dependency installed into a remote virtual environment survived resync while removed source files disappeared.
- A silent command completed after 142 seconds with a one-minute idle timeout.
- A sandbox created with a one-minute hard TTL disappeared after 67 seconds despite activity refreshes.
- Heartbeat changed the real auto-stop interval, and status reported a provider-stopped sandbox as not ready.
- An injected lost response after a successful provider allocation failed the command and deleted the single owned allocation without retrying create.
- An independent source-blind check received HTTP 401 anonymously and HTTP 200 with the expected fixture using the preview token. The API key was never sent to the preview host.

All seven test sandboxes were confirmed deleted. One provider deletion remained in progress for several minutes; the CLI reported its bounded timeout, retained the claim, and subsequently cleaned the claim after the API confirmed absence.

All Go packages passed local race tests before the base update; the large CLI package passed with a 15-minute limit after exhausting an initial five-minute budget. Focused race tests passed on the updated base for Daytona, manifest pruning, heartbeats, and claim transactions. Vet, dead-code checking, and the documentation build passed. Final structured Codex review completed with no accepted or actionable findings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 ### Fixed
 
+- Made direct Daytona sandboxes private, preserved dependencies across syncs, enforced native TTL and idle heartbeats, reported authoritative readiness, and verified allocation rollback and credential-safe redirects.
 - Fixed brokered Windows bootstrap on images with built-in OpenSSH by sharing the CLI's installed/system/PATH command resolution; centralized common bootstrap fragments, pinned downloads, and portable OS metadata across both runtimes.
 - Unified E2B, Modal, and Cloudflare Sandbox run retention, bounded cleanup, and final timing; preserved command exit codes on cleanup failure, applied Modal keep-on-failure to setup/sync failures, and checked Modal archives before creation with staged workspace replacement.
 - Required exact host/project-scoped Semaphore job ownership claims and fresh provider verification before stopping jobs.
--- a/docs/features/daytona.md
+++ b/docs/features/daytona.md
@@ -130,10 +130,16 @@ crabbox stop --provider daytona swift-crab
 - **`warmup`** creates a Daytona sandbox from `daytona.snapshot`, waits for it to
   become ready, records Crabbox labels, then prints a normal Crabbox lease ID and
   slug.
+  Both direct creation paths use private previews and Daytona's native hard TTL.
+  Allocation is recorded before readiness polling; failed startup triggers
+  ownership-checked cleanup and preserves the recovery claim if cleanup fails.
 - **`run --id`** resolves a Daytona sandbox, uploads a Crabbox sync-manifest
   archive through Daytona toolbox file APIs, extracts it in the sandbox, and
   executes the command through Daytona toolbox process APIs. The command transport
   is Daytona's SDK — not direct SSH.
+  Resync preserves installed dependencies and remote-only files, deleting only
+  source paths removed from the sync manifest. Periodic activity refreshes keep
+  quiet commands alive until their command deadline or the hard sandbox TTL.
 - **`list`** and **`status`** discover sandboxes only when Daytona labels bind
   them to the Daytona provider and a canonical Crabbox lease. Direct IDs with
   missing or mismatched ownership labels are rejected.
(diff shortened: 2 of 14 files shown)


steipete:
Maintainer decision: private-by-default previews are intentional and approved as the security fix in this PR. Unconditional anonymous exposure was the reported bug, so preserving it as an implicit compatibility fallback would defeat the requested change. This affects newly created direct sandboxes; this patch does not change the visibility of existing sandboxes.

For consumers upgrading from anonymous access, obtain the standard preview link and token from Daytona and send the preview token in `x-daytona-preview-token`. Obtain a fresh token after a sandbox restart. For browser sharing without custom headers, use a Daytona signed preview URL with an explicit expiry. Neither flow requires putting the Daytona API key in the preview request. These are Daytona's existing [authenticated preview access paths](https://www.daytona.io/docs/en/preview/), not a new Crabbox preview command.

The review's `pnpm install` failure happened before Crabbox was executed. Actual post-fix provider evidence from this task follows; credentials, preview URLs, and resource IDs are omitted.

Real dependency reuse, provider idle policy, quiet execution, authoritative status, and cleanup output:

```text
PASS real dependency installed in remote .venv
PASS resync preserved installed dependency and removed deleted source
PASS heartbeat changed real provider auto-stop to 2 minutes
PASS heartbeat restored real provider auto-stop to 1 minute; sandbox private
PASS quiet command exceeded idle timeout, elapsed=142s
PASS status reports stopped and not ready despite original ready label
PASS cleanup confirmed absent
PASS lost create response: allocated exactly one sandbox, command failed, owned allocation deleted; sandbox=<redacted>
```

Independent source-blind preview response records (the provider was still destroying after its initial delete timeout; the authenticated response still returned the unique test fixture):

```json
{
  "phase": "after_stop_while_provider_state_destroying",
  "anonymous": {
    "http_status": 401,
    "fixture_content_retrieved": false,
    "body_bytes": 184
  },
  "authenticated": {
    "http_status": 200,
    "fixture_content_retrieved": true,
    "body_bytes": 50
  }
}
```

A separate create-time TTL test used `warmup --provider daytona --ttl 1m --idle-timeout 5m`. Activity refreshes continued; the operator API returned 404 after 67 seconds. The CLI creation output reported `warmup complete total=887ms`. Final exact-ID API verification also returned 404. This test did not reset TTL through a later API call. All seven task sandboxes were confirmed absent, including the one with delayed provider destruction, and the temporary credential was removed.

The final rebased head `9c1207600651cd609b268d9606940b92ca132e6d` has passing [Go, Windows, docs and other CI checks](https://github.com/openclaw/crabbox/actions/runs/33044043961), a passing required Release Check, and clean structured autoreview. The stale generated coverage table from the first CI attempt was regenerated and verified.

nickpismenkov:
cool

PR #1562, opened 2026-08-27
fix: preserve Daytona claims on unresolved stop

steipete:
Daytona `stop` could report a successful release and erase its local recovery claim without proving that the sandbox was deleted. After any resolution failure, a fallback issued another lookup using the current endpoint and credentials, then treated a 404 as sufficient proof of absence. This also masked authorization and transport failures, and could discard the only recovery record when the current scope differed from the original one.

Remove that fallback and return the original resolution error with the claim unchanged. Normal exact-owned deletion still uses the existing ownership checks and waits for confirmed destruction before removing the claim. This adds no account metadata or configuration changes. The provider documentation and changelog describe the retained-claim behavior.

Verification:

- Added public CLI-boundary regressions for missing resources, alternate saved scopes, authorization/service/transport failures, and successful exact-owned deletion. Six cases reproduced false success and claim loss before the fix.
- `GOTOOLCHAIN=go1.26.5 go test -race ./internal/providers/daytona -count=1` passed; provider vet, full CLI build, docs-link checks, and isolated Codex autoreview passed.
- The rebuilt executable passed all eight loopback API scenarios, including exact claim-byte preservation and exactly one verified deletion.
- A live Daytona check against a previously deleted, task-owned sandbox returned exit 4 with the original missing-sandbox diagnostic and preserved the claim bytes. Independent provider GETs returned 404 before and after. This test allocated no new resources; its temporary credential and claim state was removed.

This PR is limited to claim retention during `stop`. It does not change command cancellation, workspace sync, Tailscale, idle-timeout overrides, or JUnit collection.

### Captured real Daytona behavior

The following is a redacted excerpt of the captured API/CLI result for the exact source tree committed as `e6b4576d9fdaee3b50be346d2e208127d4422f0a`. The binary was built from that frozen tree before committing. This is actual Daytona service evidence, not the loopback fixture: a previously deleted, task-owned sandbox was independently confirmed absent, then an isolated recovery claim was supplied to the rebuilt CLI. Resource identifiers are redacted; no credential values are included.

```json
{
  "binarySHA256": "d9beb3ed0a7ac0e3ac7d584a158f11a99c4ace5ab669d5abe6b18c2d9fd659d5",
  "newAllocations": 0,
  "preflightExactGET": 404,
  "exit": 4,
  "claimUnchanged": true,
  "claimSHA256Before": "896a041bd3307265c57df42d503f98b37c614b2f858beb45429893afee8e0c9c",
  "claimSHA256After": "896a041bd3307265c57df42d503f98b37c614b2f858beb45429893afee8e0c9c",
  "postflightExactGET": 404,
  "passed": true,
  "temporaryCredentialAndClaimStateRemoved": true,
  "stderr": "daytona claim <task-lease> is bound to missing sandbox <task-sandbox>\n"
}
```

The intended compatibility decision is explicit: unverified absence, including after native TTL, must remain an error with a retained recovery claim. Restoring success would reintroduce the unverified deletion claim. The normal exact-owned deletion path remains unchanged and passes the separate executable loopback fixture (exit 0, one DELETE, subsequent observation, claim removed). This PR does not claim a new real-service successful-deletion run.

The automated review's attempted test did not reach the Go regression: its environment failed in a `pnpm install` setup step. The actual targeted race test, executable fixtures, and real Daytona result above were run separately and passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 ### Fixed
 
+- Preserved Daytona recovery claims and lookup errors when `stop` cannot verify the sandbox, instead of reporting release from an unverified not-found response.
 - Fixed portable Node coordinator control heartbeats deadlocking subsequent lifecycle operations, releases, and graceful shutdown.
 - Made direct Daytona sandboxes private, preserved dependencies across syncs, enforced native TTL and idle heartbeats, reported authoritative readiness, and verified allocation rollback and credential-safe redirects.
 - Fixed brokered Windows bootstrap on images with built-in OpenSSH by sharing the CLI's installed/system/PATH command resolution; centralized common bootstrap fragments, pinned downloads, and portable OS metadata across both runtimes.
--- a/docs/providers/daytona.md
+++ b/docs/providers/daytona.md
@@ -161,6 +161,9 @@ deletes the sandbox. `heartbeat --idle-timeout` changes the provider's auto-stop
 policy as well as Crabbox metadata. Status readiness comes from Daytona's live
 state, never a previously stored `ready` label. Explicit stop and rollback wait
 for confirmed provider deletion, with a bounded cleanup deadline.
+If `stop` cannot resolve the claimed sandbox, it returns the lookup error and
+preserves the local recovery claim. A missing sandbox in the current account or
+API endpoint does not prove deletion in the original scope, even after native TTL.
 
 API, toolbox, and archive-upload clients refuse redirects that change scheme,
 host, or effective port. Custom endpoints require HTTPS except for loopback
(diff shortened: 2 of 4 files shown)


steipete:
@clawsweeper re-review

The PR body now includes a redacted excerpt of the actual Daytona API/CLI result for the frozen source tree committed as e6b4576d9fdaee3b50be346d2e208127d4422f0a: independent GETs returned 404 before and after, the rebuilt CLI returned exit 4, and the recovery claim hashes remained identical. No new cloud resources were allocated. The separate executable loopback tests cover the unchanged successful deletion path.

The intended compatibility contract is explicit: unresolved absence, including after native TTL, returns the original lookup error and retains the recovery claim. The fix deliberately removes the prior unverified success result. No account-scope inference or new deletion authority is introduced.

All CI checks are now green after retrying the checksum-service HTTP/2 failure. The source head is unchanged. Please review the added runtime evidence and the documented contract; this is a review-only request.

PR #1564, opened 2026-08-28
fix: save failure bundles from unwritable directories

steipete:
## Summary

Automatic failure-bundle collection fails when Crabbox runs from a read-only or otherwise unwritable working directory, losing the diagnostics needed to investigate the original command failure.

Keep `.crabbox/captures/` as the normal project-local destination. When creating the destination fails specifically because it is unwritable, save the bundle in the existing private Crabbox user-state directory and report its actual path. If both destinations fail, report both errors without changing the original command exit status.

The fallback does not apply to ownership, symlink, file-type, or privacy-validation failures, nor to archive/read errors. Explicit stdout/stderr capture paths are unchanged. No new flags, credentials, dependencies, or provider-specific behavior are introduced. Documentation describes the state location and retention, and the changelog records the fix.

## Directory ownership and review follow-up

The failure-bundle owner retains the verified directory through file creation, publication, writing, and cleanup. Unix operations use `openat`, `renameat`, and `unlinkat`. Windows retains ancestor handles without delete sharing, performs relative NT access checks instead of the failing directory `ReOpenFile` path, and publishes/deletes through the original file handle with a retained destination directory. Pathnames are not reused as mutation authority after directory verification.

Existing write access is checked without creating a temporary name. The destination is made private before file materialization; an already-unwritable directory is not made writable as a side effect. Incomplete bundles are removed through the retained owner, and cleanup errors are reported.

Regressions cover parent-directory replacement, temporary-file substitution, successful and failed cleanup, original command exit status, Unix mode/macOS ACL protection, and Windows fallback without changing an unwritable directory's DACL. The Windows CI job explicitly runs the native capture and private-output tests.

## Verification

- Independent focused race run: **84 tests including subtests passed, zero failures or skips**. Includes actual permission-denied fixtures and deterministic directory-replacement coverage.
- `go vet ./...`, Windows-targeted vet, and `scripts/check-docs.sh` passed. The docs gate includes 14 passing tests plus command, provider, and internal-link validation.
- Codex P0/P1 review of the retained-directory implementation completed with no accepted/actionable findings.
- Linux amd64 and Windows amd64 CLI test binaries cross-compiled successfully; cross-compilation is not native execution.
- Earlier native Windows runs exposed an incorrectly owned test fixture and the directory `ReOpenFile` access failure. The fixture now explicitly establishes current-user ownership; the capture owner no longer uses that reopening path. The native failure-capture/private-output step now passes on `bc470e69fcc6d8d46ef1060de2a9c08041558982` in [CI run 33143450443](https://github.com/openclaw/crabbox/actions/runs/33143450443). The remaining exact-head checks must also pass before merge.
- A prebuilt CLI `run --help` smoke completed successfully with isolated home/config/state directories and no credentials. The earlier review's 30-second `go run` timeout occurred during cold dependency downloads rather than CLI execution.

No hosted worker leases or production services were used or changed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,6 +105,9 @@ jobs:
       - name: Test pond mesh cancellation
         run: go test ./internal/cli/ -run PondMeshCancel -count=1 -v
 
+      - name: Test native Windows failure captures and private outputs
+        run: go test ./internal/cli/ -run '^(TestFailureBundle|TestNativeWindowsFailureBundle|TestPrivateRunOutput|TestManagedAttestKeyWindows|TestArtifactOutputWindows)' -count=1 -v
+
       - name: Test Windows SSH and rsync transport selection
         run: go test ./internal/cli/ -run '^(TestWSL2TransportStagesWithoutWindowsAutomount|TestWorkspaceOwnerWSL2InputStreamPreparationIsBoundedAndReplayable|TestWindowsWSLNativeToolProbeUsesDirectCommandOutput|TestWindowsWSLNativeToolPathsRejectsWindowsShims|TestDirectSSHExecutableSelection|TestSSHCommandContextUsesNativeWindowsOpenSSHEvenWhenPATHSelectsMSYS2|TestResolveWindowsNativeRsyncPairUsesExactSibling|TestResolveWindowsNativeRsyncPairFailsClosedWithoutSibling|TestBindWindowsNativeRsyncSSHPreservesOwnedArguments|TestWindowsNativeRsyncCommandUsesSiblingPairAndEnvironment|TestResolvedSSHCopyCommandUsesNativeWindowsSiblingPair|TestResolvedSSHCopyCommandFailsClosedWithoutNativeWindowsSibling)$' -count=1 -v
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 ### Fixed
 
 - Unified provider-owned routing for stop, retry, rescue, and WebVNC commands, preserving scope, explicit false release settings, and Kubernetes environment selectors without exposing URL credentials.
-- Saved automatic failure bundles in private user state when the project capture destination is unwritable, preserving security checks and the command exit status.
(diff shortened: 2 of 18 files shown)


steipete:
Addressed the directory-ordering finding in cd5c136f. The destination is now checked for existing write access without creating a name, then secured before any bundle file is materialized. Windows additionally denies delete sharing while the file is open and publishes through that original handle rather than a source pathname.

Added native Windows tests that attempt deletion/substitution, deliberately pass a decoy source name, verify that the original handle is published, and check fallback without changing an unwritable directory's DACL. The Windows CI job now explicitly runs these tests and related private-output coverage. Unix permission and macOS ACL tests cover the corresponding ordering.

Local focused race tests passed 66/66 with no skips, full Go vet and docs validation passed, and the corrected security boundary passed independent Codex P0/P1 review. The prebuilt CLI's `run --help` also exited successfully in 1.488 seconds with isolated home/config/state directories; the previous live-check timeout was during cold dependency downloads.

The updated PR body records proof and the new exact-head native Windows CI requirement. @clawsweeper re-review

steipete:
The parent-directory finding is addressed in bc470e69fcc6d8d46ef1060de2a9c08041558982. `failureBundleOutput` retains the verified directory for the complete capture lifetime. Unix creation/publication/cleanup use `openat`/`renameat`/`unlinkat`; Windows retains traversal handles without delete sharing and uses relative NT operations and the original file handle. The capture code no longer re-resolves a pathname after directory verification or uses directory `ReOpenFile`.

The **native Windows failure-capture/private-output test step has passed** on this exact head in [CI run 33143450443](https://github.com/openclaw/crabbox/actions/runs/33143450443). It exercises permissive-directory hardening, parent-directory replacement attempts, temporary-file deletion/substitution, rooted publication, cleanup/handle release, and fallback without changing an unwritable directory's DACL. Independent local race coverage passed 84 cases with no failures/skips, full Go vet and docs checks passed, and the retained-owner change passed Codex P0/P1 review.

The full CI run is still finishing, so merging remains gated on all required checks. @clawsweeper re-review

PR #1609, opened 2026-08-29
feat(incus): replay-safe leases and native container checkpoints

steipete:
## What Problem This Solves

Incus leases could not participate in replay-safe external orchestration: `warmup --lease-id` was rejected, and retrying an interrupted allocation could not adopt a durable request identity. Incus also had no native checkpoint/fork implementation, so prepared Linux environments could not be reused through Crabbox's existing checkpoint commands.

## Why This Change Was Made

Extend the existing Incus provider, rather than add a second orchestration layer. Allocation uses the shared fixed-lease owner and binds the request to the actual daemon, project, configuration, instance name, and UUID. Native container checkpoints publish private, non-updating, non-expiring images that survive deletion of the source instance.

Persist capture and deletion recovery identity before remote effects. Forks replace managed SSH and machine identity while stopped, before any inherited credentials can be exposed. The canonical checkpoint store serializes capture/deletion and refreshes usage under the same record lock, preventing a late writer from resurrecting a deleted checkpoint. Ordinary leases retain their existing autostart policy.

The interrupted-allocation live test also exposed a shared CLI boundary bug: provider-level acquisition unlocked before registration, so a concurrent replay could change the claim and the losing caller could delete the successful caller's lease. A per-fixed-ID, cancellable cross-process lock now spans CLI acquisition and preparation; exact claim-generation checks remain unchanged. Fixed-ID post-acquisition failures preserve the adopted lease, matching the existing checkpoint-fork policy. The shared acquisition owner now carries its committed claim snapshot rather than having callers reread a possibly newer generation. Checkpoint inventory also tolerates unpublished reservations and concurrent removals while still rejecting corrupt published metadata; creation no longer writes the same record twice.

## User Impact

- Retry the same fixed Incus lease ID without allocating another instance; conflicting intent or ownership fails explicitly.
- Capture prepared container root-disk state, delete the source lease, and fork independent environments with fresh access identity.
- Verify and retry cleanup after interrupted creation, publication, and deletion without losing the recorded resource identity.
- Conflicting checkpoint deletion/pruning reports a busy operation instead of corrupting its record.

Native checkpoints/forks are deliberately container-only in this increment. Ordinary VM leasing remains supported, but native VM snapshots are not advertised as safe forks. Memory/process state, attached non-root disks, mounted workspaces, and arbitrary application-credential rotation are outside the native capture contract. No new dependency, user configuration option, or agent-orchestration behavior is introduced.

## Evidence

Passed local validation:

```sh
go test -race -count=1 ./internal/cli -run 'Test(FixedAcquisition|FixedLeaseCreateIntent|Warmup.*Failure|WarmupFailedFixed|Checkpoint|ClaimAcquiredLease|AdapterRegistrationRejects)'
go test -race -count=1 ./internal/providers/incus ./internal/providers/localcontainer ./internal/providers/aws ./internal/providers/machine0
go vet ./internal/cli ./internal/providers/incus ./internal/providers/localcontainer ./internal/providers/aws ./internal/providers/machine0
bash scripts/check-docs.sh
GOOS=linux GOARCH=arm64 go build -trimpath -o /tmp/crabbox-incus-linux ./cmd/crabbox
go build -trimpath -o /tmp/crabbox-incus-macos ./cmd/crabbox
```

The final inventory repair additionally passed the full checkpoint/fixed-acquisition subset under `-race` and the Incus provider suite. Codex autoreview of the complete candidate is clean through P2. A mechanical rebase brought in current main's independent SSH/egress fixes; the Incus and checkpoint changes were unchanged. Exact-head CI is the final full-suite gate. Earlier full local suite attempts exceeded Go's default ten-minute `internal/cli` package budget; they are not claimed as passing.

Real Incus 7.4 container proof passed with the repository's `scripts/live-incus-checkpoint-smoke.sh`: normal sync/run, same-ID replay, capture without changing the source identity, deletion of the source before fork, retained workspace/dependencies, fresh machine/SSH identity, strict-host-key SSH probes rejecting the source private key, checkpoint deletion while the fork remained usable, and cleanup. Source `cbx_b059d4a092f7`, fork `cbx_94ba9d6995a8`, checkpoint `chk_0f69a25baf05cdaf`.

Additional process-fault proof passed: SIGKILL the CLI after the instance starts, confirm ordinary run refuses incomplete acquisition, then run two concurrent identical warmup replays and verify the original instance name/UUID. SIGKILL capture during the daemon's image-publication operation, recover the available private/non-updating/non-expiring image from the durable pending record, delete the source, and fork twice. Both forks retained a workspace marker and a 128 MiB disk payload; the source key and each sibling's key were rejected, and changing one fork's marker did not affect the other. Both forks remained usable after checkpoint deletion. Source `cbx_1b0ea12a1a76`; forks `cbx_7a0eb296f08a` and `cbx_5cd6505bf40f`. Final direct daemon inventory: **0 instances and 0 checkpoint images**.

Net change: production +1,282 lines; tests +1,393; docs/proof +262. Production growth implements two provider capabilities plus the ownership, offline credential replacement, durable recovery, and shared CLI/checkpoint concurrency boundaries; it reuses the existing fixed-lease and native-checkpoint APIs, removes the old Incus allocation path, and removes duplicate claim/metadata publication.

Before-change binary proof: `warmup --provider incus --lease-id cbx_aabbccddeeff --class container` exited 2 with `provider=incus does not support fixed idempotent lease IDs`.

Regression tests failed before the corresponding fixes for interrupted capture metadata, incomplete-allocation deletion reply loss, recovered-image deletion reply loss, premature incomplete-fork startup, inherited image expiry, concurrent capture/delete record resurrection, and ordinary-lease autostart overrides.

Dependency contracts were inspected in the pinned Incus v7.1.0 client/source and current upstream source. Live testing uses a task-owned Ubuntu 26.04 ARM64 VM with signed Incus 7.4 and ZFS 2.4.1; default confinement is retained. The distro Incus 6.0.5 package initially had an AppArmor stacked-profile signal mismatch, independently reproduced with unchanged Crabbox and resolved by upgrading to current upstream Incus.

VM live-execution limitation: the default community Ubuntu VM image stops at `error: prohibited by secure boot policy.` A stock Incus control VM reproduces this without Crabbox. Allocation and failure cleanup were exercised, but successful VM execution is not claimed; Secure Boot was not disabled. Native container lifecycle proof is independent of that limitation.

Release-note context: Incus gains replay-safe lease allocation and native container checkpoints/forks with pre-boot identity replacement and durable cleanup recovery.


steipete:
Landing evidence for head `40021f699a1f53fc52db20f08bcc2110b5ecd169`:

- [Full CI](https://github.com/openclaw/crabbox/actions/runs/33234499284) passed, including `go test -race ./...`, all Go modules, core coverage, vet/deadcode, builds, Windows, Apple VM, Worker, scripts, and docs. Connector E2E smokes, CodeQL, Docs UI Proof, and the required release-policy check also passed.
- The exact focused local commands and real Incus 7.4 native/fault-injection results are recorded in the PR body. Recovered checkpoint `chk_9eb3d9c923f5c719` produced two independent forks; strict SSH probes rejected source and sibling keys, and both forks survived checkpoint deletion. Final daemon inventory was zero instances and zero checkpoint images; the temporary source-key copy was removed.
- Fresh Codex autoreview of the complete candidate is clean through P2. The final rebase changed only changelog context in our patch; the implementation and regression tests were preserved.

ClawSweeper's remaining scope question is resolved by retaining the documented native-checkpoint contract: root-disk application data, including application credentials, is intentionally captured. Images stay private and ownership-bound; managed SSH and machine identity are replaced before a fork starts. This is not an arbitrary-application secret scrubber, and the provider/security documentation says so explicitly. No security default is weakened to hide that distinction.

The live tests found and fixed two shared boundaries, not just provider symptoms: fixed-ID CLI calls now serialize through registration/preparation without weakening exact claim-generation checks, and checkpoint inventory distinguishes published records from incomplete reservations while still rejecting corruption. Failed fixed-ID replays preserve adopted leases. The remaining VM Secure Boot/image limitation is documented and independently reproduced; native VM checkpointing is deliberately unsupported rather than booting inherited credentials.

Issue #1391, opened 2026-08-17
feat(run): download selected evidence from failed SSH runs


coygeek:
## Summary

Add an explicit opt-in way to retrieve selected remote evidence files when an SSH-backed command exits nonzero. Keep the current `--download` success-only behavior as the default.

## Problem to solve

Failed tests, validators, and model or browser workflows often write their most useful structured report before returning a nonzero status. Crabbox currently skips every `--download remote=local` request when the remote exit code is nonzero, even when the requested file exists and the lease is about to be destroyed by `--stop-after always`.

This was reproduced with an exact current-main binary on a disposable Tart macOS lease. The command wrote `reports/failure-proof.json` and exited 23. Crabbox created its normal failure bundle, stopped the lease, returned 23, and did not create the requested local download file.

The behavior matches the documented contract: downloads run only after a successful remote command. The missing failed-run retrieval path is therefore a feature gap rather than a bug in the existing flag.

## Proposed behavior

Provide an explicit failed-run download policy for bounded, caller-selected files. The exact CLI syntax is a product choice; examples include a separate `--download-on-failure remote=local` flag or an `on=success|failure|always` policy attached to download specifications.

The policy should:

- attempt only the caller-selected remote files;
- preserve the workload's original nonzero exit status and failed-run classification;
- use the same path-collision preflight, containment rules, owner-only local permissions, and SSH target support as normal downloads;
- make retrieval failures visible without silently converting the workload failure into success; and
- run before a configured stop policy destroys the lease.

## Acceptance criteria

- A failing SSH-backed command can write a report, exit nonzero, and have that explicitly selected report retrieved locally before lease teardown.
- The Crabbox process still reports the original workload failure; successful evidence retrieval does not turn the run green.
- Existing `--download` specifications remain success-only unless the caller selects the new failure policy.
- A missing or unreadable failure-evidence path produces a clear retrieval diagnostic while retaining the original command exit and failure context.
- Local destination collisions are rejected before remote execution, as they are for current captures and downloads.
- Retrieved files and newly created output directories keep Crabbox's current owner-only permission contract.
- POSIX, WSL2, native-Windows, and macOS SSH targets receive coverage appropriate to the existing single-file download transport.
- Delegated providers remain unchanged unless they advertise and implement the corresponding bounded failed-run retrieval capability.
- Regression coverage includes `--stop-after always` and proves retrieval occurs before the lease is removed.

## Affected area

- `crabbox run` evidence and download policy.
[...]

PR #1631, opened 2026-08-29
feat(run): download evidence after confirmed Linux SSH failures

steipete:
Implements the approved ordinary Linux SSH first slice of https://github.com/openclaw/crabbox/issues/1391.

Failed workloads can write their most useful report immediately before exiting nonzero. This adds repeatable `--download-on-failure remote=local` without changing success-only `--download` behavior.

Retrieval requires a fresh Crabbox-owned workload-start/exit marker pair and a clean SSH completion. The wrapper preserves the inner login shell and its exit status, including workload exit 255; transport loss, setup failures, cancellation/deadline, and post-success JUnit policy failures do not authorize retrieval. The observed command cannot replay on a fallback SSH port. Protocol markers are removed before stderr reaches logs or captures and leave no remote marker files.

Selected files are fetched before failure bundles and teardown through the existing single-file transport and private atomic writer. Collision preflight includes both download policies and all other output destinations. Each file has a 30-second budget; a failed retrieval warns and continues without replacing the workload exit or classification. Unsupported OS targets and delegated execution reject the flag.

Validation: focused regression and help-contract tests passed; Codex autoreview is clean. Full Linux Go/race, all-module, provider lifecycle, coverage, Worker, Node/container, generation, scripts, and docs gates passed; core coverage was 91.0%. [Final-head CI](https://github.com/openclaw/crabbox/actions/runs/33257622278) is green. The first full race run caught a test-only shared-output-buffer race, fixed in the follow-up commit; the corrected witness tests passed twenty repeated AWS race runs and ten local repetitions. A separate proof rerun also needed the standard umask restored for existing private-directory fixtures. No artifact-freshness, provider-ownership, coordinator, or credential changes.

Live proof used clean client `925638baa29667bef4866bb68c509f1fe3456b7b` against one retained AWS Linux SSH lease `cbx_eaaf28e65a11`, EC2 instance `i-0d9623b7fa2d0826a` (`c7a.8xlarge`, `eu-west-1`).

| Case | CLI exit | Run |
|---|---:|---|
| seed | 0 | `run_62185f89b12a` |
| failed-multiple | 23 | `run_613330d63f1f` |
| workload-255 | 255 | `run_a5a3ea591fdd` |
| success-skips-failure-policy | 0 | `run_d8bf4724aa6b` |
| collision-before-execution | 2 | `preflight; no run` |
| transport-loss | 255 | `run_30837045ef7c` |
| transport-file-remains | 0 | `run_e0dc68068b00` |
| stop-always | 23 | `run_e3c645e4a57d` |

The exit-23 case retrieved both existing files despite the missing middle path, left the success-only destination absent, and emitted retrieval messages before the automatic failure bundle. The workload-255 case also retrieved its selected report. A deliberate real SSH disconnect after writing a remote report did not download it; a subsequent clean run confirmed that report existed remotely and that the collision-rejected command had never executed. No command was replayed. The final exit-23 run retrieved its report before `--stop-after always` released the lease.

Retrieved files were mode 0600; newly created output directories were 0700. The protocol markers did not appear in captured output and create no remote files. Exact evidence bytes and permissions were verified and archived locally, then transient output destinations were removed. Final explicit stop and inventory verification found no owned leases.

Client SHA-256: `6279da3cda7611a9f0671b66c353e6d701127d7b8681140f6e80ac1abcce54bc`. Pre-execution coordinator lookup retries: 0. Full proof logs and output manifest are retained in the lane.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Added
 
+- Added Linux SSH `--download-on-failure` retrieval after an owned nonzero workload exit, preserving the exit code and collecting selected evidence before failure bundles and teardown.
+
 - Added opt-in Machine0 native size selection context to live catalog discovery while preserving configured defaults and exact fixed-lease replay.
 - Added durable fixed-ID Incus leases and private container disk checkpoints that survive source deletion, with ownership-checked cleanup and fresh SSH identity before fork startup.
 
--- a/docs/commands/run.md
+++ b/docs/commands/run.md
@@ -564,6 +564,30 @@ command execution. On Unix-like hosts, Crabbox-created download, capture, proof,
 and failure-bundle files use owner-only permissions (`0600`), and newly created
 output directories use `0700`.
 
+Use repeatable `--download-on-failure remote=local` to retrieve explicitly
+selected evidence after a nonzero workload exit on ordinary Linux SSH runs.
+Crabbox requires a fresh owned workload-start/exit marker pair and successful
+SSH completion; a workload exit of 255 is distinguishable from SSH transport
+loss. The observed workload is never retried on a fallback SSH port. Setup,
+sync, hydration, acquisition, preflight, transport, cancellation, and timeout
+failures do not authorize these downloads. A JUnit policy failure after a zero
+workload exit does not authorize them either, nor does a failed
+`--require-artifact-change` guard. When both flags are used, a confirmed nonzero
+workload exit can download evidence while the freshness guard remains
+`not-evaluated`; the download does not claim that the file changed.
+
+Retrieval precedes the automatic failure bundle and lease teardown, including
+`--stop-after always`. A missing/unreadable file or local write failure produces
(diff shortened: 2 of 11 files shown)


steipete:
The current `9f772a30f93b82ded6719f04b22151c29352e928` head still has the lease-output collision identified in triage. `preflightRunLocalOutputs` checks captures/downloads, while `--lease-output` is checked separately for writability; it is not compared with the download destinations. `preflightAttestPaths` does not fill that gap when no receipt is requested.

For an otherwise valid retained run (`--keep --stop-after never`), using the same local destination—or an alias—for `--lease-output` and `--download-on-failure` lets the later report replace the retained session JSON. The success-download path shares the omission.

I’m repairing the missing common preflight check using the existing canonical path/alias comparison, with regressions requiring exit 2 before acquisition and unchanged existing bytes for both download policies. Please keep this head stable and hold the merge until that repair and proof are in. The authoritative workload-exit, JUnit policy, retrieval, bundle and teardown ordering will remain unchanged.

PR #1704, opened 2026-08-31
feat: carry requested fixed lease IDs through delegated warmup

zozo123:
## What Problem This Solves

`crabbox warmup --lease-id <id>` already validates the flag and takes the fixed-acquisition lock for every backend,
but only the SSH-lease path forwards the ID. The delegated branch silently drops it. So a delegated provider that
advertises `IdempotentLeaseIDBackend.SupportsRequestedLeaseID` passes the capability check and is then handed
nothing to be idempotent about — the caller's fixed ID disappears without a warning.

## Why This Change Was Made

`WarmupRequest` now carries `RequestedLeaseID`, and `run.go` passes the same trimmed value it already passes to
`Acquire`. A delegated adapter can then consume it exactly as the two closest existing patterns do
(`internal/providers/external/backend.go:39-43` and the sealosdevbox equivalent).

This is deliberate enabling plumbing, not a behavior change: observable behavior is byte-identical for every
provider shipped today, because no delegated-run adapter currently advertises the gating capability. It is split
out so that the first adapter to opt in does not have to change core in the same PR. As of this commit the only
reader is the synthetic backend in the new test.

Non-goals: no adapter opts in here; the pre-existing inconsistency where warmup gates on trimmed-non-empty while
`checkpoint fork` gates on `flagWasSet` is left alone, since changing it is user-visible and belongs in its own PR.

## User Impact

No user-visible change today. It removes a silent-drop footgun for the next delegated provider that wants
replay-safe fixed lease IDs, and makes `--lease-id` mean the same thing on both paths.

## Evidence

New test file `internal/cli/warmup_fixed_lease_id_test.go` (187 lines) asserting the ID reaches a capable delegated
backend verbatim, that an incapable backend still rejects with the same error and exit code, and that an empty
`--lease-id` behaves as before.

The tests were mutation-checked rather than assumed: five separate mutations to `internal/cli/run.go` were applied
one at a time and each was confirmed to FAIL the new test before the file was restored — including deleting the
`RequestedLeaseID` assignment from the delegated `WarmupRequest` literal, and dropping the `TrimSpace`.

```
$ go test ./internal/cli/ -run TestWarmup -count=1     # ok
$ gofmt -l $(git ls-files '*.go')                       # clean
$ go build ./... && go vet ./internal/cli/              # clean
```

Note: `TestCheckpointCaptureBuiltBinaryContract` fails 4 subtests here, but fails identically on a pristine
`origin/main` worktree — pre-existing and unrelated.

### DIFF
--- a/docs/provider-backends.md
+++ b/docs/provider-backends.md
@@ -177,13 +177,25 @@ type IdempotentLeaseIDBackend interface {
 }
 ```
 
-Direct AWS, Machine0, and local-container backends implement this capability;
-coordinator-backed leases support it through the coordinator wrapper. External
-backends support it only when their configured protocol explicitly advertises
-idempotent lease IDs. `crabbox warmup --lease-id` rejects other backends before
-provisioning. Built-in direct adapters reuse `core.AcquireFixedLease` for
+Direct AWS, Machine0, Incus, and local-container backends implement this
+capability; coordinator-backed leases support it through the coordinator
+wrapper. External backends support it only when their configured protocol
+explicitly advertises idempotent lease IDs. `crabbox warmup --lease-id` rejects
+other backends before provisioning. Built-in direct adapters reuse `core.AcquireFixedLease` for
 durable intent and replay mechanics while keeping resource creation,
 reconciliation, and identity validation provider-owned.
+
+Every backend that implements this capability today is an SSH-lease backend, and
+those read the requested ID from `AcquireRequest.RequestedLeaseID`.
+`WarmupRequest.RequestedLeaseID` is the delegated-run counterpart: core now
+carries the value across the delegated warmup path, so a delegated-run adapter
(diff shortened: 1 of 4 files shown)


zozo123:
Closing this.

The review is right and the objection is structural, not fixable by editing this PR: nothing in the tree consumes `WarmupRequest.RequestedLeaseID`, so its only evidence is a synthetic backend. Every current implementer of `SupportsRequestedLeaseID` is an SSH-lease backend, so there is no real delegated-provider trace to supply.

Rather than keep a new exported field alive with no reader, the field should land in whichever change first consumes it — the islo canonical-identity work, where it can arrive with a real provider trace. #1708 already owns the identity story for that adapter.

Nothing here is lost: the diff is four files and reproducible from this branch when a consumer exists.

PR #1705, opened 2026-08-31
docs: record why the islo provider stays delegated-run

zozo123:
## What Problem This Solves

Islo's delegated-run design and cleanup semantics were not documented together, making it easy to mistake the human SSH login helper for an unattended lease transport or interpret a successful `crabbox stop` as verified deletion. The earlier PR wording also conflated an accepted DELETE response with the client's separate handling of HTTP 404 as idempotent success.

## Why This Change Was Made

This documentation-only change preserves @zozo123's original contribution: explain why Islo remains delegated-run, distinguish resource identity from name-based addressing, and document which observations Crabbox actually uses for existence and cleanup. The rationale is now scoped to the current adapter: it uses unattended exec/files/lifecycle APIs but does not provision or manage a per-lease SSH endpoint and credential lifecycle. Human SSH setup remains a separate login workflow; the docs do not claim that no Islo API could ever support SSH automation.

The cleanup wording now follows `DeleteSandbox` and `Stop`: a successful stop means the client accepted either a non-error delete response, including an accepted-but-pending 202, or a 404 as idempotent success. It then removes the local claim without a follow-up read. This is not a promise that Islo accepted a new delete or that the resource is already absent. Inventory listing is not proof of a particular sandbox's absence.

The by-id discussion is conditional: an actual tombstone matching the recorded id, with `status: deleted` and a deletion timestamp, can provide resource-specific evidence; a missing response is not a tombstone. The adapter does not retain ids for teardown verification or use that endpoint today. Unsupported UUIDv7, rename/failed-create details, and exact 503/idempotency promises were removed. The docs instead note the optional, currently unused `request_id` field and require any future retry integration to establish its semantics rather than infer that a failed create allocated nothing.

This addresses the delete-result finding in the [existing review](https://github.com/openclaw/crabbox/pull/1705#issuecomment-5482108496) while retaining the contributor's rationale and operational distinctions.

## User Impact

Operators and future adapter contributors get a clearer boundary between a client-accepted result, name lookup, and evidence about one exact resource. Runtime behavior, provider selection, configuration, credentials, and API calls are unchanged. No changelog or release changes are included. Original contribution: Yossi Eliaz (@zozo123).

## Evidence

- Checked the final two-document diff against the current Islo provider's `Spec`, SSH helper, `DeleteSandbox`, and `Stop` paths. The delete client explicitly accepts both non-error responses and HTTP 404; `Stop` does not perform post-delete verification.
- `scripts/check-docs.sh` passed on the final text: 59 commands, 81 providers, 244 Markdown files, 14 docs-site tests, and 10 Zed boundary tasks.
- `git diff --check` passed. The existing focused Islo delete-client regression passed with the race detector in 3.340 seconds.
- Managed Codex review of the combined PR documentation diff was scoped-clean at the requested P0-only threshold. That result is not a claim that a broader review scope was run.

No live Islo provider call was made for this documentation change, and no live deletion or SSH-lease behavior is claimed. Hosted CI is awaiting refreshed checks after the updated branch is pushed; previous-head checks are not presented as proof of the final text. This PR is not yet merged.

### DIFF
--- a/docs/features/islo.md
+++ b/docs/features/islo.md
@@ -274,10 +274,12 @@ with it today:
   that resolves is therefore not proof that a particular resource is live.
   Crabbox addresses sandboxes by name for lifecycle calls because it generates
   and normalizes those names itself and rejects non-Crabbox ones.
-- **Get-by-name is authoritative for existence.** `GetSandbox`
-  (`internal/providers/islo/client.go`) is what `status`, SSH resolution, and
-  lease reuse consult, and its `404` after a delete is immediate and
-  authoritative.
+- **Get-by-name is what the adapter treats as authoritative for existence.**
+  `GetSandbox` (`internal/providers/islo/client.go`) is what `status`, SSH
+  resolution, and lease reuse consult, and a `404` from it is how the adapter
+  concludes that a name does not resolve. It answers a question about a name at
+  one point in time; it is not a guarantee about a delete that was only
+  accepted.
 - **The list endpoint is eventually consistent and must not be used to prove
   absence.** It can keep returning a sandbox for seconds after that sandbox's
   own `GET` reports `404`. Crabbox calls `ListSandboxes` only for inventory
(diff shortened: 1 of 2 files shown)


zozo123:
@clawsweeper re-review

Blocking findings addressed and the PR body updated with the fixes plus live behaviour proof.

PR #1710, opened 2026-08-31
feat(checkpoint): dispose of sources held by unresolved captures

steipete:
<details>
<summary>Additional instructions</summary>

Keep **Allow edits from maintainers** enabled.

</details>

## What Problem This Solves

An interrupted ordinary native capture can leave no image identity while still holding its source VM. Normal stop and checkpoint deletion correctly refuse to discard that uncertainty, but operators have no canonical way to dispose of the source while preserving the possible image-storage obligation.

## Why This Change Was Made

Adds `checkpoint abandon <checkpoint-id> --id <canonical-lease-id> --expected-provider-resource-id <immutable-source-id>`, initially supported by Machine0. Admission must positively observe the exact source in the selected account and match its unchanged local claim. The existing checkpoint journal records a separate source-cleanup account binding, then the existing claim-bound release owner performs disposal. Replay finishes only after exact source absence and claim finalization.

The terminal phase is `abandoned`, and the original image submission remains unresolved. Source-cleanup provenance never becomes image authority. The command neither submits another image nor starts, adopts or deletes one; fork, restore, deletion and prune keep refusing the retained record. No new registry, configuration or duplicate provider deletion path is introduced.

## User Impact

Operators can deliberately dispose of a held source without erasing the unresolved snapshot obligation. `inspect --verify` reports `reconcile_image` after source disposal; the journal must remain for investigation of the original provider operation. Other providers reject abandonment until they implement the source-attestation capability.

## Evidence

- The built-binary baseline rejected the new recovery command. All 17 candidate cases passed, covering wrong source/account/generation, another unresolved checkpoint, process death after binding or removal, a lost ordinary-save response, late images, terminal replay and retained image uncertainty.
- Full checkpoint/Kong CLI race tests passed in 218.107 seconds; the complete Machine0 provider race suite passed in 45.405 seconds. Focused account/provenance and final binding/abandonment race tests, vet, build and diff checks passed.
- The candidate wrote and completed an abandonment record; the immutable `26b735` older binary then rejected fork, restore, delete, local-only deletion and prune without changing that journal. The candidate reopened and replayed it with exactly one source removal and no new saves or starts.
- Independent owner review has no outstanding findings. Managed Codex review was scoped-clean at P0; its one later recovery-guidance wording correction was independently checked and covered by the final focused tests.
- Production net +275, tests +515, documentation/changelog +36. Growth supplies explicit source-only recovery with current identity/account binding and retained image uncertainty across the existing ownership boundaries.

The tested binary was built from `91f6710b` plus the frozen patch before commit. All fourteen changed files match committed source `27da5f081d8dde35faa5fe6f81c59bb12da557ad`; the binary retains its pre-commit build metadata. The same frozen binary then recovered the task-owned live Machine0 source that triggered this work. Two identical canonical calls returned `retiring` (10.406 seconds) and `abandoned` (2.056 seconds). Independent account-bound inventory confirmed the exact source VM absent; the canonical claim became its released tombstone. The retained audit still reports `unresolved_capture` / `reconcile_image`, with no known image identity. All recovery processes settled, and all original journals and recovery evidence remain. No image call, recapture, replacement allocation or claim/journal rewrite outside the canonical owner occurred. Source disposal does not prove image absence.

Two supplementary local audit assertions initially assumed that a released claim retained its live resource IDs and scope. Source inspection confirmed the existing compact tombstone clears those IDs and retains its intent scope; the corrected local audit passed. The canonical live wrapper was unchanged and passed on its sole run.


steipete:
Maintainer disposition: accept the documented source-only recovery command and downgrade boundary. Source disposal retains the unresolved image journal and does not assert image absence or authorize image deletion.

The final ancestry refresh at `761eb9e8227ae7068e576995626b970e1f85ef29` has the same complete Git tree (`0cfb429790c485f3de6869a48c5cb2532a99448d`) as the reviewed and tested `76101b1b605f998c5a61093f29c4d238d892390c`. Combined capture/abandonment binary-contract race tests passed. The required release workflow will pass before landing through the existing maintainer bypass; repository protections remain unchanged.

PR #1723, opened 2026-09-01
fix: keep managed Mac bootstrap passwords private

steipete:
## What Problem This Solves

Fixes an issue where operators provisioning a managed Mac desktop could have its generated account password copied into bootstrap logs and process arguments. The password file also initially inherited the caller's umask. AWS's initialization daemon records user-data stdout and stderr even when bootstrap succeeds ([source](https://github.com/aws/ec2-macos-init/blob/48b8f5d3f9927e6a5324a699ab6186cc1a59d29e/lib/ec2macosinit/userdata.go#L58)).

## Why This Change Was Made

The shared macOS recipe disables tracing, creates the credential file with mode 0600, and supplies the same initial password-setting command through stdin. The canonical generator updates both the Go CLI and TypeScript coordinator. Command failures remain visible and stop bootstrap; there are no authentication, Secure Token, retry, configuration, or provider-policy changes.

The VNC documentation also distinguishes SSH access through loopback from the native listener's possible other interfaces and the security-group boundary. Handwritten production code grows by three lines for explicit tracing shutdown and the two security-contract comments; generated projections have no net line growth.

## User Impact

New managed Mac bootstraps keep the password private at its creation and avoid disclosing it through tracing or the account tool's arguments. Existing instances are not modified by this patch. Credentials potentially exposed by earlier bootstrap should be rotated through the normal authenticated account change flow, keeping the stored credential consistent with the account.

## Evidence

- Separate pre-fix reproductions failed on password-bearing stderr, password arguments, and mode 0644 when the account tool consumed the file.
- The complete generated shell executes under temporary paths with inert system commands. Normal and inherited-trace callers cover fresh creation, password reuse, mode 0600 before consumption, no password arguments or output, readiness completion, and visible command failure.
- `node --test scripts/generate-bootstrap.test.mjs`: 13/13 passed on the refreshed candidate, including compiled Go/TypeScript parity, shell and PowerShell execution, and stale-generation checks.
- `go test -race ./internal/cli -run '^(TestAWSUserDataMacOSProfile|TestSharedBootstrap.*)$' -count=1`: passed. Generation, syntax, diff, docs links, and docs build checks passed. Managed Codex review completed with no accepted/actionable findings.
- Native macOS 15.7.9 on the existing managed AWS Mac: created an isolated ordinary non-admin account (UID 1000), ran the generated first-password setter with real `dscl` through non-TTY stdin, and authenticated the generated credential using real `dscl -authonly`. The file was mode 0600 and its value appeared in neither captured output stream. The test account's Secure Token stayed disabled; no token policy was changed. Normal account deletion completed, and an independent SSH check confirmed the exact account and private home/state directory were gone. The existing desktop account and its enabled Secure Token were unchanged.

The native proof completed at **2026-09-01T15:49:27.886598Z** against head `dc4fa1abcf10c627f47c9bc77ee488516143fe7c`. It used the complete generated script with host-global paths mapped into the private fixture directory and service/readiness commands made inert. Password generation, filesystem operations, `dscl` initial setting, and credential authentication were real. This closes the fresh native setter/prompt-sequence gap raised in review.

```json
{"sshExit":0,"stage":"complete","nativeInitialSetter":true,"nativeCredentialAuthentication":true,"privateFileMode":true,"noCredentialInCapturedOutput":true,"nonAdmin":true,"secureTokenBefore":"disabled","secureTokenAfter":"disabled","cleanupComplete":true,"fullSystemBootstrap":false,"screenSharingAuthentication":false}
```

Proof limits / Rank-up disposition: a new EC2 boot and Screen Sharing authentication as the temporary account were not exercised. Replaying host-global bootstrap on the retained, active Mac would change its SSH/services, so those steps remained isolated. The evidence proves the changed native credential setter and resulting account authentication, not a full fresh-host/VNC run. The earlier authenticated rotation of the existing desktop account was separate evidence and is not substituted for this first-setting proof. An earlier full generator run hit the existing Apple VM verifier's 60-second cold Go compilation limit; the unchanged isolated check and subsequent complete suites passed without a timeout change.

Named follow-up: the existing recipe publishes the canonical password file before the account setter succeeds. A setter failure can leave a nonempty file that a rerun skips. This preexisting publication-order issue is unchanged here and belongs to the bootstrap credential lifecycle owner.


steipete:
The fresh native setter proof is now in the PR body. On the managed AWS Mac running macOS 15.7.9, the generated recipe successfully set the first password for a new ordinary non-admin account using real `dscl` and non-TTY stdin. Independent native authentication accepted that generated credential. Its file was mode 0600, the credential was absent from captured output, and both normal cleanup and a separate SSH inspection confirmed the account and private home/state directory were removed.

This addresses the changed setter/prompt-sequence risk. For the broader Rank-up request, full host bootstrap and Screen Sharing authentication as that test account remain explicitly untested: only global paths and service/readiness commands were isolated to protect the retained active Mac. The actual credential creation, setter, and authentication were not stubbed. The existing desktop account, SSH, Screen Sharing, TCC, and token policy were unchanged. The body contains the redacted receipt and the proof boundaries; the original separate regression runs remain the before-fix evidence.

Issue #1730, opened 2026-09-02
ASCII Box release hides bounded cleanup progress for stale claims


shunkakinoki:
## Summary

ASCII Box release can spend its full bounded cleanup window without progress output, including when the native Box has already been deleted and the only remaining state is a stale Crabbox claim. To an operator, both automatic `--stop-after always` cleanup and explicit `crabbox stop` appear hung.

## Observed incident

During the 2026-09-02 incident recorded in https://github.com/openclaw/crabbox/issues/1727#issuecomment-5502666862:

1. The pre-fix run exhausted its 20-minute SSH-readiness ceiling.
2. Cleanup emitted no terminal result for more than two additional minutes.
3. After interrupting the owner, explicit `crabbox stop --provider ascii-box --target linux --id cbx_9463a02d8f1d` also emitted no progress for two minutes.
4. Native `box delete --yes --json bx_gxym9gf4` was accepted, and `box info bx_gxym9gf4 --json` returned 404.
5. A second explicit stop using the patched endpoint build still produced no progress while reconciling the stale local claim.

The release path is bounded by `boxReleaseTimeout`, so this is not proven infinite execution. The defect is that a bounded multi-minute reconciliation is operationally indistinguishable from a hang, and an authoritative native deletion does not provide a fast claim-reconciliation path.

## Expected behavior

- Emit periodic, structured release progress including the current phase and remaining deadline.
- Return promptly when an exact native-not-found result plus complete provider inventory proves the originally claimed Box is absent.
- Retain the claim when absence is uncertain, inventory is incomplete, identity changed, or provider access failed.
- Preserve cancellation and the existing exact-ownership fencing.
- Never start a replacement lease as part of cleanup recovery.

## Suggested regression coverage

- Exact `GetBox` not-found plus complete inventory absence removes only the unchanged original claim without issuing another delete.
- Not-found plus incomplete/failing inventory retains the claim.
- Replacement identity in inventory retains the claim.
- Progress output is emitted during a slow bounded release and stops after cancellation/timeout.
- `ReleaseLease` remains bounded when the Box CLI child does not settle.

Related execution-routing fix: https://github.com/openclaw/crabbox/pull/1728

PR #1731, opened 2026-09-02
fix(ascii-box): persist deletion operations for safe cleanup retries

shunkakinoki:
## Status

Merged to `main` as `67ccaeff1bc63d9d7d10ebacec678622d816c815` after maintainer-authorized landing, refreshed scoped review, and green exact-head CI for `fcdd7c52e917926404fe9ca457307b0073e3d259`. Both original source objections were rechecked against the repaired implementation and regressions and are addressed; the [obsolete changes-requested review](https://github.com/openclaw/crabbox/pull/1731#pullrequestreview-5086441381) was dismissed with that explanation. This is a high-confidence source/regression repair with bounded real-service status proof, not a claim of completed native lifecycle proof.

The original live harness PASS was rejected: official `box --json delete` returned exit zero with a pending operation, not completed deletion. The old buggy path had already removed the local claim. Box disappearance alone does not prove cleanup, and that claim has not been recreated or given fabricated evidence.

## Repair

Validate the deletion operation's ID, kind, exact Box target, known state, and completion timestamp. Initial deletion and snapshot-guard retries wait for explicit completion within the existing release budget. Pending, processing, blocked, malformed, changed, canceled, and failed observations never grant completion authority.

After valid native acceptance, timeout/cancellation/polling failure preserves the exact operation reference in the original ownership claim. Acceptance and completion use separate, domain-separated bindings covering provider scope, Box generation, repository owner, and claim content. Later stop reads that operation without SSH teardown or another deletion request; explicit completion and native absence checks must pass again inside the claim fence before removal. Endpoint refresh cannot inject evidence, and unreleased exit-zero-only witnesses are rejected rather than upgraded.

Published-claim setup rollback uses that same finalizer and original claim without nested locks or later-owner adoption. Unpublished attempts retain their separate absence guard; creation errors preserve the generation already validated during readiness. No new CLI option, dependency, credential requirement, or external receipt-admission path is added. Hard termination before durable acceptance publication remains a conservative recovery boundary.

## Verification

All four exact-head workflows succeeded: [CI](https://github.com/openclaw/crabbox/actions/runs/33616758824), [connector checks](https://github.com/openclaw/crabbox/actions/runs/33616758835), [Release Check](https://github.com/openclaw/crabbox/actions/runs/33616758857), and [docs visual proof](https://github.com/openclaw/crabbox/actions/runs/33616758836).

Red-to-green rollback and replacement-generation regressions passed. Focused tests passed in 3.395s and the provider race suite passed in 14.910s. Provider vet, full CLI build/invocation, diff checks, and independent/scoped source reviews passed. Coverage includes pending-reference persistence, blocked retries, completed-operation confirmation, cancellation, changed claims/operations, protected evidence, and old-witness rejection. A fresh root race run of the four original-objection regressions also passed in 2.100s.

```sh
GOTOOLCHAIN=go1.26.5 GOMAXPROCS=2 go test -race -p 2 ./internal/providers/asciibox -count=1
GOTOOLCHAIN=go1.26.5 GOMAXPROCS=2 go vet -p 2 ./internal/providers/asciibox
GOTOOLCHAIN=go1.26.5 GOMAXPROCS=2 go build -p 2 -trimpath -o bin/crabbox ./cmd/crabbox
```

## Bounded native observation — not lifecycle proof

At **2026-09-02 10:29:20 UTC**, a separate Go 1.27 task-only probe built against this candidate called its actual `GetDeletionOperation` through official native CLI `0.1.211-ascii-prod1`. Its exact two-command allowlist permitted only `status` and `deletion status`; no allocation, stop, delete, adoption, or claim handling occurred.

The candidate decoded operation `bdop_7097f6e8e28a4a46b8a4dd89f4ad6cc5`, kind `box`, target `bx_sxgz3k87`, as **`blocked`**, with no completion timestamp and **`completed:false`**. Test PASS establishes only successful real-service parsing/status reads. `cleanupVerified`, `boundClaimRetryValidated`, and `endToEndProof` are all **false**. The probe's isolated local auth homes were removed; this says nothing about provider cleanup. A repeat candidate read at 2026-09-02 10:51:14 UTC still reported the same blocked operation with no completion timestamp. The original resource remains unverified cleaned, with its old local claim absent. All task-owned local secret files and temporary credential/SSH homes were removed afterward; that local cleanup is not provider deletion completion.

A later read-only recheck at **2026-09-02 15:12:26 UTC** observed that same exact operation as **`completed`**, with native completion time **2026-09-02 14:56:07 UTC**. The pinned candidate probe made only the two allowed status reads; it did not send another delete, allocate a resource, or recreate a claim. This now verifies terminal native operation status, but no fresh inventory-absence check or live bound-claim retry was performed: `cleanupVerified`, `boundClaimRetryValidated`, and `endToEndProof` remain false. The new isolated native home and task credential window were removed.

Probe binary SHA-256: `191ffad2e1a20b61f899df476616f9b133fccbc0d78649330ceaadefe9be4261`. Native CLI SHA-256: `f5fc3e9d05ef0512efc4e93ce7131b542f4b55f810a18c0a28721ef560adf0fe`.

Fresh positive create/use/completed-delete proof with independent absence verification, plus live bound-claim retry proof, remain outstanding. Broader release progress, deadline output, and child-settling work in https://github.com/openclaw/crabbox/issues/1730 remain separate and open.

Original implementation by [Shun Kakinoki](https://github.com/shunkakinoki); contributor history and co-author credit are preserved.


steipete (dismissed):
Two concrete problems remain in the new absence path at exact head 387a1e5d13d1e73859d2cdec03d7757ff68fa4fa:

1. `exactBoxForRelease` treats info 404 plus absent inventory as completed cleanup, and `releaseExactBox` then returns success before checking native deletion completion. ASCII documents that an accepted deletion hides the Box from normal reads immediately, while its operation may still be processing or blocked. After an interrupted delete, this path can therefore forget the claim and report release before completion is known. Preserve the claim unless completion is established for the original resource, and add a regression for a pending deletion hidden from both info and list. Sources: [deletion lifecycle](https://docs.ascii.dev/box/data-retention.md), [delete endpoint](https://docs.ascii.dev/box/api/reference/boxes/permanently-delete-box-data.md).

2. The new helper does not check `ctx.Err()` after successful inventory. A list implementation that cancels the context and returns an empty list still reaches claim removal, unlike the existing confirmation path. Add the cancellation check and a `404 -> cancel during empty-list response` regression requiring cancellation and claim retention.

These are source/documentation findings, not claimed live reproductions. The existing unchanged-claim fence is useful but does not establish provider deletion completion or restore the canceled operation. The progress/deadline requirements in https://github.com/openclaw/crabbox/issues/1730 also remain outside this patch, so that issue should not be closed wholesale by this change.

PR #1734, opened 2026-09-02
fix(local-container): make OOM retry guidance capacity-aware

### LINKED ISSUE #1677: fix(local-container): avoid impossible memory-limit retry advice
## Summary

When a local-container memory limit exceeds the selected Docker daemon's available memory, Docker accepts the larger cgroup limit but the guest can still use only the daemon's smaller capacity. If the workload is then OOM-killed, Crabbox correctly reports `resource_exhaustion=memory` but always advises the user to increase `localContainer.memory` or `--local-container-memory`.

That advice is ineffective when the configured limit is already above the daemon's capacity. The run context, timing JSON, and `inspect --json` also omit the requested limit and effective capacity, so the failure output does not reveal why increasing the same flag cannot help.

## Steps to reproduce

1. Build exact current `main`:

   ```sh
   git clone https://github.com/openclaw/crabbox.git
   cd crabbox
   git checkout ac9de6f2a868f8bab70a1cb32449c17f5e46341c
   go build -trimpath -o ./bin/crabbox-main ./cmd/crabbox
   crabbox_main="$PWD/bin/crabbox-main"
   ```

2. Use a disposable Docker Desktop daemon configured with about 8 GiB and no unrelated running containers, then record its capacity:

   ```sh
   docker info --format 'daemon_mem_bytes={{.MemTotal}}'
   ```

3. Create a retained local-container lease with a 12 GiB memory limit and install Python inside it:

   ```sh
   "$crabbox_main" run \
     --provider local-container \
     --local-container-runtime docker \
     --arch arm64 \
     --local-container-memory 12g \
     --no-sync \
     --no-hydrate \
     --keep \
     --slug issue-memory-capacity-main \
     --timing-record off \
     --stop-after never \
     --shell 'grep "^MemTotal:" /proc/meminfo; printf "cgroup_memory_max="; cat /sys/fs/cgroup/memory.max; sudo apt-get update >/dev/null && sudo DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends python3 >/dev/null; python3 --version'
   ```

4. Compare the configured limit with the capacity Docker exposes:

   ```sh
   container_id="$(docker ps --filter name=crabbox-issue-memory-capacity-main --format '{{.ID}}')"
   docker inspect "$container_id" --format 'configured_memory_bytes={{.HostConfig.Memory}}'
   docker stats --no-stream --format 'effective_stats_limit={{.MemUsage}}' "$container_id"
   "$crabbox_main" inspect --provider local-container --id issue-memory-capacity-main --json
   ```

5. Touch 9 GiB of memory, which is below the configured 12 GiB cgroup limit but above the daemon's available memory:

   ```sh
   "$crabbox_main" run \
     --provider local-container \
     --id issue-memory-capacity-main \
     --no-sync \
     --no-hydrate \
     --keep-on-failure \
     --stop-after never \
     --timing-record off \
     --timing-json \
     -- python3 -c 'x = bytearray(9 * 1024 * 1024 * 1024); [x.__setitem__(i, 1) for i in range(0, len(x), 4096)]; print(len(x))'
   ```

6. Read the cgroup evidence, then clean up:

   ```sh
   docker exec "$container_id" cat /sys/fs/cgroup/memory.events
   "$crabbox_main" stop --provider local-container --id issue-memory-capacity-main
   ```

## Expected behavior

Crabbox should keep the verified memory-exhaustion classification, but its recovery guidance should distinguish a binding container limit from a lower Docker or Podman host capacity. When the requested limit already exceeds the effective capacity, the digest should not recommend increasing the same limit as though it were the active constraint.

The failure evidence should expose enough bounded resource context to tell which limit applied. Exact field names are an implementation choice. The regression is closed when this reproduction identifies the daemon or guest capacity as the smaller bound and gives actionable advice, while an ordinary OOM against a genuinely binding `localContainer.memory` limit still recommends increasing that limit or reducing concurrency.

## Actual behavior

The exact current-main reproduction reported:

```text
daemon_mem_bytes=8317267968
MemTotal:        8122332 kB
cgroup_memory_max=12884901888
effective_stats_limit=64.12MiB / 7.746GiB
```

`inspect --json` included the provider, image, daemon identity, container ID, and lease metadata, but no requested memory limit or effective capacity.

The 9 GiB allocation exited 137 with `oom_kill 1` and `OOMKilled=true`. Crabbox classified the failure correctly, then printed:

```text
resource_exhaustion: memory
hint: increase the memory limit or reduce workload concurrency before retrying
```

Increasing the existing 12 GiB local-container limit cannot raise the selected Docker Desktop daemon's roughly 7.75 GiB capacity.

## Affected area

- Local-container creation passes the configured value directly to Docker or Podman as `--memory` without comparing it with the selected daemon's capacity.
- Local-container OOM evidence correctly owns cgroup counter collection in `internal/providers/localcontainer/backend.go`.
- `printRunFailureDigest` in `internal/cli/run_failure.go` emits one generic memory hint without the requested or effective bound.
- Run timing and lease inspection do not currently expose local-container memory limits or effective capacity.

## Runtime or environment

- Release baseline: [`v0.47.0`](https://github.com/openclaw/crabbox/releases/tag/v0.47.0) at [`0a55378990265271f5830a465aa643605d7d41f3`](https://github.com/openclaw/crabbox/commit/0a55378990265271f5830a465aa643605d7d41f3).
- Exact current `main`: [`ac9de6f2a868f8bab70a1cb32449c17f5e46341c`](https://github.com/openclaw/crabbox/commit/ac9de6f2a868f8bab70a1cb32449c17f5e46341c).
- Current-main binary version: `dev`.
- Current-main binary SHA-256: `098a6139430006d34b25d6c5b119dd0f73e755365156cb59b5f11c909b24394d`.
- Host: macOS ARM64 with Go 1.26.5.
- Runtime: Docker client and server 29.7.2, Linux ARM64 engine, `8,317,267,968` bytes reported by `docker info`.
- Provider and guest: `local-container`, Ubuntu 26.04 ARM64.
- No coordinator configured.

## Evidence

The setup and capacity readback completed as run `run_1cfbb2eae032`. The OOM reproduction completed as run `run_103221066e9d`; Crabbox reported exit 137, `blocked_stage=resource_exhaustion`, `resource_exhaustion=memory`, and `retry_likely=false`. The container cgroup recorded `oom_kill 1`, and Docker inspection reported `OOMKilled=true`.

Closed issue [#1237](https://github.com/openclaw/crabbox/issues/1237) and merged pull request [#1287](https://github.com/openclaw/crabbox/pull/1287) added the accurate per-command cgroup OOM classification and the current recovery hint. They did not cover a configured memory limit that exceeds the selected daemon's capacity, so this report keeps that classification and narrows the remaining guidance failure.

The complete 76-commit range from the release baseline through exact current `main` was inspected. The complete live corpus of 17 open issues, all 10 open pull requests, all 13 open issues authored by `coygeek`, and relevant closed issues and merged pull requests were screened for duplicates, fixes, or supersession. No current item covers an oversized local-container memory limit producing ineffective OOM retry advice.

After the reproduction, `crabbox stop` deleted the retained lease and container. Route-specific doctor reported zero local-container leases.

## Impact

Users can spend another expensive build or test run increasing a limit that is not the binding constraint. The correct classification says memory was exhausted, but the missing capacity distinction makes its recovery instruction unreliable on Docker Desktop and other Docker or Podman daemons whose available memory is lower than the accepted container limit.

## Additional context

This report does not request a new preflight probe or a change to Docker's memory-limit semantics. It concerns the post-failure guidance and evidence Crabbox emits after it has already verified an OOM for the command.

steipete:
## Summary

Fixes https://github.com/openclaw/crabbox/issues/1677.

After a positively evidenced local-container OOM, report the actual container memory and memory-plus-swap settings together with the selected runtime's observed total RAM when available. Advice distinguishes a potentially binding container limit from a limit already at or above runtime RAM, instead of always recommending that users increase the same flag. Total RAM is explicitly not free memory or a universally exact effective limit; unknown capacity degrades to checking constraints and reducing demand. Retrying the same retained ID does not resize it.

Runtime-specific observation and advice stay in the local-container adapter. Core carries bounded, sanitized presentation data through the existing failure-evidence interface. The snapshot survives timing-report reconstruction and one-shot failure bundles. Fresh inspect and non-wait **JSON** status opt into output-only observations; plain status, ordinary resolution, heartbeat, waiting, claims and lifecycle authority remain unchanged. No admission control, resizing or automatic retry is introduced.

## Source and CI verification

The currently tested candidate is `8228f7945cd4f518de0702cd68408e3b43cae4a5`. The original patch was rebased onto `bd41ddbc9d55a5b5dfb68464470cb586ccca1624` without losing main's current-script-only failure capture. The subsequent plain-status fix has a red/green routing regression.

- Focused Go race coverage: 92 top-level tests and 207 passing events, no failed or skipped selected tests, unchanged 60-second test deadlines and no test retries.
- Plain-status follow-up: `go test -race -count=1 -timeout=60s ./internal/cli -run '^Test(MemoryDiagnosticsPresentationRouting|Status|Heartbeat)'` passed after reproducing the undisplayed-probe regression.
- Affected-package `go vet`, documentation link checks and `git diff --check` passed. Independent source reviews were scoped-clean at the configured P0 threshold.
- Clean candidate/reference binaries were built with Go 1.27.0. [Exact-head main CI](https://github.com/openclaw/crabbox/actions/runs/33597601672) passed; this is distinct from manually stopped Testbox workflow runs.

## Real runtime proof

The binaries were exercised through actual Crabbox CLI invocations on separate, exclusively owned Linux amd64/rootful/cgroup-v2 Blacksmith Testboxes. Actual runtime RAM was 8,220,540,928 bytes. Every pressure container was placed below a separately measured task-owned 1 GiB, zero-swap parent. The oversized case used an actual 12 GiB own limit while touching at most 3 GiB in retained/page-touched chunks of at most 64 MiB. No host swap, overcommit, service or kernel policy was changed.

| Lane | Captured behavior and independent judgment |
| --- | --- |
| Docker 28.0.4 | [Full run](https://github.com/openclaw/crabbox/actions/runs/33611496021) plus [read-only presentation follow-up](https://github.com/openclaw/crabbox/actions/runs/33615334596): C1–C8 pass. Real binding/oversized OOMs, ordinary-exit negatives, retained actual settings, unlimited/faulted capacity, one-shot bundle evidence, JSON/plain/wait behavior and complete cleanup. |
| Podman 4.9.3 | [Full run](https://github.com/openclaw/crabbox/actions/runs/33633909208): C1 and C4–C8 pass. Six separately correlated fresh OOM kills, preserved ordinary exits, retained settings, real 12 GiB target/reference cases, controlled capacity failure, one-shot deletion and complete cleanup. C2/C3's informative available-capacity output remains pending; all target OOMs in this run conservatively reported unknown capacity. |

The Docker oversized case reports the observed 12 GiB own limit and smaller runtime RAM, with the guidance: “The actual container limit is at or above observed runtime RAM; raising --local-container-memory does not add RAM.” The reference prints “increase the memory limit or reduce workload concurrency before retrying.” The disclosure keeps the separate parent constraint and swap semantics explicit rather than treating total RAM as an exact effective bound.

The Podman proof independently maps the actual pending allocator's host/namespace PID, birth, UID, namespaces and cgroup-directory identities before permitting allocation in that same process. Native scope, payload leaf and visible-root limits/counters remain distinct. This accommodates Podman's finite nested payload cgroup without relying on names alone or modifying production runtime placement.

Actual nonempty `crabbox-artifacts/timings.json` members were read from eight failure bundles in each full lane. One-shot facts survive automatic deletion. Read-only presentation evidence combines classified bounded runtime-call traces, actual CLI output, all-container lifecycle observations and independent pre/post container/claim state; empty event history alone is not treated as proof.

All resources from the cited completed runtime-proof runs were cleaned through normal CLI/provider operations, with independent terminal and claim/key readback. Testbox backing workflows can show cancelled because normal stop cancels them; those cancellations are not reported as successful CI jobs. Earlier failed fixture/provider attempts are not counted as passing behavior proof.

**Known verification limitation:** Docker exercised the informative available-capacity branches; native Podman exercised real OOM classification, actual settings, safe unknown-capacity fallback, presentation and cleanup. Podman's informative available-capacity C2/C3 branches were not demonstrated, and its independent contract judgment remains blocked for those branches—not a full C1–C8 pass. A later read-only diagnostic attempt never obtained a runner and exposed a separate queued-Testbox cleanup defect; its native state is completed, while its local claim/key are preserved for that separate fix. That attempt is not counted as passing proof and no cleanup or capacity-probe success is invented here.

## Configuration and security

No new credentials, provider permissions or configuration options are required. Optional runtime observations have a bounded aggregate budget and cannot erase positive OOM evidence. Actual retained-container settings take precedence over a new invocation's requested memory value. Presentation fields are not persisted as claims, container labels or actuator inputs. Documentation covers the output and its limits.

### DIFF
--- a/docs/commands/inspect.md
+++ b/docs/commands/inspect.md
@@ -65,6 +65,14 @@ key path, port, user, and host). Empty fields render as `-`.
 label map. Secrets such as broker tokens, provider keys, and VNC passwords are
 never included in either output mode.
 
+[Local Container](../providers/local-container.md#memory-failure-evidence)
+adds fresh, read-only `diagnostic.memory.*` labels to the returned view. They
+describe actual container settings and, when available, total RAM from its
+verified runtime route, not free memory or a universal effective limit. These
+facts are not persisted into claims or container labels, and do not reconstruct
+a previous command's OOM evidence. For a deleted one-shot container, use the
+run's timing/failure-bundle snapshot instead.
+
 For brokered leases, labels describe the current coordinator lease record,
 including allocation identity, target, capacity, timing, and capability facts.
 They are not a fresh read of provider tags. Only the documented
--- a/docs/commands/run.md
+++ b/docs/commands/run.md
@@ -693,9 +693,10 @@ not reconstruct secrets or hidden local shell state. Short-circuit explanations 
 When an SSH backend supplies per-run memory
 exhaustion evidence, the summary and digest use
 `blocked_stage=resource_exhaustion resource_exhaustion=memory retry_likely=false`
-and recommend increasing the memory limit or reducing workload concurrency.
-Evidence read failures are warnings and do not replace the original command
-failure.
+and prefer the provider's bounded contextual hint. Without usable context,
+advice is to reduce memory demand and inspect active limits and runtime capacity
+before retrying. Exit 137 alone is not positive OOM evidence. Evidence read
+failures are warnings and do not replace the original command failure.
 
 Use `--timing-json` to emit a final JSON timing record with provider, lease ID,
 slug, run ID, machine type, repo path, remote workdir, lease acquisition,
@@ -703,6 +704,15 @@ bootstrap, sync phases, command phases, command duration, command-path total,
 end-to-end duration, exit code, normalized `runStatus`, optional `errorKind`,
 stop command, artifacts, and Actions run URL when available. Failed runs also
 include `blockedStage`, `resourceExhaustion`, and `retryLikely` when classifiable.
+Optional `failureEvidence` contains the provider's classification, sanitized
+`hint`, and bounded string-valued `details`. Invalid optional presentation fields
+do not erase valid OOM classification. The same snapshot is copied into local
+failure bundles and the deferred digest, so one-shot deletion does not lose it.
+For [Local Container](../providers/local-container.md#memory-failure-evidence),
+actual container settings, total runtime RAM, and swap are separate observations,
+not an exact effective or free-memory bound.
+App finalization emits timing after cleanup and the failure digest; the executable
+can subsequently append its existing exit diagnostic.
 After an automatic cleanup attempt, `leaseStopped` reports whether the release
 owner confirmed that lease-based recovery is no longer available. An accepted
 release alone does not set it to true. `leaseStopError` independently records a
--- a/docs/commands/status.md
+++ b/docs/commands/status.md
@@ -113,6 +113,11 @@ coordinator has retained recent samples (bounded to the latest 60 per lease) for
 portal trend charts. Under `--wait`, JSON is printed once the lease is ready or
 terminal rather than on every poll.
 
+For Local Container, non-wait JSON status includes fresh read-only
+`diagnostic.memory.*` labels, as described in
+[`inspect`](inspect.md). These optional observations are not requested by
+`status --wait`, heartbeat, or other lifecycle callers, and do not change claims.
+
 ## See also
 
 - [`inspect`](inspect.md) — fuller lease and provider detail.
--- a/docs/providers/local-container.md
+++ b/docs/providers/local-container.md
@@ -162,11 +162,42 @@ is reported as `resource_exhaustion=memory`; an OOM from an earlier command on a
 reused lease does not classify a later ordinary failure. Docker/Podman cgroup
 paths and counter parsing remain inside this provider.
 
-The failure digest marks this condition non-retryable with the unchanged
-configuration and suggests increasing `localContainer.memory` (or
-`--local-container-memory`) or reducing workload concurrency. If cgroup evidence
-cannot be read, Crabbox prints a diagnostic warning and preserves the original
-command failure and ordinary user-command classification.
+If the post-command counter cannot be read, a newly observed runtime
+`OOMKilled` transition can also establish OOM; a prior `OOMKilled=true`, reset
+counter, or exit 137 alone cannot. Evidence read failures are warnings and
+preserve the original exit and ordinary failure classification.
+
+After positive OOM evidence, the adapter optionally observes capacity through
+the same captured runtime/context/connection, with a 500ms aggregate budget
+(including bounded command-output draining), no retries, and bounded output.
+Probe failure never discards positive OOM evidence. Actual inspected container
+settings from **before the command** take precedence over invocation config;
+reusing `--id` does not apply a new `--local-container-memory` value. Legacy
+claims do not reconstruct the originally requested memory string.
+
+The digest, timing JSON `failureEvidence`, and local failure bundle keep these
+separate facts: container memory limit, combined memory-and-swap setting, and
+runtime total RAM when the captured runtime identity can still be verified.
+Limits are `finite`, `unlimited`, or `unknown`; a zero memory-and-swap setting is
+`default`, not a claim about available swap. Missing, malformed, overflowing,
+unverifiable, or timed-out observations remain unknown. There is no synthesized
+effective limit: total RAM is neither free memory nor necessarily the complete
+bound. Parent cgroups, rootless constraints, VM capacity, and swap availability
+can matter; Crabbox does not walk host cgroup ancestors or add privileged probes.
+The container's mounted Docker socket is not used to observe its runtime.
+
+Below observed RAM, advice to recreate with a larger limit is conditional on
+the limit being binding and actual headroom. At or above observed RAM, raising
+the flag does not add RAM; inspect runtime/VM, parent, and swap constraints.
+Unknown evidence advises reducing demand and checking active limits, not
+blindly raising a flag. No automatic resizing, refusal, or retry is introduced.
+
+`inspect` and non-wait `status --json` expose fresh facts in returned labels
+under `diagnostic.memory.*`, with `settings_phase=current`. These are output
+only, not stored claim/container labels or lifecycle authority. They do not
+reconstruct an old OOM. Failure snapshots survive one-shot deletion; fresh
+inspection requires a retained container. Ordinary resolution, heartbeat,
+checkpoint, release, and `status --wait` do not request capacity diagnostics.
 
 Provider flags:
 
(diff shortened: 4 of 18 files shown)


steipete:
Landed in https://github.com/openclaw/crabbox/commit/dec8db5bde43d149970d96afef0f4022a11d447f via https://github.com/openclaw/crabbox/pull/1734, from tested head `8228f7945cd4f518de0702cd68408e3b43cae4a5`.

Source verification included scoped-clean independent reviews, focused race coverage (92 top-level tests / 207 passing events), affected-package vet and docs checks. The plain-status regression was reproduced and then passed with `go test -race -count=1 -timeout=60s ./internal/cli -run '^Test(MemoryDiagnosticsPresentationRouting|Status|Heartbeat)'`. [Exact-head CI](https://github.com/openclaw/crabbox/actions/runs/33597601672) and the connector/CodeQL/release/docs checks were green before merge.

Actual CLI proof used native Docker 28.0.4 ([OOM sequence](https://github.com/openclaw/crabbox/actions/runs/33611496021), [presentation follow-up](https://github.com/openclaw/crabbox/actions/runs/33615334596)) and Podman 4.9.3 ([full sequence](https://github.com/openclaw/crabbox/actions/runs/33633909208)) on separate Blacksmith Testboxes. It covered fresh OOM counters, preserved ordinary exits 7/137, actual retained limits, 12 GiB own limits under a separate bounded parent, controlled capacity failure, actual timing-bundle payloads, one-shot cleanup and read-only status/inspect behavior. The completed proof runs have confirmed inner/outer cleanup; their manually cancelled backing workflow jobs are not reported as passing CI jobs.

Caveat retained in the PR: Docker exercised informative available-capacity guidance. Podman verified real OOM behavior and conservative unknown-capacity fallback; its informative C2/C3 branches were not demonstrated. The later never-started diagnostic Testbox's local cleanup defect remains separate, with its recovery state preserved. Neither limitation is represented as a passing test.

The task checkout is synchronized to clean `main`. https://github.com/openclaw/crabbox/issues/1677 is closed.

PR #1740, opened 2026-09-02
fix: preserve lease activation deadlines after create recovery

steipete:
A coordinator create can succeed after the client loses its response. After confirming the same creation request, the CLI waited for readiness inside the 90-second response-recovery context. A still-provisioning lease therefore failed early even when time remained in the original creation budget. A late initial response could also move readiness past the original deadline. A real cloud desktop warmup exposed the fixed-ID failure after a connection reset.

Move activation into the existing creation owner and use its original deadline for both direct and recovered results. Fixed PUT and token-bound POST recovery now only confirm creation; they no longer own separate readiness loops. Validate lease identity before and after activation, preserve caller cancellation, and retain the existing distinction between explicit recovery of fixed-ID leases and token-bound cancellation of ordinary creates. No timeout values, retry policy, configuration, dependencies, or storage schema change.

The test-only follow-up joins heartbeat/recorder fixture owners, preserves sequential scenarios within checkpoint provider families while overlapping independent families, and fixes descriptor enumeration in the Tart fixture. Production behavior and existing assertions remain unchanged.

Verification:

- Eight virtual-clock cases cover fixed PUT and ordinary POST, activation beyond the recovery window, the original creation deadline, caller cancellation, and an earlier caller deadline. Four failed before the repair; all eight pass afterward. They verify unchanged create intent, no polling before confirmation, and the correct cancellation identity.
- Additional real-process/HTTP proof used the unmodified pre-fix `50b8309e1a2b` CLI and exact-head `0c74856fcbe9` CLI. A loopback coordinator fixture accepted the first request, closed its actual TCP response, confirmed an identical replay, and published activation only after 105 real seconds. No virtual clock or production timeout override was used. Both fixed PUT and token POST failed on the old binary after 91.66 seconds; the repaired binary completed normal warmup after 108.52 and 107.74 seconds respectively. Each case had one logical fixture create and exactly one unchanged replay. The ordinary case also returned a different canonical lease ID and verified that polling used it.
- Real caller SIGINT at 20 seconds stopped both repaired paths: fixed-ID recovery sent no automatic cancel-create, while ordinary recovery sent exactly one cancellation bound to the original create attempt. All six cases completed normal fixture release; local listeners, fixture state and generated test keys were removed afterward. The coordinator/provider state and downstream SSH are synthetic fixtures, so this is CLI/HTTP recovery proof, not an injected live-provider failure.
- A separate normal AWS On-Demand native-Windows desktop allocation using the repaired `4e0a7aae` CLI completed in 219.34 seconds, including stable final-user SSH and guest-loopback VNC readiness. Its changed lifetime owner is byte-identical to this PR head. Canonical release later completed and task keys were removed. The original failed allocation's canonical cleanup also completed. This normal live allocation did not experience an injected response loss and is not presented as recovery proof.
- The broader coordinator/client/cleanup/checkpoint sibling race run passed 128 selected top-level tests and subtests. Scoped vet, formatting and diff checks passed. Before the main refresh, exact `0c74856` [CI](https://github.com/openclaw/crabbox/actions/runs/33669547425) and [Crabbox Release Check](https://github.com/openclaw/crabbox/actions/runs/33669547447) passed. Refreshed `718bca275c7f` [CI](https://github.com/openclaw/crabbox/actions/runs/33704727703) and [Release Check](https://github.com/openclaw/crabbox/actions/runs/33704727741) both passed on attempt 1, including the full Go race suite, coverage, module, Windows transport and release checks.
- Managed Codex review of the original three-file repair was scoped-clean at P0. The fresh ClawSweeper review of exact `718bca275c7f` reports no actionable code/security findings and accepts the CLI/HTTP recovery proof. Direct review covered the lifetime owner, recovery siblings and complete diff.

Redacted real HTTP proof output:

```text
before / fixed PUT: response EOF -> identical PUT confirmed -> context deadline exceeded (91.656s)
before / token POST: response EOF -> identical POST confirmed -> one durable cancellation (91.668s)
after  / fixed PUT: response EOF -> identical PUT confirmed -> activation after 105s -> warmup complete (108.519s)
after  / token POST: response EOF -> identical POST confirmed -> activation after 105s -> warmup complete (107.736s)
after  / SIGINT: fixed=0 automatic cancellations; ordinary=1 token-bound cancellation (~20s)
```

Rank-up disposition: the requested real CLI recovery, readiness and cancellation evidence is recorded above, with normal live-provider allocation reported separately. Deliberately losing a real provider create response remains unperformed: an already-active lease cannot exercise pending activation, and doing so would require another billable create or production fault injection. The local socket-loss reproduction directly demonstrates the defect and repair without altering production coordinator state. This limitation is explicit; no claim of live-provider fault-injection coverage is made.

Production code is 12 lines smaller: three activation branches become one owner. Across the current 13-file PR, tests add 115 net lines (including the fixture follow-up), and command documentation adds six. No changelog, release or package publication. Server-side cancellation diagnostics and provider cleanup behavior are outside this repair.

Main integration: newer main independently parallelized individual checkpoint scenarios. The refresh keeps the reviewed parallel-family structure and sequential scenarios by removing only those 34 nested parallel calls; all 13 task-owned source/test/docs files remain byte-identical to `0c74856fcbe9`. Other main improvements are preserved. Fresh integrated race checks passed: 130 top-level tests, 300 named tests including subtests, the separate Tart descriptor regression, and scoped vet. Managed Codex review of all 13 refreshed paths was scoped-clean at P0; prior binary/live results above remain tied to their stated source revisions.


steipete:
Pushed the fixture repairs in [0c74856](https://github.com/openclaw/crabbox/commit/0c74856fcbe9e47088ba25c73eebe1f677ac23f6); the recovered-activation production change is unchanged.

The checkpoint suite retains all 68 scenarios and its existing build owner. Seven independent fixture families now overlap while each family's native operations remain sequential. Under the rejected per-case fanout, native commands hit their watchdogs; that version is not included. Unexpected watchdog expiry now fails the fixture instead of accidentally satisfying a negative assertion. Heartbeat and recorder fixtures explicitly join their resources, and the Tart descriptor check reads names before its existing inode comparisons.

All 68 checkpoint scenarios, focused cleanup/process race checks, and Tart startup/process race checks passed. Negative controls proved the cleanup, watchdog, and descriptor-leak assertions; temporary injections were removed. The final ten-file change is tests only (+145/-70), and managed Codex review reported no actionable P0 findings.

The full Darwin coverage gate is **still red** at its unchanged ten-minute CLI package limit. Checkpoint execution improved from about 181 to 94 seconds, but that does not establish a complete package speedup or a green coverage ratchet. An unchanged hydration fixture also failed before creating its SSH log; its cause remains unresolved. No timeout, coverage threshold, or test omission was used to hide these results.

[Linux CI for the new head](https://github.com/openclaw/crabbox/actions/runs/33669547425) is running. Independent approval and the current main release freeze remain landing gates; this update does not bypass either or claim deployment.

PR #1764, opened 2026-09-03
fix(artifacts): validate glob paths by component

### LINKED ISSUE #1751: fix(artifacts): validate artifact globs by path component
## Summary

Artifact glob validation is not path-component aware. `--artifact-glob` and `--require-artifact` accept patterns rooted under `.git` or `.crabbox`, even though the matcher always excludes those directories, while rejecting ordinary filenames that merely contain two consecutive dots.

A protected `--require-artifact` path therefore passes local validation, provisions a lease, runs the workload, and then fails an artifact requirement that can never match. Conversely, a safe relative filename such as `reports/result..json` is rejected as though it contained traversal.

## Steps to reproduce

1. Fetch and build the exact live default-branch revision:

```sh
git clone https://github.com/openclaw/crabbox.git
cd crabbox
git checkout 9ab5b3ba7e0500f34c8c70ad564fbe6e2c298e35
go build -trimpath -o ./bin/crabbox-main ./cmd/crabbox
crabbox_main="$PWD/bin/crabbox-main"
```

2. Ask for a required artifact under Crabbox's protected workspace directory and make the workload create both that file and an ordinary marker:

```sh
"$crabbox_main" run \
  --provider local-container \
  --local-container-runtime docker \
  --target linux \
  --timing-json \
  --stop-after always \
  --require-artifact .crabbox/evidence/current-main-proof.txt \
  --shell 'mkdir -p .crabbox/evidence && touch .crabbox/evidence/current-main-proof.txt workload-executed.txt'
```

3. Check a safe relative artifact filename containing consecutive dots:

```sh
"$crabbox_main" run --provider local-container --require-artifact reports/result..json -- true
"$crabbox_main" run --provider local-container --artifact-glob reports/result..json -- true
```

## Expected behavior

Artifact patterns containing an exact `.git`, `.crabbox`, or `..` path component should fail local validation with exit 2 before provider acquisition or workload execution because the artifact matcher cannot return those paths. Safe relative filenames that merely contain `..`, such as `result..json`, should remain valid.

The red/green closure signal is that both artifact flags reject protected or traversal components before any lease or workload side effect, accept `reports/result..json`, and retain the existing runtime exclusion of protected directories during matching and collection.

## Actual behavior

The protected-path reproduction provisioned local-container lease `cbx_0d20a20bd350`, ran the complete shell command, and then exited 7 with `missing required artifact: .crabbox/evidence/current-main-proof.txt`. Run `run_c6aa8350cf17` reported `leaseStopped=true`, so cleanup completed, but the lease and workload were unnecessary.

Both safe-filename commands exit 2 before acquisition with `contains unsupported characters or non-relative path: reports/result..json`.

## Affected area

- `internal/cli/run_artifacts.go`: `safeArtifactGlob` rejects any `..` substring but does not reject exact protected path components; the generated matcher independently excludes `.git` and `.crabbox` at every depth.
- `internal/cli/run.go`: artifact-glob validation runs before repository discovery and provider acquisition, so this is the existing fail-fast boundary for both flags.
- SSH-backed and delegated providers that reuse `ValidateRunArtifactGlobs` or `ValidateRequiredRunArtifactGlobs`.

## Runtime or environment

- Release baseline: [`v0.48.1`](https://github.com/openclaw/crabbox/releases/tag/v0.48.1) at [`c766cde9e877ade7cb13d241d106d5694b842deb`](https://github.com/openclaw/crabbox/commit/c766cde9e877ade7cb13d241d106d5694b842deb).
- Exact current `main`: [`9ab5b3ba7e0500f34c8c70ad564fbe6e2c298e35`](https://github.com/openclaw/crabbox/commit/9ab5b3ba7e0500f34c8c70ad564fbe6e2c298e35), 16 commits after the baseline.
- Current-main binary version: `dev`.
- Current-main binary SHA-256: `a0dfe47db2a2912a64560be49048b7a8ea53c8a9eedc702fa0db4efa30489790`.
- Build host: macOS ARM64 with Go 1.26.5.
- Reproduction provider: `local-container` with Docker 29.7.2 and the pinned Ubuntu 26.04 ARM64 image.

## Evidence

`safeArtifactGlob` currently uses `strings.Contains(glob, "..")`, so it rejects consecutive dots anywhere in a filename. Its remaining checks only enforce a relative path and an allowed character set; they do not reject `.git` or `.crabbox` components. The generated runtime matcher then rejects those exact components and prunes both directories from `find`, making every accepted protected-root pattern unmatchable.

The complete 16-commit range from `v0.48.1` through exact current `main`, the complete live corpus of 20 open issues, all 12 open issues authored by `coygeek`, all 26 open pull requests, and relevant closed issues and merged pull requests were checked for duplicates, fixes, or supersession. No current item owns component-aware validation for the existing artifact glob flags.

Closed issue https://github.com/openclaw/crabbox/issues/475 and merged pull request https://github.com/openclaw/crabbox/pull/502 fixed dangling symlinks after a glob reaches runtime matching. Closed issue https://github.com/openclaw/crabbox/issues/1492 and merged pull request https://github.com/openclaw/crabbox/pull/1628 added exact-path freshness checks. Open pull request https://github.com/openclaw/crabbox/pull/1739 changes shell quoting in `run_artifacts.go` but leaves `safeArtifactGlob` and both failures unchanged.

Focused current-main artifact tests pass under the race detector, which confirms the existing supported cases but contains no regression for protected input rejection or safe filenames containing consecutive dots. The broad race suite was not used as issue evidence because managed-host process and user-directory restrictions caused unrelated failures and the `internal/cli` package reached its 15-minute timeout.

## Impact

Typos or generated configuration that point an artifact flag at a documented protected directory can consume provisioning time and execute a workload before failing a proof condition that was impossible from the start. Legitimate tools that emit filenames containing consecutive dots cannot use those files with artifact collection or existence requirements. The protected files themselves remain excluded, so this is a non-security validation and side-effect-ordering bug.

## Additional context

This report does not request a new preflight capability or a change to protected-directory policy. It asks the existing artifact flag validator to apply the documented path rules at its current pre-acquisition boundary.

steipete:
## Summary

Fixes https://github.com/openclaw/crabbox/issues/1751. Thanks @coygeek for the reproduction.

Artifact validation rejected harmless names such as `reports/result..json` while accepting explicit `.git` and `.crabbox` components that the runtime matcher always excludes. Check path components instead of banning every consecutive-dot substring, and reject protected components at the shared artifact-glob flag boundary before lease acquisition.

Keep the lexical helper separate from the glob policy: artifact-schema paths, artifact-change paths, and delegated downloads retain their own exact-path rules. Runtime traversal, symlink handling, archive bounds, cancellation, and ownership are unchanged. Update the artifact documentation and Unreleased changelog.

## Verification

- New regressions failed on the original implementation; focused race-enabled tests passed after the fix.
- `go test -race -timeout=15m ./internal/cli -run 'Test(SafeArtifactGlob|RunArtifact|RunCommandRejectsProofAndArtifactsWithSyncOnly|ArtifactChangePathValidation|ValidateDelegatedRunFilePath|ParseRequireArtifactSchemaSpec)' -count=1`
- Existing generated shell fixtures executed successfully: required `result..json` matched, collection preserved its exact bytes, and protected-directory and symlink controls remained covered.
- Built-CLI before/after proof passed all 22 cases covering both flags, protected components, safe consecutive dots, traversal rejection, `.gitignore`, and profile expansion. Each deliberately stops at validation with no lease allocation or state writes; this is not a cloud lifecycle claim.
- Independent source audit, structured Codex review, and `git diff --check` passed.

No new dependencies, credentials, configuration migration, or release changes.


steipete:
## Observed CLI proof and maintainer disposition

This is a maintainer-authored fix. The Unreleased entry is intentional and stays at landing; the contributor-only changelog restriction does not apply.

The following results came from the built native CLI in an isolated configuration. Each command deliberately included `run --sync-only`, so an accepted pattern reached the later incompatible-mode guard without allocating a lease. An exit code of 2 alone is not the assertion: the observed diagnostic identifies which validation boundary ran. Both `--artifact-glob` and `--require-artifact` produced these before/after results:

| Pattern | Before | After |
|---|---|---|
| `.crabbox/evidence/proof.json` | `cannot be combined with --sync-only` | `excludes protected path components` |
| `reports/.git/config` | `cannot be combined with --sync-only` | `excludes protected path components` |
| `reports/result..json` | `contains unsupported characters or non-relative path` | `cannot be combined with --sync-only` |
| `reports/../proof.json` | `contains unsupported characters or non-relative path` | `contains unsupported characters or non-relative path` |
| `reports/.gitignore` | `cannot be combined with --sync-only` | `cannot be combined with --sync-only` |

Profile-expanded protected `artifactGlobs` likewise changed from `--artifact-glob cannot be combined with --sync-only` to `--artifact-glob excludes protected path components`. All 22 before/after cases passed their exact-diagnostic assertions; no state files were written.

Matching and collection were separately executed through the real generated shell scripts against a temporary Git/filesystem fixture, including the exact bytes of `.artifacts/result..json`, protected paths, traversal, and symlink controls. This is CLI and local generated-script proof, not a claim of a remote provider lifecycle.

Focused race suite passed:
```text
go test -race -timeout=15m ./internal/cli -run 'Test(SafeArtifactGlob|RunArtifact|RunCommandRejectsProofAndArtifactsWithSyncOnly|ArtifactChangePathValidation|ValidateDelegatedRunFilePath|ParseRequireArtifactSchemaSpec)' -count=1
ok github.com/openclaw/crabbox/internal/cli 50.723s
```

Candidate binary SHA-256: `1a961ebd05c2da4522cdddda5f91122fcc321a7947a4e22e45b5c06b9ea56830`.

Every substantive hosted check passed for `a80925dbfc8fb3fe9aacefa0239ec14a0dc0d162`, including CI, release validation, connector smokes, CodeQL, and docs rendering. No CI retries were needed.

steipete:
Merged as https://github.com/openclaw/crabbox/commit/9e5c7a3481353b08fa68b53c7b2c62474721119c. The reported issue is closed, the maintainer changelog entry is on `main`, and the main checkout is fast-forwarded and clean. Post-merge race validation of `TestSafeArtifactGlob` and `TestRunCommandRejectsProofAndArtifactsWithSyncOnly` passed. The full pre-merge CLI/script proof and exact-head green CI are recorded above.

PR #1785, opened 2026-09-04
fix: isolate ASCII Box host trust per lease

### LINKED ISSUE #1748: ASCII Box reused IP is blocked by provider-wide known_hosts state
## Summary

ASCII Box SSH targets use the provider key directory's shared `known_hosts` file rather than `UseLeaseKnownHosts`. When ASCII Box assigns a previously used public IP to a new VM with a different host key, a new `crabbox run` remains in `ssh-auth` readiness until its 20-minute timeout.

## Reproduction

1. Acquire and stop an ASCII Box lease.
2. Acquire a later lease that receives the same public IP with a different host key.
3. Run a command through Crabbox.

Observed with Crabbox revision `da6484dc0ed73003ccc9850f40e4f170d7795590`; inspection of current `origin/main` (`c9cdc35987531d6f71c317693cb18cbbb0dbe8d9`) shows ASCII Box still does not call `UseLeaseKnownHosts`.

The provider reports the Box ready. `box ssh` succeeds, and direct SSH with the ASCII Box key succeeds. Crabbox's exact probe fails with `REMOTE HOST IDENTIFICATION HAS CHANGED`, identifying the stale entry in the shared ASCII Box `known_hosts` file. Removing only that IP with `ssh-keygen -R <ip> -f <ascii-box-known-hosts>` lets the already-running `crabbox run` recover immediately without starting a replacement Box.

## Expected

ASCII Box host trust should be scoped to the Crabbox lease, as it is for providers that call `UseLeaseKnownHosts`, so ephemeral IP reuse cannot poison later leases while strict host-key verification remains enabled.

steipete:
## Summary

Fix https://github.com/openclaw/crabbox/issues/1748 by applying the existing protected lease-scoped SSH trust helper at the ASCII Box adapter boundary. Both readiness/reuse and ownership-guarded teardown receive the canonical local lease ID. The native Box authentication key, advertised gateway endpoint, legacy IP fallback, and disabled ControlMaster behavior are preserved.

A new Box can reuse an endpoint without inheriting another lease's host key. A changed key within the same lease remains rejected. Retained leases from older clients explicitly enroll on their first connection with this client; shared pins are neither copied into an unproven lease scope nor deleted. The provider docs and Unreleased changelog explain the behavior and credit @shunkakinoki for the report.

## Verification

- Red control: the new adapter regressions failed on the previous source for both IP and gateway endpoints, and unsafe lease paths reached readiness.
- `go test -race ./internal/providers/asciibox -count=1` passed after the fix, including claim/cleanup and guarded-teardown regressions.
- Real native OpenSSH proof replaces a synthetic server key at the exact same loopback host/port: a fresh lease connects; the old lease and provider-wide file reject the changed key; both old trust files remain byte-identical. It reuses the existing forwarding fixture, with no account keys, sshd, or external destinations.
- Broader native SSH forwarding, coordinator cleanup, and host-trust regressions passed under the race detector.
- `go vet ./internal/providers/asciibox ./internal/cli`, docs-link checks, and `git diff --check` passed. Managed Codex review returned no actionable findings in its default P0 scope.
- Source-bound real ASCII lifecycle proof is in progress. The official CLI and existing E2E credential passed authenticated read-only doctor checks; create/sync/run/reuse/destroy and independent native absence must finish before landing.
- Full hosted CI passed on the exact PR head: five substantive workflows and 22 jobs, including [CI](https://github.com/openclaw/crabbox/actions/runs/33835961249). Native deletion completion remains the landing hold.

## Boundaries

The core host-trust implementation and production ownership/deletion fences are unchanged. Provider tests reuse the existing isolated-user-directory utility; they do not touch operator configuration. Native loopback SSH proof is separate from the real provider lifecycle requirement, and neither a timeout nor deletion acceptance alone will be treated as completed cleanup.

Closes https://github.com/openclaw/crabbox/issues/1748.

## Real provider proof and current hold

The frozen clean candidate is `0d29ac2ab4d784329476d4efd00f7f4eb6f16cf9`; its built CLI SHA-256 is `495251babd02ad9010bc1e3f05289fad01998ffe6a35ad0cd7ea2d4170ab9c10`. Tested with the official ASCII CLI `0.1.211-ascii-prod1` and one task-owned Box.

- Warmup succeeded. A deliberately wrong synthetic key was then placed only in the task-owned provider-wide known-hosts file, leaving the correct lease-scoped pin untouched.
- Sync transferred 2,169 files (33.7 MiB), remote execution printed `ascii-host-trust-pass`, and the 22-byte artifact downloaded correctly. The remote source hash matched the candidate.
- A separate retained `--no-sync` invocation printed `ascii-reuse-pass`. The lease-scoped trust and stale shared trust files remained byte-identical through both runs.
- Guarded stop reached its three-minute deadline and returned `context deadline exceeded`. The local claim correctly retained its bound deletion-operation reference.
- Independent native reads report Box `bx_pp783kt6` absent, but operation `bdop_53e2ade7b3f34fdc83113cd1cbc6a4d0` remains `blocked`, with `completedAt: null`. **Absence is not completion: cleanup proof remains incomplete, and this PR must not land yet.** No claim/trust file was forcibly removed or completion fabricated.

The long silent wait also reproduces the diagnostic problem tracked in https://github.com/openclaw/crabbox/issues/1730. That is separate from the host-trust fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Isolate ASCII Box SSH host trust per lease so recycled IPs and gateway endpoints do not block new boxes, while retaining same-lease host-key checks and the native authentication key. [Issue 1748](https://github.com/openclaw/crabbox/issues/1748). Thanks @shunkakinoki.
 - Avoid unnecessary Git metadata lookups during configuration loading, lease claim refreshes, and sync planning while preserving repository and credential trust boundaries. [PR 1783](https://github.com/openclaw/crabbox/pull/1783). Thanks @steipete.
 - Preserve finish-submission and receipt-verification errors, attempt counts, and recovery guidance when terminal run recording times out, without changing retry limits or receipt verification.
 - Closed and joined lease-owned SSH connection masters after confirmed brokered deletion, preserving native connection reuse and lease/host-key isolation while retaining failed local cleanup for a local-only retry. [PR 1774](https://github.com/openclaw/crabbox/pull/1774). Thanks @steipete.
--- a/docs/providers/ascii-box.md
+++ b/docs/providers/ascii-box.md
@@ -115,6 +115,14 @@ BOX_ORG
    lookups, and cancellation retain the claim without recording completion. The
    shared native CLI SSH key is retained.
 
+SSH host trust is separate from the shared native authentication key. Readiness,
+reuse, and guarded teardown use a protected `known_hosts` file for the exact
+Crabbox lease, so a new Box may reuse an IP or gateway endpoint without inheriting
+another lease's host key. A changed host key within the same lease is still
+rejected. Existing leases from before lease-scoped trust enroll on their first
+connection with the new client; Crabbox does not copy or remove pins from the
+old provider-wide file.
+
 If this release observes a valid native deletion acceptance but cannot finish
 waiting because of a timeout, cancellation, or operation lookup failure, it
 durably records the exact operation ID and its claim binding before returning
--- a/internal/cli/ssh_forward_real_unix_test.go
+++ b/internal/cli/ssh_forward_real_unix_test.go
@@ -75,6 +75,11 @@ type forwardSSHServer struct {
 
 func newForwardSSHServer(t *testing.T, user string, allowedPorts ...int) *forwardSSHServer {
 	t.Helper()
+	return newForwardSSHServerAt(t, user, "127.0.0.1:0", allowedPorts...)
+}
+
+func newForwardSSHServerAt(t *testing.T, user, address string, allowedPorts ...int) *forwardSSHServer {
+	t.Helper()
 	_, key, err := ed25519.GenerateKey(rand.Reader)
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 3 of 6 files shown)


shunkakinoki:
Fresh live evidence for exact head `0d29ac2ab4d784329476d4efd00f7f4eb6f16cf9`:

- Built with repository-pinned Go 1.26.5; focused `internal/providers/asciibox` tests passed.
- A real ASCII Box was allocated and reached SSH readiness through the lease-scoped target; sync and the remote command completed successfully (`command=2.916s`, `run total=55.062s`). This reproduces the downstream path without the prior shared known-hosts mismatch.
- Guarded release then exhausted its bounded deletion wait with `context deadline exceeded`; a subsequent direct provider read returned authoritative 404 for the exact Box. Crabbox retained the claim because it did not confirm the asynchronous deletion before its deadline.

This is positive live proof for the host-trust portion, but not a successful Crabbox teardown, so I am not claiming this PR fully cleared. The remaining lifecycle blocker is the separate cleanup/absence-reconciliation work in https://github.com/openclaw/crabbox/issues/1730 and https://github.com/openclaw/crabbox/pull/1789. No host-key verification was disabled and no trust entry was removed.

The remaining branch action here is maintainer-owned removal of the contributor-authored Unreleased entry.

PR #1620, opened 2026-08-29
fix(gcp): verify created image provenance

vincentkoc:
Supersedes https://github.com/openclaw/crabbox/pull/1595. GitHub does not allow that closed pull request to reopen after its branch was force-pushed.

## What Problem This Solves

Typed GCP leases selected through mutable image or snapshot names could boot from a different artifact than the one recorded for the pool. Cleanup and uncertain-create recovery also lacked a durable numeric resource identity, so a same-name replacement or ambiguous provider response could target the wrong instance.

## Why This Change Was Made

- Resolves typed image and snapshot selections to numeric provenance before mutation, persists the observed identity, and verifies the exact boot-disk attachment.
- Carries the raw numeric provider resource ID through generic cleanup custody instead of relying on an instance name.
- Recovers response-lost uncertain creates only when the durable create attempt, lease state, authorization, and compare-and-swap fences still match.
- Migrates eligible legacy release and expiry paths to raw numeric identity behind the same state and authorization fences.
- Treats an exact provider `404` as absence only after a second read and a successful compare-and-swap; replacement or ambiguous identity fails closed.

This is the GCP provenance core only. Typed cohort behavior is deliberately excluded.

## User Impact

Typed GCP pools can prove that registered, recovered, released, and expired instances refer to the intended provider resource and booted from the exact selected image or snapshot. Ambiguous creation, provenance mismatch, and same-name replacement stop safely instead of creating duplicate capacity or deleting the wrong instance.

## Residual Risk

Historical `serverId` values above `2^53` may already have lost precision. They are used only as corroborating evidence during the fenced legacy migration; successful migration persists the provider's raw numeric ID and requires a strict reread before any destructive action.

## Maintainer Disposition

The owner approves the bounded migration cutoff and its residual risk:

- Automatic migration is limited to fully bound, eligible brokered GCP leases under fenced deletion or expiry.
- Pre-upgrade workspace leases and interrupted workspace or tokenless creates without raw provider IDs remain fail-closed and require manual operator recovery.
- Name-only, cross-zone, project-inventory, and same-name replacement adoption are not permitted.
- A historical `serverId` above `2^53` is lossy corroboration only. It is accepted only with exact scope and labels, compare-and-swap fences, and a raw provider-ID reread before destructive action.

## Evidence

- Exact head: `813fb0be6f0a571e9fb2fc5cd37569c8fbf11c9a`.
[...]

PR #1802, opened 2026-09-04
fix: share concurrent provider token refreshes

steipete:
## Problem and change

Azure and GCP independently implemented expiring credential caches but neither coordinated concurrent misses. Azure's ordinary reconciliation inventory starts four resource reads together, producing four OAuth exchanges on a cold client. Concurrent GCP reads similarly duplicate service-account or metadata exchanges, including the metadata retry sequence.

Replace both copies with one small, instance-scoped expiring-token cache. Concurrent callers share one pending refresh; successful credentials are cached; failed refreshes clear so later calls can retry. Keep provider-specific credential loading, trust validation, error construction, retry budgets, clock units, expiry anchors and refresh margins in their adapters.

No global/shared-account cache, persistent token storage, expired-token fallback, automatic 401 retry, new credential source, or provider API permission is introduced. Azure retains its 30-second margin; GCP retains 60 seconds for service-account credentials and 300 seconds for metadata. Existing private test cache setup is consolidated into fixture helpers, with no production compatibility shim. Architecture documentation and Unreleased notes are updated.

Integrated the newly landed GCP provenance work from https://github.com/openclaw/crabbox/pull/1620. Scoped GCP clients copy only a completed token record; pending refreshes and later replacements stay independent. This preserves the existing scoped-client contract and is covered by dedicated cache and provider regressions. All new provenance tests remain intact.

## Regression and runtime proof

- Two new integration regressions failed on unchanged source: four token exchanges instead of one for Azure reconciliation and concurrent GCP resource reads. Both pass after the fix.
- Shared-cache and provider tests cover concurrent cold/expired reads, exact refresh margins, client isolation, failed-refresh recovery, no expired-token fallback, GCP shared transient retries, and real locally generated JWT signing without changing credential source.
- Full Worker suite: 2,368 passed, 3 skipped. Typecheck, lint, format check, Worker build and 245-file documentation link check passed.
- The normal compiled coordinator was run in workerd/Miniflare against strictly intercepted synthetic Azure responses. The real POST admin orphan-sweep route in report-only mode performed four identical ARM resource GETs, with OAuth exchanges reduced from four on the base to one on the candidate. Unauthorized and non-admin requests remained 401/403; inventory was empty with zero errors or deletions. Runtime instances were disposed.
- This is compiled-runtime proof with synthetic provider responses, not a live Azure/GCP account or cloud-resource lifecycle claim. Final candidate bundle matches the tested SHA-256 4ef056b7b404ba70724f2026c31aca08c4c804cf7fd95c0f1c921d9978d9bd3a.
- Managed Codex review is scoped-clean for its visible bundle; its security-sensitive-path exclusions omit the token-cache files. Root and independent semantic review covered the cache implementation directly. No actionable finding remains.

Hosted CI remains the landing gate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Reuse one Azure or GCP token refresh across concurrent requests, reducing duplicate authentication traffic while preserving credential isolation, refresh margins, and retry behavior.
+- Reuse one Azure or GCP token refresh across concurrent requests, reducing duplicate authentication traffic while preserving credential isolation, refresh margins, and retry behavior. [PR 1802](https://github.com/openclaw/crabbox/pull/1802). Thanks @steipete.
 - Skip unnecessary APT translation, AppStream, and command-not-found downloads during minimal Linux bootstrap while preserving required package indexes, signature checks, and later operator defaults. [PR 1794](https://github.com/openclaw/crabbox/pull/1794). Thanks @steipete.
 - Report the correct install, build, or test failure stage from supported phase markers, preserving original exit codes and keeping later artifact-collection failures separate. [PR 1795](https://github.com/openclaw/crabbox/pull/1795). Thanks @steipete.
 - Restore current Tenki CLI inventory and legacy-claim recovery, and retain ownership claims until the exact session acknowledges termination; obsolete workspace/project settings now give migration guidance before creating a lease. [PR 1741](https://github.com/openclaw/crabbox/pull/1741). Thanks @eddiewang.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Reuse one Azure or GCP token refresh across concurrent requests, reducing duplicate authentication traffic while preserving credential isolation, refresh margins, and retry behavior.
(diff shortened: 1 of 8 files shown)


steipete:
Maintainer disposition for https://github.com/openclaw/crabbox/pull/1802 at `c67f3969ecee6f949492e0ad11f2d3e0727dee91`: the changelog finding misclassifies this as an external contributor submission. This is maintainer-directed agent work authored and submitted through the maintainer account. Keeping the Unreleased entry is the repository's required maintainer/agent landing behavior, not an exception to it.

The review reports no implementation or security defect and accepts the compiled-runtime proof. Local Worker validation, independent source review, and managed review are complete; exact-head hosted CI remains the landing gate. No repository rule or merge protection is being disabled.

steipete:
Landed https://github.com/openclaw/crabbox/pull/1802 as `b6f675665a2bbcdca4ca3c8efa2dae30ed91e33b`. The merge tree exactly matches the reviewed candidate against its actual parent; the affected cache/Azure/GCP tests passed again on main.

The root cause was duplicated cache state without a shared pending refresh. The common owner now coalesces concurrent refreshes, while credential sources, trust checks, expiry margins, and retry policy remain in the provider adapters. GCP scoped clients copy completed tokens only, preserving independent refresh state. All five exact-head workflows and 22 substantive jobs passed on the first attempt; compiled coordinator proof showed four OAuth exchanges becoming one without weakening admin admission. Maintainer release notes are included.

PR #1803, opened 2026-09-04
refactor: share delegated provider command intent

steipete:
## What changed

Five delegated providers independently decided whether a command meant literal argv or shell source, then built the same Bash invocation with slightly different helper plumbing. Move that decision into core-owned `CommandIntent`, with owned argument/source snapshots and adapter-supplied shell rendering. Cloudflare Sandbox, Superserve, Crownest, Vercel Sandbox, and Nomad now use it; their five private command builders and ten redundant predicate wrappers are deleted.

Providers still own execution transport, environment, working directory, timeout, and lifecycle. Existing outer shell serialization and login-shell behavior stay intact. Other providers with genuinely different command behavior are not silently normalized here. These five adapters still pass no profile literal-argument map; this refactor does not claim new end-to-end literal protection through their transports.

This is an internal behavior-preserving refactor. Provider architecture documentation is updated; no artificial user-facing changelog entry is added.

## Validation

- A deterministic differential corpus compared 4,022 command cases per provider (20,110 records total), including rendered argv, serialized transport payload, and errors. Every baseline/candidate record matched.
- Native Bash/sh tests execute literal spaces, empty arguments, shell metacharacters, environment assignments, raw/explicit shell source, and exit 42 through both direct argv and serialized transport paths. Profile interpolation tests cover the constructor's literal-argument boundary.
- All five affected provider packages passed `go test -race -timeout=15m`; focused core/native/profile tests passed. `go vet` for core and all five providers, the CLI build, and the 245-file documentation link check passed.
- The broader local CLI suite exhausted its 15-minute package budget without an assertion failure. The active test had only run 14 seconds; the diagnostic JSON-timed rerun subsequently passed the originally active Hybrid SSH test and thousands of other cases, but a late subprocess test exhausted the remaining package budget. That late cleanup-budget test passed in isolation (66.449 seconds). The hosted Go test job has passed; the Worker image-build job was canceled near its time limit and has received one isolated retry after its terminal log showed a canceled image build without a source assertion failure. Neither local full-suite attempt is reported as passing.
- Independent semantic review and managed Codex review found no actionable issue in the selected scope.

No credentials, permissions, provider resources, or cleanup policy change. Native local command execution is not a claim of testing five hosted provider accounts.

### DIFF
--- a/docs/provider-backends.md
+++ b/docs/provider-backends.md
@@ -396,6 +396,16 @@ preflight, timing JSON, and SSH key storage. Keep that helper surface narrow: if
 a provider needs broad command orchestration, the behavior probably belongs in
 core instead.
 
+`ParseCommandIntent` owns the distinction between literal argv and shell source
+for delegated POSIX command adapters. It reuses core shell inference and literal
+argument handling, snapshots the input, and rejects a missing command. The
+result's `Argv` method applies an adapter-supplied shell prefix only when needed;
+an explicitly empty shell source remains valid. Cloudflare Sandbox, Superserve,
+Crownest, Vercel Sandbox, and Nomad use this boundary. They retain their existing
+transport serialization, shell choice, working directory, environment, and
+execution lifecycle. These adapters currently pass no literal-argument map;
+the extraction does not change profile-literal propagation through transports.
+
 Claim-only recovery adapters may use `shared.ResolveProviderClaimStrict` to
 resolve an exact provider/scope-bound claim before a slug while preventing a
 canonical lease ID from falling through. `shared.ValidateClaimBinding` compares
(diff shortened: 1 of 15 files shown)


steipete:
Landed as `46faf00a83421ef758606321e7bdce193a0157da`. Five adapters previously owned copies of the same shell-versus-argv inference. They now use one immutable command-intent representation, while native transport encoding remains in each adapter. The old builders and predicate wrappers are deleted. This does not enable new profile-literal propagation or reinterpret the other providers' single-argument command contracts.

20,110 baseline/candidate records matched, native Bash/sh direct and serialized execution passed, and all five exact-head workflows passed after the Docker audit-stall repair. The two earlier local full-CLI attempts exhausted their package budgets; that limitation remains recorded rather than being relabeled a pass. Hosted full Go race and final post-merge provider/core checks passed. Normal squash merge completed, the actual merge tree matched its preview, and visible main is clean. No user-facing behavior change or artificial changelog entry.

PR #1806, opened 2026-09-04
refactor: share exact-claim touch transactions

steipete:
## What changed

Static SSH, Local Container, and Machine0 independently orchestrated the same durable touch transaction: require an exact carried claim, authorize its provider identity, validate a timeout replacement, prepare lifecycle labels and a timestamp, compare-and-swap, then publish the updated server.

Move that common transaction policy into `shared.CommitClaimTouch`. Provider callbacks retain authorization and lazy preparation; the existing core claim CAS remains the only storage writer and still checks the checkpoint journal. Local Container hydrates its recorded runtime scope before reading effective config. SSH keeps its architecture/status/cache projection; Local Container keeps private-label filtering; Machine0 keeps its native-metadata overlay and timestamp ordering. Static lease IDs remain supported.

This is conceptual policy deduplication, not a claim of net line-count reduction. There is now one owner for the transaction order and failure boundary. No native resource mutation, new API permission, missing-claim recreation, or config/CLI behavior is introduced. Architecture documentation is updated; no artificial user-facing changelog entry is added.

## Verification

- Baseline and candidate race suites passed for SSH, Local Container, and Machine0; the shared suite passed. Vet and the CLI build passed.
- A shared ordering/race regression verifies that missing snapshots, authorization refusal, and invalid overrides cannot reach preparation; a concurrent claim change during preparation cannot be overwritten; successful output is the exact persisted snapshot. Existing provider tests cover timeout persistence, replacement/removal races, identity mismatches, private metadata filtering, and architecture/cache behavior.
- Native proof used a real loopback Docker SSH container and a static SSH lease to that same owned host. Both providers persisted an explicit seven-minute idle timeout, and a fresh CLI process without an override preserved it. Identity, creation anchors, and TTL stayed unchanged.
- Static release removed only its claim and left the host running; Local Container stop then deleted the exact container. Native no-such-object readbacks and empty claim storage verified cleanup. Generated task keys/state were removed. A first fixture attempt correctly failed static destination admission before connection; the retry explicitly bound the observed loopback endpoint. A final harness assertion was corrected for Docker's lowercase error wording, then exact absence was reverified without creating another resource.
- Machine0 is covered by its identity/claim fixtures and source-equivalence review, not a claimed native Machine0 lifecycle. Its changed path adds or removes no native API call.
- Native-tested binary SHA-256: `4d254a124610de3a202c7503a0c7030d50fcf4509eb8a9df4e8e110082c49835`, built before commit from the same production contents.
- Independent semantic review and managed Codex review found no actionable issue in scope. Hosted exact-head CI remains the landing gate.

### DIFF
--- a/docs/provider-backends.md
+++ b/docs/provider-backends.md
@@ -423,6 +423,16 @@ Use `shared.CloneLabels` for plain writable label copies; it returns an empty
 non-nil map for nil input. Keep preservation helpers local when missing, empty,
 and non-empty source values have provider-specific meaning.
 
+`shared.CommitClaimTouch` owns the exact-snapshot touch transaction used by
+Static SSH, Local Container, and Machine0: require the carried claim, run the
+adapter's authorization, validate an explicit idle replacement, prepare labels
+and one timestamp, then call the existing core claim compare-and-swap once.
+Preparation is deliberately lazy: authorization may first hydrate a recorded
+runtime route. Adapters retain identity checks, persisted-timeout defaults,
+label/TTL representation, and public result projection; they update caches only
+after the committed claim is returned. The helper does not mutate a native
+resource, create a missing claim, or replace core's checkpoint-journal fence.
+
 Lifecycle polling is the exception that belongs in
 `internal/providers/shared`, not command core. `shared.Poll` centralizes only
 the repeated read mechanics: last-success retention, attempt limits,
--- a/internal/providers/localcontainer/backend.go
+++ b/internal/providers/localcontainer/backend.go
@@ -1597,27 +1597,20 @@ func (b *backend) cleanupContainerSidecars(leaseID string, labels map[string]str
 }
 
 func (b *backend) Touch(ctx context.Context, req core.TouchRequest) (core.Server, error) {
-	expected, exists, set := core.ServerLeaseClaimSnapshot(req.Lease.Server)
-	if !set || !exists {
-		return core.Server{}, core.Exit(4, "local-container lease %s has no exact claim snapshot; refusing touch", req.Lease.LeaseID)
-	}
-	if err := b.AuthorizeStatusTouchClaim(ctx, req.Lease, expected); err != nil {
-		return core.Server{}, err
-	}
-	if req.IdleTimeoutOverride != nil && *req.IdleTimeoutOverride <= 0 {
-		return core.Server{}, core.Exit(2, "local-container lease %s idle timeout override must be positive", req.Lease.LeaseID)
-	}
-
-	now := time.Now().UTC()
-	if b.rt.Clock != nil {
-		now = b.rt.Clock.Now().UTC()
-	}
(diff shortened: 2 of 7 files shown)


steipete:
Maintainer acceptance for exact head `2839c8c8ee88f5810fd17a58a0ba8485db53a1cf`: no actionable code finding remains, and all exact-head workflows passed. Here is the inspectable result excerpt from the completed native loopback Docker/Static SSH run, not just a proposed test:

```text
binary_sha256=4d254a124610de3a202c7503a0c7030d50fcf4509eb8a9df4e8e110082c49835
local-warmup exit=0
local-container-override exit=0 persisted_idle_seconds=420
local-container-preserve exit=0 fresh_process_preserved_override=true
local-container identity_unchanged=true creation_and_ttl_unchanged=true
ssh-command exit=0
static-warmup exit=0
ssh-override exit=0 persisted_idle_seconds=420
ssh-preserve exit=0 fresh_process_preserved_override=true
ssh identity_unchanged=true creation_and_ttl_unchanged=true
static-stop exit=0 host_preserved=true
local-stop exit=0
exact_task_containers_confirmed_absent=2
claims_absent=true generated_keys_and_state_removed=true
completed=true
```

Both providers used the same actual task-only Linux container and real SSH transport. Static SSH received an explicit loopback destination after the repository-supplied destination was correctly refused. The two container absence checks cover the first fixture attempt and the corrected completed run; the initial PATH/runtime-selection and final case-sensitive diagnostic assertion corrections are described in the body. No host was destroyed by Static SSH release.

The binary was built from the production source committed as `e8a254e374f6b23f4150871573c963914477e169`; the subsequent merge only integrated the shared Docker CI fix. Machine0's unchanged native operations are fixture-tested, not claimed as a live VM run: its changed behavior is the local exact-claim transaction already exercised directly and by the two live adapters. Authorization, label preparation, optional-timeout validation, and the existing revision/checkpoint fence retain their ordering. I accept that proof boundary for this behavior-preserving extraction. No merge protection is bypassed.

steipete:
Landed as `63c636a023cdb7eac7d9e7bfa1e20cacd1a09646`. Static SSH, Local Container, and Machine0 now share the exact-claim touch transaction: require the carried snapshot, authorize, validate an explicit timeout, prepare provider metadata, and commit through the existing revision/checkpoint fence. Native resource interpretation, label preparation, and provider output remain in the adapters; no second claim store or fallback path was introduced.

The actual Docker/Static SSH proof above verified seven-minute overrides, fresh-process persistence, unchanged identity/creation/TTL, Static SSH host preservation, and final exact container/claim cleanup. Machine0 is fixture-tested rather than claimed as a live VM run. All five exact-head workflows and the combined 12-package post-merge race run passed. The normal squash merge tree matched the preview and main is clean. This is shared policy ownership, not a net-line-reduction claim; no user-visible changelog entry was invented.

PR #1804, opened 2026-09-04
fix: share OpenSandbox run finalization and cleanup errors

steipete:
## Problem and change

OpenSandbox duplicated the generic delegated-run sequence and several finalization branches. A command could succeed even when deleting its newly created sandbox failed; timing and session results were assembled before or around cleanup, and failures took inconsistent paths.

Adopt the existing `shared.RunDelegatedSandbox` owner already used by E2B, Modal, and Cloudflare Sandbox. Remove OpenSandbox's copied session, keep/failure, timing, and cleanup bookkeeping. Cleanup-only failures now fail the run; command exit 42 remains 42 when cleanup also fails; failed deletion retains the recovery claim and a kept session. Final timing includes cleanup.

Provider authority stays local: exact endpoint and ownership-marker checks, repository authorization, operation lock, ambiguous-create recovery, absolute TTL, resume validation, command transport, and native deletion. The shared lifecycle now has an explicit `AdmitReuse` boundary: an authorized retained session may be reported even if readiness fails, without refreshing activity, reclaiming prematurely, or suggesting a rerun of an unusable sandbox. Cancellation after resume is checked before local claim mutation.

This is maintainer-directed agent work. Maintainer Unreleased notes and provider/architecture documentation are included.

## Regression and native proof

- New cleanup assertions fail on the original OpenSandbox implementation: success incorrectly returns nil, and command failure loses the secondary cleanup cause. Both pass after migration.
- All five affected race-test packages passed: shared, OpenSandbox, E2B, Modal, and Cloudflare Sandbox. Vet, CLI build, and the 245-file docs link check passed.
- Tests cover failed reuse admission, unchanged claims before/after TTL rejection, cancellation immediately after resume, exact-once unlock, cleanup failure with command exits 0/42, retained recovery claims, and final timing output. Existing ownership/lock/ambiguous-create tests remain intact.
- Live native proof used OpenSandbox server 0.2.3 and execd v1.0.22 with real Docker sandboxes, the compiled candidate CLI, isolated synthetic repository/state, and task-only local authentication. The task launcher restricted the upstream server's Docker publish/probe addresses to loopback; lifecycle and API calls were not mocked.
- Creation, initial/replacement archive sync (including removed files), environment forwarding, status/list, sync-only, reuse, pause/resume, unchanged absolute expiration, original exit 42, and keep-on-failure passed. SIGINT produced a canceled result, and a later remote command confirmed the canceled process no longer existed.
[...]

PR #1807, opened 2026-09-04
ci: fix image audit stalls and Linux fixture publication race

steipete:
## Summary

Avoid duplicate npm advisory requests inside the Node runtime image build. The ordinary Worker CI `npm ci` is unchanged and retains audit reporting; only the Docker build's install and production prune use `--no-audit`. Lockfile resolution, lifecycle scripts, production dependency pruning, image contents, and the job timeout are unchanged.

This repairs a shared CI bottleneck exposed by https://github.com/openclaw/crabbox/pull/1803 and https://github.com/openclaw/crabbox/pull/1804: repeated Worker attempts spent roughly five minutes in Docker install, then stalled in the combined build/prune layer until the 15-minute job timeout. The saved npm output reports retirement of the advisory endpoint. An earlier complete layer log independently shows production pruning consuming six minutes; the actual Node bundle build takes under a second.

## Verification

- Independent source review and managed Codex review: no actionable findings.
- Credential-free local Docker BuildKit build, Linux amd64, from the same tracked Worker source: install 10.4 seconds; combined build/prune 3.1 seconds; full image produced successfully.
- Started the resulting image without network access: Node v22.22.3, 63 installed dependency records, and `node --check /app/dist-node/server.mjs` passed.
- The original Dockerfile reproduced the stall: install 242.8 seconds, build/prune 423.4 seconds. Both completed images have byte-identical server bundle, source map, and package manifest, with identical versions/integrities for all 63 production dependency records. Both passed the offline syntax check. No Docker image has been published.
- Exact-head hosted Worker CI passed: the Docker build completed in 24 seconds and the whole Worker job in 2m42s, with ordinary host audit retained: https://github.com/openclaw/crabbox/actions/runs/33864879738/job/100997351999.

The shared Docker audit fix has now landed through the independently reviewed, exact-head green consumer https://github.com/openclaw/crabbox/pull/1804. This branch retains its history; the remaining effective change against current main is the Linux fixture repair below.

## Linux fixture follow-up

The initial full CI run passed the race suite and image checks, then reproduced an independent foreground-child fixture race in the required Linux supervision gate: waiting for the PID file to exist could read the empty file created by redirection before `printf` wrote its contents. The fixture now publishes a separate ready marker only after the PID write succeeds. No production transport, process-group/exit/absence assertion, or CI no-skip gate is changed.

The repaired foreground-child cancellation case passed 25 consecutive invocations of the compiled Linux arm64 test binary in a credential-free, network-disabled native Docker container, with normal container init and only the test binary mounted. Managed review found no actionable issue. All nine required top-level supervision tests also passed with zero skips in that native environment.

No credential, dependency, user-facing behavior, or changelog change.

### DIFF
--- a/internal/cli/ssh_wsl_cleanup_linux_test.go
+++ b/internal/cli/ssh_wsl_cleanup_linux_test.go
@@ -365,15 +365,17 @@ func TestWSL2ProductionCleanupKillsEntireStagingGroup(t *testing.T) {
 	}
 	t.Run("foreground child cancellation", func(t *testing.T) {
 		childPath := filepath.Join(t.TempDir(), "child")
+		readyPath := childPath + ".ready"
 		command := `sleep 60 &
 child=$!
 trap 'kill "$child" 2>/dev/null || :; wait "$child" 2>/dev/null || :; exit 74' TERM
-printf '%s' "$child" >` + shellQuote(childPath) + `
+printf '%s' "$child" >` + shellQuote(childPath) + ` && : >` + shellQuote(readyPath) + `
 wait "$child"
 `
 		f := startWSLLinuxFixture(t, command, nil, len(command), wslLinuxHelper, 500)
 		f.guard(t)
-		waitForTestFile(t, childPath, 5*time.Second)
+		// Redirection creates childPath before printf has published its PID.
+		waitForTestFile(t, readyPath, 5*time.Second)
 		data, err := os.ReadFile(childPath)
 		if err != nil {
 			t.Fatal(err)
--- a/worker/Dockerfile.node
+++ b/worker/Dockerfile.node
@@ -4,13 +4,14 @@ FROM node:22-bookworm-slim@sha256:e21fc383b50d5347dc7a9f1cae45b8f4e2f0d39f7ade28
 
 WORKDIR /app
 COPY package.json package-lock.json ./
-RUN --mount=type=secret,id=npmrc,target=/root/.npmrc,required=false npm ci
+# CI's install retains audit reporting; do not repeat advisory lookups in layers.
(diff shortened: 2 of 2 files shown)


steipete:
Maintainer acceptance for `556c7bb233d87a734bfcea8ee0d89e7635e2bf38`: no actionable finding remains; all exact-head checks passed, including the previously failing required Linux supervision fixture gate.

The image audit change already landed with https://github.com/openclaw/crabbox/pull/1804 after original/candidate image equivalence and a green consumer run. This PR's remaining effective change is test-only: publish a completion marker after the child PID is fully written, instead of mistaking file creation for completed publication. Every PID/group/cancellation/absence assertion and no-skip gate remains in place. Native Linux proof passed 25 focused repetitions and all nine required top-level supervision tests with zero skips. No production cancellation behavior or timeout was weakened, and no merge protection is bypassed.

steipete:
Landed as `52a35f5fc318578f85d20a11320ebef36648f1b5`. The Docker audit duplication had already landed through https://github.com/openclaw/crabbox/pull/1804; this final merge contains only the Linux PID-publication fixture repair. A file's existence was not evidence that its PID write had completed, so the fixture now waits for a separate completion marker without weakening any process identity or cleanup assertion.

The original image build spent 242.8 seconds installing and 423.4 seconds building/pruning; the no-audit variant retained identical compiled artifacts and production dependency records. Ordinary host audit remains enabled. The repaired head passed all 22 substantive checks, with a 15-second hosted image build and the formerly failing Linux fixture gate green. The CLI merge path hit a GitHub rate limit; the normal REST endpoint merged the exact reviewed SHA under the same maintainer identity, with no administrative bypass. The actual merge tree matched the preview. Required Linux supervision tests passed again on compiled merged main with zero skips, and the visible checkout is clean.

PR #1810, opened 2026-09-04
fix: share Nomad run lifecycle and fence claim refreshes

steipete:
## Summary

Move Nomad's persistent-shell allocation through the existing shared sandbox Run lifecycle, deleting its independent keep/failure/cleanup/timing bookkeeping. Warmup and fresh Run also use one provider-owned job-creation operation. Nomad still owns registration, remote job metadata authorization, allocation selection, exec transport, purge evaluation, and confirmed-absence claim removal; finite scheduler-owned batch jobs are not folded into the sandbox model.

The deeper refactor exposed two real local-claim races: retained refresh and reuse admission could reread a replacement claim, then overwrite it with stale job/allocation metadata; a removed claim could be recreated. Both now fence the originally validated revision, and reuse carries the returned revision into its label update. Explicit reclaim is not permission to adopt a concurrent successor.

Nomad now advertises standard run-session handles (`--lease-output`). Finalization preserves a primary command exit or cancellation when deletion fails, reports cleanup-only failure, retains recovery sessions until deletion is confirmed, and emits timing after cleanup. Shared timing includes the configured workdir. Retained failures refresh idle activity without extending the absolute expiry label. The existing noncancelable local claim-lock wait is unchanged and documented.

## Verification

- Both claim races reproduced against the original paths; replacement/removal regression cases and a kept-run mutation during execution now pass.
- Provider tests cover cleanup-only failure, exit 42 plus cleanup failure, cancellation plus cleanup failure, retained recovery state, final timing, whitespace-only IDs, and valid session metadata.
- Full race suites passed for Nomad, shared, OpenSandbox, E2B, Modal, and Cloudflare Sandbox. Vet, CLI build, complete deadcode check, docs-site build, and generated provider-matrix check passed.
- Managed Codex review and independent semantic review found no actionable findings.

## Native proof

Actual Nomad 2.0.5 Darwin/arm64 agent, actual OrbStack Docker Linux/arm64 allocations, and compiled candidate CLI. No provider API mocks, external cluster, production credentials, or paid resources. Agent listeners were loopback-only; raw-exec, automatic Consul registration/joining, image GC, and daemon-wide dangling-container cleanup were disabled.

```text
cli_sha256=bab59e1f7815238a61273a70e830a791ee6a91aaad639bad1f7933534e5fdefe
warmup=passed
initial_archive_sync_and_environment_forwarding=passed
replacement_archive_sync_including_deleted_file=passed
sync_only=passed
reused_command_exit=42
fresh_success=passed
fresh_command_failure_exit=42
keep_on_failure_then_reuse=passed
cancel_after_command_started=canceled
all_six_exact_jobs_after_cleanup=404
all_matching_allocation_containers_absent=true
all_matching_local_claims_absent=true
agent_stopped=true listeners_absent=true generated_auth_and_claim_state_removed=true
```

The captured fresh success/failure/retention/cancellation runs verified session output and automatic or explicit cleanup using independent native API and Docker readbacks. Cleanup-error combinations and concurrent claim replacement/removal were tested with controlled fixtures, not asserted as live injected API faults.

Harness notes: Doctor correctly rejected an initially non-private config file; permissions were corrected to 0600. Nomad job-list metadata can be null, so capture switched to exact job reads; the interrupted capture's automatically cleaned job was independently reconciled before repeating the live-capture scenario. The long task allocation path produced a nonfatal Nomad task-local `api.sock` warning; HTTP allocation exec, archive mounts, and lifecycle operations still passed. No unsupported broader Nomad capability is claimed.

Docs, generated capability matrix, and the maintainer Unreleased changelog are updated. No dependency or credential changes.


steipete:
Landed as 794421f97ae62f6baaa2d8d9730c6a0b8961be2d. The merged tree matches the reviewed integration tree exactly; main is synchronized and clean.

Nomad now uses the shared run-session lifecycle while retaining native job creation, allocation execution, ownership admission, and exact-job purge in its adapter. The substantive bugs were claim refresh/reuse overwriting a replacement revision or recreating a removed claim, plus cleanup outcomes not participating in the standard final run result. Regression fixtures fail on the original claim paths and pass with the captured-revision fences. Retained failures now refresh idle activity without changing absolute expiry.

Native proof used Nomad 2.0.5 and Docker: warmup, inspection, archive replacement, sync-only, environment forwarding, reuse, exit 42, fresh success/failure, keep-on-failure, explicit stop, and cancellation. All six exact jobs, allocation containers, and local claims were verified absent; the isolated agent was stopped and its task-owned runtime/auth state removed. Cleanup-error injection and concurrent claim mutations were fixture tests, not claims of live injected faults. The existing noncancelable local claim-lock wait remains unchanged.

All 24 non-skipped exact-head checks passed on the first attempt, including full Go races, release checks, and connector smokes; only optional codesmith was skipped. Managed precommit/preland and independent semantic reviews were clean. Postmerge Nomad/shared race tests also passed. No production fleet, release artifact, or real cloud account was changed.

PR #1812, opened 2026-09-04
feat(local-container): add a no-hostname knob (CRABBOX_LOCAL_CONTAINER_NO_HOSTNAME)

atrawog:
## Summary

Adds a **no-hostname knob** for the local-container provider:
`CRABBOX_LOCAL_CONTAINER_NO_HOSTNAME=1` (config `LocalContainer.NoHostname`)
skips the `--hostname <lease>` create arg.

**Why:** environments where the lease container shares a host UTS namespace
cannot set a hostname at all — rootless (uid-1000) postures hit crun's
`sethostname: Operation not permitted` (no caps in the host UTS), and root
postures with host UTS hit podman's own config validation (`cannot set
hostname when running in the host UTS namespace`, which also fires for
`--network host`). Nested-runtime use (rootless containers running podman,
OpenCharly's crabbox beds) needs the hostname-less lease: the loopback
published SSH port contract already works in those postures.

The knob is opt-in and changes nothing for existing users.

## Test plan

- `go build ./internal/...`, `go vet ./internal/...`, `gofmt -l` clean.
- `go test ./internal/providers/localcontainer/ -count=1` — PASS.
- `go test ./internal/cli/ -run 'Env' -count=1` — PASS (env-list test
  updated with the new variable).
- Behavioral proof (OpenCharly bed, rootless nested podman): with
  `CRABBOX_LOCAL_CONTAINER_NO_HOSTNAME=1` the lease provisions, syncs, runs
  and streams without touching sethostname — the full `crabbox run
  --provider local-container` loop completes; without the knob it fails at
  container start on the hostname class above.

## Checklist

- [x] config field + env binding (+ the env-list test) · [x] create args skip · [x] Go suite green


atrawog:
Closed: head moved to the opencharly-org fork (opencharly/crabbox) — recreated as a fresh PR.

PR #1815, opened 2026-09-04
fix: support GCP metadata authentication in workerd

steipete:
## Summary

GCP metadata authentication passed `redirect: "error"` to fetch. Node accepts that option, but workerd rejects it before sending a request; the existing transient-error handler then retried an unsupported option for almost a minute.

Use workerd-supported manual redirect handling and reject every 300–399 response with the existing non-retried trust error before inspecting its marker or body. The endpoint, Google request/response markers, token cache, timeout, service-account path, and legitimate transient retry policy stay unchanged. Redirect destinations and token-looking bodies are never used.

## Verification

The regression fails on unchanged main. Full Worker suite: **2,516 passed, 3 skipped**. TypeScript, lint, formatting, Worker dry-run build, docs build, and diff checks passed. Tests cover 300/399 boundaries, common redirects, a body-unread 304 without a marker, repeated rejection without caching, and existing healthy/concurrent/transient/deadline behavior. Managed Codex review and independent semantic review found no accepted findings.

Real `GCPClient` authentication ran in native **workerd 2026-08-26**, without a fetcher override, preloaded token cache, or patched clock, against strictly synthetic metadata/Compute responses:

```text
unchanged baseline: 500 after 59.035s, zero outbound requests
candidate healthy: 200 after 11ms, one metadata + one Compute request
candidate 302/307/308: trust error in 4ms, one metadata request each
candidate missing marker: trust error in 2ms, one metadata request
candidate transient 503: recovery after 1.007s, two metadata + one Compute request
```

Both runtimes were disposed. This proves the actual runtime transport and refusal behavior with synthetic upstreams; it is not live GCP IAM, attached-service-account, or cloud-lifecycle proof. No dependency, configuration, credentials, production deployment, or cloud resource changes are required. Changelog and provider documentation are updated.


steipete:
Merged as 8165fa35c2db80af7abfebe6196c479ee314faa4; the landed tree matches the verified combined-main preview.

The metadata loader previously requested Fetch redirect mode `error`, which native workerd rejects before making any outbound request. Node-based stubs accepted that option, so unit tests missed the production-runtime incompatibility. Manual redirect handling is supported by workerd; every 3xx response now immediately fails the existing trust boundary before marker checks or token-body consumption. Token caching and normal transient retry behavior remain unchanged.

Native workerd A/B proof exercised the real GCPClient authentication path with synthetic metadata/Compute upstreams and no fetch override or preloaded cache: baseline failed after 59 seconds with zero requests; the candidate authenticated, rejected 302/307/308 without following them, rejected missing markers, and recovered from a transient 503. All runtimes were disposed. Full Worker tests and static/build/docs checks passed; all exact-head substantive checks passed without reruns. The final main readback contains the documented fix and changelog entry. This proof does not claim a production GCP account deployment.

PR #1818, opened 2026-09-04
fix: preserve literal intent through delegated command transport

steipete:
## Summary

Classify command intent once, then quote its execution argv without interpreting it again. Apply this to the five adapters already using the shared command-intent owner: Cloudflare Sandbox, Superserve, Crownest, Vercel Sandbox and Nomad.

Profile expansion already knows which arguments are literal, but the delegated RunRequest handoff dropped that metadata. The adapters then called an operator-aware renderer on already-classified execution argv. As a concrete result, an executable named `FOO=x` could become an assignment that silently succeeded instead of invoking the executable and returning its exit status.

- Preserve an owned copy of profile literal positions in the ordinary run and prewarm request projections.
- Respect a marked literal first argument during assignment inference.
- Add CommandIntent.ShellCommand for literal POSIX transport quoting and remove the five adapters' old rendering wrappers.
- Preserve existing explicit-shell behavior, shell prefixes, native argv/stdin/API transports, environment filtering, cwd, lifecycle and cleanup ownership.

Docs and a maintainer Unreleased entry describe the user-visible fixes. No dependencies, configuration or secrets are changed. This does not standardize the ambiguous single-string behavior of every other provider.

## Proof

Two focused regressions failed before the patch: the final Superserve payload was `FOO='x'` and exited 0 without running the fixture executable, and a marked literal first word incorrectly introduced a Bash layer. Both now pass.

The actual App.Run profile test verifies exact argv and literal positions at both provider admission and delegated execution, including literal `&&` adjacent to a genuine operator. Existing lifecycle tests exercise final command payloads in all five adapters.

Native interpreter proof executes 22 direct/string-transport cases with real Bash/sh, covering quoted arguments, command substitutions as data, glob text, intentional operators and assignments, explicit/empty shell source, exit 42, assignment-shaped executable names and literal separators. The separator marker remains absent. Two additional tests execute the actual captured Superserve API payload and Nomad stdin script through real local shells. All pass with no skips. These are synthetic local interpreter/transport checks, not hosted cloud-account lifecycle claims.

All five adapter race suites and focused core/profile/prewarm race tests pass. Vet, CLI build, docs generation and whole-repository dead-code checks pass. Managed review is clean at the configured blocking threshold; independent semantic review found no actionable production issue. The first expanded native test needed its fixture binary directory in the test process PATH as well as the child environment; that harness setup was corrected before the final native run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,14 +2,13 @@
 
 ## Unreleased
 
-- Preserve literal profile arguments and assignment-shaped executable names across Cloudflare Sandbox, Superserve, Crownest, Vercel Sandbox, and Nomad command transports without reinterpreting them as shell syntax.
-
 - Preserve existing Cloudflare container workspaces when sync upload or extraction fails, clean partial archives, and enforce full-checkout size limits before fresh allocation. [PR 1814](https://github.com/openclaw/crabbox/pull/1814). Thanks @steipete.
 - Protect Nomad runs from overwriting replacement claims or recreating retired leases, and expose standard run-session handles with cleanup-aware final outcomes that preserve the original command exit. [PR 1810](https://github.com/openclaw/crabbox/pull/1810). Thanks @steipete.
 - Preserve individual AWS, Azure, and GCP candidate failures in successful leases' provisioning history across market and regional fallback, including previously omitted GCP on-demand failures. [PR 1811](https://github.com/openclaw/crabbox/pull/1811). Thanks @steipete.
 - Allow fresh brokered AWS Windows and WSL2 leases to bootstrap through advertised SSH port 22 while preserving an explicitly selected final workload port. [PR 1801](https://github.com/openclaw/crabbox/pull/1801). Thanks @steipete.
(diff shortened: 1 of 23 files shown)


steipete:
Landed as https://github.com/openclaw/crabbox/commit/0f11089ee08fde86fb8e2088a326ac7e856497b3. The actual merge tree matches the reviewed integration tree, and main is synchronized and clean.

The original problem was a loss of command classification between the CLI and delegated providers: profile placeholders marked as literal could be parsed again as operators or environment assignments. The fix carries literal positions through RunRequest and gives the already-classified CommandIntent ownership of transport serialization. Cloudflare Sandbox, Crownest, Superserve, Vercel Sandbox, and Nomad now use that common path; provider transport, working-directory and environment setup remain local.

All 23 substantive exact-head checks passed without a CI rerun, including the full Go test job, mandatory executed Linux supervision fixtures, release validation and connector smokes. Native local interpreter/transport proof covers 24 cases, including a literal executable named FOO=x, quoted operators, mixed literal and intentional shell syntax, and original exit42. This is native local execution proof, not hosted-provider live execution. Post-merge races passed for all five adapters plus shared and the focused CLI classification/admission tests.

The separate local full CLI race run hit its initial 15-minute package budget without an earlier assertion failure; its 30-minute diagnostic rerun subsequently passed in 1329.338 seconds (7,846 tests/subtests passed, 14 skipped). The hosted CI gate and its configured limits were not changed. Changelog and command documentation are included.

PR #1820, opened 2026-09-04
fix: share safe archive sync for Upstash and Tensorlake

steipete:
## Summary

Give Upstash Box and Tensorlake the existing shared archive transaction owner: build and check the full snapshot before allocating a fresh sandbox, upload it before touching the workdir, stage delete-sync extraction before replacement, and arm bounded cleanup before the upload can leave partial bytes.

Upstash previously removed the old workdir before local archiving/upload. A failed upload or corrupt tar could therefore erase a reused workspace. Tensorlake already staged replacement, but its independently implemented upload path lost cleanup custody on a partial upload. Both now use PrepareDelegatedArchive and RunDelegatedArchiveSync; the old archive/planning/extraction wrappers are removed.

Native boundaries stay in the adapters. Upstash uploads to absolute paths below /workspace/home while its exec commands use root-relative workspace paths. Tensorlake retains its sbx cp upload and bash exec transport. No provider authentication, endpoint scope, claim authorization, or lifecycle teardown contract changes. SmolVM is intentionally excluded because its supported storage-backed /workspace root needs a separate publication/staging contract.

Shared archive cleanup still preserves the original sync outcome, but reports execution failures as warnings. Its shell retains removal failures and attempts both archive and staging cleanup even under errexit. Successful non-delete sync still merges files; replacement uses the existing shared staged-swap/rollback behavior.

## Verification

Before/after execution used the actual provider sync functions, real local sh/tar/filesystem operations, and synthetic upload/CLI seams. This is native local execution proof, not Upstash or Tensorlake cloud-live proof.

- Baseline Upstash: partial upload and corrupt archive both removed the existing old.txt. Candidate: old.txt retains its original contents; task archive/staging files are absent. Successful replacement removes old files and installs incoming files.
- Baseline Tensorlake: a partially written upload remained after the error. Candidate: the partial upload is removed; corruption preserves the old workspace; successful replace and merge modes preserve their respective semantics.
- Both public Run admission tests now reject an oversized full archive before any provider call. The existing shared full-snapshot guard and ForceSyncLarge behavior are reused.
- Existing cleanup/cancellation regression executes the generated cleanup under real sh -ec with a failing rm, verifies its warning, and retains the original cancellation error.
- Race suites passed for Upstash, Tensorlake, Cloudflare, Nomad, OpenSandbox and shared, plus focused core archive preparation/sync tests. Vet, CLI build, docs build and full deadcode check passed.
- Managed pre-commit review and independent semantic review completed without outstanding actionable findings. One semantic review finding exposed shell cleanup-error suppression; that was corrected and covered before committing.

Changelog and both provider docs are updated. No new dependencies, configuration fields, real cloud allocations, or secret changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 ## Unreleased
 
-- Preserve Upstash Box workspaces when archive upload or extraction fails, clean partial Upstash Box/Tensorlake uploads, and check complete archive limits before creating either sandbox. Thanks @steipete.
 - Cloudflare: retain recovery claims when teardown fails, preserve command/cancellation outcomes, and fence reuse and cleanup against replaced local claims. [PR 1817](https://github.com/openclaw/crabbox/pull/1817). Thanks @steipete.
+- Preserve Upstash Box workspaces when archive upload or extraction fails, clean partial Upstash Box/Tensorlake uploads, and check complete archive limits before creating either sandbox. [PR 1820](https://github.com/openclaw/crabbox/pull/1820). Thanks @steipete.
 - Preserve existing Cloudflare container workspaces when sync upload or extraction fails, clean partial archives, and enforce full-checkout size limits before fresh allocation. [PR 1814](https://github.com/openclaw/crabbox/pull/1814). Thanks @steipete.
 - Protect Nomad runs from overwriting replacement claims or recreating retired leases, and expose standard run-session handles with cleanup-aware final outcomes that preserve the original command exit. [PR 1810](https://github.com/openclaw/crabbox/pull/1810). Thanks @steipete.
 - Preserve individual AWS, Azure, and GCP candidate failures in successful leases' provisioning history across market and regional fallback, including previously omitted GCP on-demand failures. [PR 1811](https://github.com/openclaw/crabbox/pull/1811). Thanks @steipete.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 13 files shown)


steipete:
Maintainer disposition: the review identifies no implementation defect and reports the patch correct. I accept this bounded repair under the requested high-confidence-fix workflow. Real shell, tar and filesystem regressions demonstrate preservation of the previous workspace on partial upload or corrupt archive, replacement/merge behavior, and cleanup of partial transport files. Provider upload seams are injected: this is native fixture proof, not a claim of Upstash or Tensorlake cloud-live testing.

The shared transaction owns prepared-archive admission, staging and temporary-file cleanup; adapters retain native upload paths and execution protocols. Full hosted checks and scoped managed review are required before landing. The changelog addition remains because this is maintainer/agent work, for which the repository explicitly requires an Unreleased entry.

steipete:
Landed in https://github.com/openclaw/crabbox/commit/4e7402e785f2b183468c7153e54eb15754ffdb9d. Upstash Box and Tensorlake now share the prepared-archive transaction instead of implementing separate admission, upload, extraction and cleanup policy. Upstash no longer deletes the previous workspace before upload; both adapters retain their native transport/path boundaries and clean partial upload state.

All substantive hosted checks passed on the first attempt, including the full Go race and mandatory native Linux fixture/build gates. All six affected provider/shared race packages passed again after merge. The resulting main tree exactly matched the preview. Native shell, tar and filesystem tests cover failed upload, corrupt archives, replace/merge behavior and cleanup; provider transports are injected, so this is high-confidence fixture proof rather than cloud-live lifecycle proof.

The shared cleanup path now also exposes removal failures without replacing the original sync result. SmolVM was intentionally excluded from this transaction adoption because its supported /workspace root can be a mounted disk; treating that as an ordinary renameable directory would be an unsafe abstraction.

PR #1819, opened 2026-09-04
fix: stop fresh acquisition retries after rollback failure

steipete:
## Summary

Make reported acquisition cleanup failure a shared, cause-preserving veto on allocating a fresh lease. Ordinary direct AWS, Azure, GCP and Hetzner acquisition already share a two-attempt bootstrap retry loop, but their rollback outcomes were inconsistent: warning-only errors in three adapters, an ordinary joined error in Hetzner. Both representations allowed another allocation after cleanup failed.

`JoinAcquireCleanupError` retains both causes and marks the failed rollback. The shared retry owner checks that marker before classifying bootstrap retry and emits the combined recovery diagnosis. Bootstrap exit5 remains the primary exit; cleanup failures remain inspectable through errors.Is. Successful cleanup still permits the existing retry budget.

Adapters keep their resource identities, deadlines and deletion order. AWS still attempts instance then exact key-ID cleanup, including its existing key attempt after an instance error. Hetzner still deletes an owned-created key first and makes no server DELETE after key failure. Fixed-ID AWS acquisition and brokered cleanup are untouched. GCP also retains cleanup-client reconstruction failure even when its existing fallback client returns nil; that fallback is not proof of selected-scope cleanup.

## Verification

The baseline GCP regression made two creates/two cleanup attempts despite a failed delete. The fixed public Acquire paths now stop after one allocation when rollback fails; successful-cleanup controls keep two attempts. Tests cover AWS client/instance/key failures, Azure/GCP delete failures, GCP cleanup-client fallback, Hetzner key-first ordering, and shared wrapped/joined markers with both keep values. The GCP SSH forwarding wrapper is now injectable, matching the other adapters, so final tests inject the classified failure at the actual bootstrap phase rather than the IP phase.

All five affected package race suites pass; combined-main races also include Cloudflare's newly landed shared-lifecycle consumer. Vet, build, docs generation and whole-repository dead-code checks pass. Managed precommit/main-integration reviews and independent semantic review have no accepted actionable finding.

These are high-confidence adapter/loop regression tests with injected provider failures, not a claim of a live cloud cleanup failure or uniformly confirmed deletion. No cloud resource was deliberately leaked for proof. The exact control flow and provider-call ordering are exercised through the actual public Acquire entry points. The focused trace is below.

```sh
[...]

PR #1821, opened 2026-09-04
fix: bind provider readiness to the created resource

steipete:
## Summary

Integrated with current main after https://github.com/openclaw/crabbox/pull/1819 landed, preserving its acquisition cleanup-failure retry veto. The six-provider race suite and native HTTP proof pass on the resolved main integration.

Bind readiness observations to the allocation being observed. RunPod already rejected a different pod ID or name; Hetzner replaced both the working server and its rollback target with the later response. A malformed readiness response could therefore send SSH or cleanup to another server.

A small shared NamedResourceIdentity owns complete, exact ID/name matching. RunPod adopts it without changing its original diagnostics, keep behavior, polling or exact created-ID rollback. Hetzner validates creation against the requested name/lease/slug/provider ownership/encoded selected key before arming server cleanup, then freezes the created server and a cloned label map. Readiness cannot replace that binding before SSH, SetLabels or acquired-cache publication.

If the creation response itself is unbound, Hetzner reports unresolved allocation evidence and withholds server deletion; only an independently created attempt key can be removed. For valid creation plus inconsistent readiness, rollback retains the original numeric ID/key. Reused shared keys remain retained, and their labels use the existing wire encoding. Legacy resolve/release behavior is unchanged.

This deliberately covers ID-addressed RunPod and Hetzner resources. It is not a generation fence for Azure/GCP reusable names, a new durable claim, or a generic provider cleanup framework.

## Before/after proof

The credential-free loopback fixture runs actual public Hetzner Acquire, core.HetznerClient HTTP encoding/decoding, and the actual IP waiter. SSH readiness is a counting stub. No real cloud allocation or replacement was induced.

Baseline, with creation42 and readiness43:

```text
POST /servers -> created42
GET /servers/42 -> observed43
SSH callbacks=1
DELETE /ssh_keys/7
DELETE /servers/43
```

Candidate with the same responses:

```text
POST /servers -> created42
GET /servers/42 -> observed43
SSH callbacks=0
error=hetzner readiness rejected: resource ID mismatch
DELETE /ssh_keys/7
DELETE /servers/42
```

The matching42 control still reaches SSH and labels42 ready. This is native Go/HTTP fixture proof, not cloud-live or real-SSH execution proof.

Regression cases cover missing/different identity, contradictory CloudID, changed lease/provider/slug/key/owner labels, mutable creation/readiness label-map aliasing, invalid creation with no server deletion, healthy endpoint changes, and retained wire-encoded shared keys. RunPod covers ID/name mismatches with both Keep values and original-ID cleanup. The cleanup-retry suite was integrated using valid newly created resource fixtures, preserving one-versus-two attempt assertions.

Race tests pass for Hetzner, RunPod, shared, AWS, Azure and GCP after parent integration. Native HTTP proof was rerun on the integrated source. Vet/build/docs checks and managed pre-commit/integration reviews passed; independent semantic review found no actionable issue. Changelog and lifecycle documentation are included. No new dependencies, configuration or credentials.

## CI follow-up

The main-target Scripts job exposed an unchanged Unikraft cancellation-fixture handshake race: the fixture published its warmup-started marker before spawning and recording the descendant PID, allowing TERM to interrupt before the asserted PID file existed. A scheduling-gap replay reproduced the same ENOENT. The marker now follows PID recording; the descendant-reaping and zero-residue assertions remain unchanged, and the corrected native test passed ten consecutive runs. This is a test-only fix, not a change to provider cancellation behavior. Full CI is rerunning on the updated head.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 - Stop direct AWS, Azure, GCP, and Hetzner bootstrap retries from allocating another machine when rollback reports a cleanup failure, preserving both the original failure and cleanup diagnostics.
-
+- Cloudflare: retain recovery claims when teardown fails, preserve command/cancellation outcomes, and fence reuse and cleanup against replaced local claims. [PR 1817](https://github.com/openclaw/crabbox/pull/1817). Thanks @steipete.
 - Preserve existing Cloudflare container workspaces when sync upload or extraction fails, clean partial archives, and enforce full-checkout size limits before fresh allocation. [PR 1814](https://github.com/openclaw/crabbox/pull/1814). Thanks @steipete.
 - Protect Nomad runs from overwriting replacement claims or recreating retired leases, and expose standard run-session handles with cleanup-aware final outcomes that preserve the original command exit. [PR 1810](https://github.com/openclaw/crabbox/pull/1810). Thanks @steipete.
 - Preserve individual AWS, Azure, and GCP candidate failures in successful leases' provisioning history across market and regional fallback, including previously omitted GCP on-demand failures. [PR 1811](https://github.com/openclaw/crabbox/pull/1811). Thanks @steipete.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 - Stop direct AWS, Azure, GCP, and Hetzner bootstrap retries from allocating another machine when rollback reports a cleanup failure, preserving both the original failure and cleanup diagnostics. [PR 1819](https://github.com/openclaw/crabbox/pull/1819). Thanks @steipete.
-- Reject inconsistent Hetzner creation/readiness identities before SSH, and keep failed-acquisition cleanup bound to the original server and key; share exact ID/name validation with RunPod. Thanks @steipete.
+- Reject inconsistent Hetzner creation/readiness identities before SSH, and keep failed-acquisition cleanup bound to the original server and key; share exact ID/name validation with RunPod. [PR 1821](https://github.com/openclaw/crabbox/pull/1821). Thanks @steipete.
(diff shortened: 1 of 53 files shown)


steipete:
Maintainer disposition: accept the fail-closed identity decision in this bounded fix. A malformed creation response does not authorize VM deletion, and an inconsistent readiness response must not redirect SSH or cleanup. For a valid created Hetzner server, rollback retains the original immutable numeric ID; a newly created key can be cleaned independently, while reused keys stay retained. This is intentionally different from name-addressed generation cleanup.

The proof is actual Go HTTP/client execution against synthetic loopback responses, with SSH represented by a counting seam; no cloud-live replacement was induced. The old dirty-stack condition is resolved and the PR targets main. CI's unrelated preexisting cancellation-fixture race was reproduced and repaired without weakening the descendant-cleanup assertion. The maintainer/agent changelog entry stays under the repository's stated policy.

steipete:
Landed in https://github.com/openclaw/crabbox/commit/ff6b0f2ab60197f9118400f96af98e05a668cbd1. RunPod and Hetzner now share exact ID/name observation matching; Hetzner keeps a frozen created-server/label binding so later readiness cannot redirect SSH, ready tagging or rollback. Inconsistent creation does not authorize server deletion, while independently created key cleanup remains separate. Existing numeric-ID rollback and shared-key retention are preserved.

The main-target Scripts job first exposed an unrelated readiness-marker/PID race in the Unikraft cancellation fixture. A scheduling-gap replay reproduced it; moving the marker after PID recording passed ten consecutive native runs without weakening cleanup assertions. The repaired head passed all five workflows and all 11 CI jobs, including race tests, required Linux fixtures and build. Six affected provider/shared race packages passed again after merge; the actual main tree matched the preview. Native HTTP proof uses the real client/waiter with synthetic responses and a counting SSH seam, not cloud-live replacement testing.

PR #1825, opened 2026-09-04
fix: share temporary environment profile ownership

steipete:
## Summary

Give Modal and Tensorlake one shared owner for a temporary shell environment profile. The owner creates a private closed local file and an independent random remote path, takes cleanup responsibility before the actual upload attempt, and removes the local file after that attempt. Close is idempotent and retains partial-upload cleanup responsibility.

Tensorlake previously registered cleanup only after successful upload, so a failed or canceled partial copy leaked the remote profile. Its adapter now registers cleanup before returning an upload error, uses a fresh bounded uncanceled cleanup context, and rechecks the captured claim and native namespace before either upload or removal. Denial before upload does not acquire remote cleanup authority. Modal adopts the same file lifetime while retaining its native SDK transport and existing run cleanup order; waiting for its claim lock is now bounded by the cleanup context.

Both adapters use the same sorted, validated shell-profile renderer and source-failure gate. The gate does not add errexit to user commands. The old duplicate renderer, validators, wrappers and unused Modal random-suffix helper are removed.

## Proof

The original Tensorlake public Run regression left a synthetic remote profile after a partial copy returned failure. The candidate removes both local and remote temporary files after error or cancellation, does not run the user command, and does not terminate a reused sandbox. Cleanup sees a live context and a deadline. Changed local claims or native namespaces prevent stale remote removal while still removing local bytes.

Credential-free native fixtures execute real file copy, shell removal and Bash profile sourcing. They verify overlapping profile lifetimes, literal quotes/newlines/command-substitution text, missing-profile refusal, and preservation of user shell failure semantics. Portable Go tests cover custody and repeated Close. Existing Modal upload-error and cleanup-before-termination regressions pass. This is native fixture proof, not a Modal or Tensorlake cloud-live claim.

Shared, Modal and Tensorlake race suites pass after integration with main and https://github.com/openclaw/crabbox/pull/1820. Vet, CLI build, docs build and deadcode checks passed; managed reviews are scoped-clean and independent semantic review has no outstanding findings.

## Boundaries

No new dependency, credential or configuration is required. The shared owner is synchronous: Close follows Upload, not concurrent with it. Remote file permissions and provider authorization remain adapter responsibilities; this PR does not claim remote-permission hardening, general diagnostic scrubbing, or whole-run Tensorlake fencing. Cleanup failures warn without replacing the primary failure. Documentation and the maintainer Unreleased entry are included.

## CI repair: cancellation at clean EOF

The full Go job exposed a preexisting Cloudflare streaming race: caller cancellation followed by a clean EOF could be mislabeled as a missing-completion provider error. A deterministic response-body fixture reproduces it; an accepted completion delivered with the same EOF still wins. The same narrow boundary exists in Azure Dynamic Sessions and is fixed there too. Both clients now inspect caller cancellation only after scanner errors and before incomplete-EOF errors; protocol decoding, accepted events and explicit read errors keep their precedence. Azure's unreachable completed flag was removed.

The original canceled-run/failed-cleanup test remains unchanged. Focused race tests cover both stream clients and the three environment-profile packages. The repaired branch is integrated with current main, including the separately landed readiness fix; full CI will rerun.


steipete:
Landed in https://github.com/openclaw/crabbox/commit/d5346bb114c9cdd2aab18fa8bed56b4b0ff43986. The merged tree exactly matches the reviewed candidate plus current main; the visible checkout is clean and synchronized.

Modal and Tensorlake now share environment-profile custody instead of independently implementing file creation, rendering and cleanup. The important behavior fix is that a failed partial Tensorlake upload still has an owner responsible for removal. Cleanup keeps the original authority, runs with a fresh bounded context, and does not replace the command's primary outcome. Native file-copy and shell fixtures establish the failure-before/fix-after behavior; this is not a cloud-live claim.

The first full CI run exposed a preexisting cancellation/clean-EOF race in Cloudflare. Deterministic regressions reproduce it and the matching Azure Dynamic Sessions boundary; the original canceled-run/failed-cleanup test remains unchanged. Repaired head dd57512d passed all 23 substantive checks, including the full Go suite, executed Linux supervision fixtures and build. Five affected provider/shared race suites also pass on the actual merged main. No outstanding review findings, new credentials, dependencies or runtime resources.

PR #1826, opened 2026-09-04
fix: unify checked SmolVM byte uploads

steipete:
## Summary

Replace SmolVM's two shell upload implementations with one checked byte-transfer owner. Both archive injection and small-file writes previously ended in an unconditional success echo, hiding failed decoding, redirection and extraction. Archive injection also checked only the final stage of a decoder-to-tar pipeline; both paths used predictable shared staging names.

The shared private helper exclusively creates a temporary directory, decodes into a private file, checks decoding before consuming the payload, and cleans only its own directory. Cleanup preserves an earlier failure and reports a cleanup-only failure. WriteFile publishes the complete private payload from the destination filesystem, so failed decoding does not truncate existing content. Archive extraction restores the caller's umask; private staging does not strip executable/read permissions from repository files. SmolVM's local external-tar archive producer is also replaced by the existing shared Go archive producer.

## Before/after verification

Credential-free native fixtures exercise the actual SmolVM HTTP client against a loopback exec service that runs the generated POSIX shell, GNU base64, tar and filesystem operations. Healthy write and archive controls pass on the baseline. Four baseline cases falsely return success: partial decoder output, a directory destination, corrupt tar data, and a decoder that emits valid archive data then exits 23. The candidate rejects all four and does not consume a failed decode.

The expanded eight-case suite also checks that cleanup failure remains visible, exit 23 survives simultaneous cleanup failure, successful publication can remain present after a cleanup error, private file modes remain 0600, archive executable mode remains 0755, and staging is absent except where cleanup was deliberately failed. Tests isolate baseline fixed staging names and remove all fixture directories. This is actual client/native-shell fixture proof, not a hosted SmolVM lifecycle or mounted-workspace proof.

The complete SmolVM race suite, vet, CLI build and documentation build pass. Managed review is scoped-clean; independent semantic review caught the umask interaction, which was fixed and regression-tested, and has no outstanding findings. Documentation and a maintainer Unreleased entry are included.

## Deliberate boundaries

This patch does not rename or make replacement of the /workspace mount root transactional. Existing sync preparation ordering and final environment-profile lifetime remain separate work. The remote trap cannot promise cleanup after SIGKILL or an ambiguous HTTP response while execution may continue. There is no retry of ambiguous extraction and no broad cleanup request. Provider identity/lifecycle protocols and credentials are unchanged; no new package dependency is added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Report SmolVM decoder, file-write and extraction failures instead of false upload success, isolate temporary upload files, and preserve existing files when decoding fails. Thanks @steipete.
 - Stop direct AWS, Azure, GCP, and Hetzner bootstrap retries from allocating another machine when rollback reports a cleanup failure, preserving both the original failure and cleanup diagnostics. [PR 1819](https://github.com/openclaw/crabbox/pull/1819). Thanks @steipete.
 - Cloudflare: retain recovery claims when teardown fails, preserve command/cancellation outcomes, and fence reuse and cleanup against replaced local claims. [PR 1817](https://github.com/openclaw/crabbox/pull/1817). Thanks @steipete.
 - Preserve existing Cloudflare container workspaces when sync upload or extraction fails, clean partial archives, and enforce full-checkout size limits before fresh allocation. [PR 1814](https://github.com/openclaw/crabbox/pull/1814). Thanks @steipete.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Report SmolVM decoder, file-write and extraction failures instead of false upload success, isolate temporary upload files, and preserve existing files when decoding fails. Thanks @steipete.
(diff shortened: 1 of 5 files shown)


steipete:
Landed in https://github.com/openclaw/crabbox/commit/227b0b3e4078c18c3dbcea56a31bbaeb9d3c06f1. SmolVM archive injection and file writes now use one checked byte-transfer owner. It replaces predictable shared staging paths with a private directory, checks decoding before extraction/publication, preserves earlier failures through cleanup, and keeps failed decoding from truncating an existing file. The duplicate local archive producer was removed in favor of the shared Go implementation.

Native HTTP/shell fixtures reproduced four false-success cases and pass eight cases after the fix, including cleanup-failure precedence and file permissions. Independent review caught the temporary umask leaking into archive extraction; the fix restores the original mask and verifies executable permissions. All five CI workflows, all 11 CI jobs, race tests, mandatory Linux fixtures and build passed on the first attempt. The resulting main tree matched its preview. This is native fixture/CI proof, not a cloud-live SmolVM lifecycle claim; mounted-workspace replacement and final environment-profile lifetime remain separate work.

PR #1834, opened 2026-09-04
fix(upstash-box): share isolated environment profile custody

steipete:
## Summary

Adopt the shared environment-profile owner for Upstash Box. One private local file and one unique remote path now have an explicit upload-attempt lifetime, so an upload error or cancellation cannot skip cleanup. The shared source wrapper rejects the entire workload after failed profile sourcing and permits an explicitly empty source body. Delete Upstash's duplicate renderer, environment-name validator, deterministic filename, and env-only JSON WriteFile transport; use the existing synchronous multipart uploader.

Keep Upstash's authority and outcome rules in the adapter. A file-only receipt captures the original API client, Box ID/name/positive creation timestamp, and the observed local claim or its absence. Each upload/removal runs under that unchanged claim-state fence and a fresh native identity check. Discovery-only raw-ID, slug, and lease-ID reuse still works without publishing a claim; this receipt cannot authorize Box deletion. Cleanup has the existing separate 15-second budget and runs before final timing/outcome selection. Cleanup-only failure remains exit 5; a primary command exit, transport error, or upload cancellation retains its outcome.

## Verification

The production newAPI factory, HTTP client, multipart uploader, exec-stream parser, and native sh process run against a loopback fixture included in the existing provider test file. The identical fixture passes one of six cases on the unchanged baseline and all six after this fix:

| Case | Baseline | Candidate |
|---|---|---|
| Healthy quoted/multiline environment | Pass | Pass |
| Server writes file then returns upload failure | Remote residue | One removal; original error |
| Cancellation after server writes file | Remote residue | One bounded removal; cancellation preserved |
| Box generation changes after upload | Stale removal | Refuses removal; exit 5 |
| Profile returns failure before a multi-command script | Later command runs | Whole script stopped; exit 9 |
| Explicitly empty source with environment | Syntax error | Valid no-op; cleanup succeeds |

Additional regressions cover raw-ID/slug/lease discovery, full legacy claim snapshots, changed/removed/appearing claims, changed native identity, denied-before-upload custody, an absent-claim fence held during mutation, two independent profiles on the same Box, idempotent close, bounded cleanup, and primary error/exit precedence. Two original partial-upload regressions also fail on the baseline and pass after the fix.

Full race suites pass for Upstash Box, shared helpers, SmolVM, Modal, and Tensorlake. Scoped vet, CLI build, and docs build pass. Managed review is scoped-clean; independent semantic review found two test robustness issues, both corrected and read back, with no remaining actionable finding. The native fixtures are local production-client proof, not hosted Upstash proof; no cloud resource or real credential was used.

## Boundaries

No new dependency or configuration option. Positive creation identity is required before environment upload; an incomplete or changed identity fails closed. Native GET and file mutation are not an atomic generation-CAS, cancellation cannot guarantee that a delayed server write stops, and this does not claim whole-run concurrency isolation or remote file mode guarantees. Existing Box retention/deletion policy remains separate. Maintainer changelog and provider documentation are included.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Clean partial Upstash Box environment uploads after failure or cancellation, isolate each profile, and refuse stale file cleanup while preserving discovery-only reuse and original command outcomes. Thanks @steipete.
 - Preserve literal profile arguments through CodeSandbox, OpenComputer, and Docker Sandbox execution, sharing command-intent parsing without changing provider shells or environment transports. [PR 1830](https://github.com/openclaw/crabbox/pull/1830). Thanks @steipete.
 
 - Reject changed Azure VM identities during acquisition readiness and use identity-checked VM/companion cleanup for failed acquisitions instead of blind name-based rollback. [PR 1827](https://github.com/openclaw/crabbox/pull/1827). Thanks @steipete.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Clean partial Upstash Box environment uploads after failure or cancellation, isolate each profile, and refuse stale file cleanup while preserving discovery-only reuse and original command outcomes. Thanks @steipete.
+- Clean partial Upstash Box environment uploads after failure or cancellation, isolate each profile, and refuse stale file cleanup while preserving discovery-only reuse and original command outcomes. [PR 1834](https://github.com/openclaw/crabbox/pull/1834). Thanks @steipete.
 - Read all AWS recovery inventory pages and retain cleanup debt when pagination is incomplete; describe interrupted provisioning without assuming a deployment caused it. [PR 1832](https://github.com/openclaw/crabbox/pull/1832). Thanks @steipete.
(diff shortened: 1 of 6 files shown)


steipete:
Maintainer disposition: the remaining changelog-authoring finding is inapplicable. This is a maintainer/agent-owned change, and the repository explicitly requires its user-visible fix to be recorded under Unreleased with the full PR link and thanks. The entry stays. The current-head review accepts the implementation and native proof; hosted Box allocation is not claimed. Final exact-head CI and the normal merge protections remain required.

steipete:
Merged as bddc2f21d6083f54d17c45a0bc77a8d49f4bcc17. The old environment-upload path could leave a remote profile behind after a partial upload or cancellation. Upstash now uses the shared per-operation shell profile owner, with a receipt bound to the original client, box identity/generation, exact profile path, and claim presence/content. Cleanup is bounded and idempotent, and a changed identity or claim refuses stale file mutation.

This receipt authorizes only that temporary file's upload/removal, not deleting a box or adopting a lease. Discovery-only reuse remains supported without inventing a claim. Environment values stay in the existing upload transport, source failure prevents the workload from running, and cleanup diagnostics do not replace the original command or cancellation outcome.

All five final-head workflows passed, including every CI job, without reruns. The normal squash merge produced the exact prepared tree, and the visible main checkout was fast-forwarded and clean. After merging, uncached Upstash/shared race suites passed and all six production-client native HTTP profile cases passed again: healthy use, lost upload acknowledgement, cancellation, changed generation, source failure, and empty source. These are synthetic local transport proofs; no hosted Box allocation or deletion is claimed. The remaining bot changelog-authoring objection was explicitly rejected because this is a maintainer-owned entry required by the repository.

PR #1843, opened 2026-09-04
refactor: share CodeSandbox run finalization

steipete:
## Summary

Move CodeSandbox runs onto the existing provider-neutral sandbox lifecycle. Failed automatic deletion now returns a failed outcome with a kept recovery session instead of success with `kept=false`. Command failures keep their selected exit code through cleanup or timing-write failures, and `--keep-on-failure` applies to cancellation and deadlines. Timing is emitted once after cleanup and includes its duration.

Remove the old finalizer and unused bridge wrappers (100 net production lines). Keep early validation, ownership verification before reclaim, partial-create rollback, native sandbox IDs, archive/mount handling, SDK command/environment transport, and standalone operations in the CodeSandbox adapter. Reused sandboxes remain retained. This does not introduce a claim-revision fence or change the vendor SDK.

## Verification

- Six regression cases fail on the baseline and pass after migration, including a controlled clock proving cleanup precedes the final timing record.
- Full CodeSandbox/shared Go race suites, scoped vet, docs build, and Linux arm64 CLI build pass.
- Built CLI + unmodified Node bridge + synthetic file SDK boundary + real Linux Bash: baseline 3/6, candidate 6/6. Cases cover success, failed deletion, nonzero exit with failed deletion, explicit keep, reuse, and real `/dev/full` timing failure with keep-on-failure. Final fixture resources, claims, and staged environment files are removed.
- Native proof uses Node v24.20.0 on Linux arm64, pinned image `node@sha256:be23f54a88d34e8824c741b19b91064094f92c1c97b194144bfc8b50d67258e2`, network disabled, read-only root, dropped capabilities and task-only temporary files. It does not use the vendor SDK implementation or a hosted CodeSandbox account; no cloud deletion is claimed.
- Managed Codex review and independent semantic review have no actionable findings. Refreshed against the finalized v0.49.1 source, preserving that release section and the original v0.49.0 notes.

## Configuration and risk

No new dependencies, credentials, configuration, command syntax, or provider permissions. This deliberately changes warning-only cleanup success into failure and preserves primary failures; callers consuming run status/retention should now receive the truthful result. Warmup/stop/pause/resume and their authorization boundaries are unchanged.

Complete native reproduction, observed baseline/candidate results and maintainer compatibility disposition: https://github.com/openclaw/crabbox/pull/1843#issuecomment-5547553934. Cancellation/deadline cause and retention currently have deterministic provider/shared regression coverage; the six-case native fixture does not independently exercise those two cases.

PR #1862, opened 2026-09-05
fix: retain cancellation causes for interrupted local bridges

steipete:
## Summary

Preserve the caller's cancellation/deadline cause when the local command runner actually interrupts a child that then exits by signal. The error previously reached provider bridges only as `signal: killed`, so a real canceled CodeSandbox CLI run was classified as a provider failure. Fix the shared executor, not each bridge or lifecycle adapter.

Capture provenance only in os/exec's context-watcher cancellation callback. Wait synchronizes that callback; join its caller cause only for a negative/signaled process exit. File-capture watchers and post-Wait cleanup retain the raw stop callback, so later cleanup cannot fabricate cancellation provenance. Preserve observed nonnegative process exits, startup errors, output-limit exit5, cleanup ownership and capture limits.

## Verification

Five native subprocess regression cases fail before the fix: ordinary/bounded pipe cancellation and file-capture cancellation/early-close/deadline. They pass after the fix; the focused suite passes10 repetitions. Existing capture/streaming/graceful cancellation/retained-writer/controller-group tests pass. A late-cancel fixture verifies observed exit23 remains primary whether the retained output writer settles or times out. Independent review caught an initially overstrict scheduler assumption in that fixture; it was corrected and repeated.

Shared/procjson, CodeSandbox, SmolVM and Upstash race suites, scoped vet and docs build pass. The broader local CLI race suite hit its15-minute package timeout; the header reports a newly started retained-lease renewal test (0s), with no preceding failed assertion. This is not counted as a pass; focused diagnosis and current-head hosted CI are pending before landing. Managed Codex and independent semantic review have no actionable findings.

Actual Linux arm64 CLI + production Node bridge + real Bash, composed with the CodeSandbox lifecycle changes in https://github.com/openclaw/crabbox/pull/1843: real SIGTERM scenarios pass2/2 after this fix versus0/2 without it. Both default cleanup and keep-on-failure now report canceled; exact fixture resource/claim/env cleanup and container removal are verified. The SDK boundary is synthetic, container network is disabled, and this is not hosted CodeSandbox or vendor-SDK certification.

## Scope and risk

POSIX signal-termination fix only: Windows forced termination reports a nonnegative process code and remains unchanged. A local bridge stopping does not prove its remote workload stopped. No new dependencies, credentials, configuration, provider flags, permissions or deletion authority. Error chains now retain both the original process error and context cause. Frozen v0.49.1/v0.49.0 notes are unchanged; maintainer Unreleased entry added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Local provider bridges: preserve cancellation and deadline causes when a POSIX child is interrupted, without masking completed command exits or output-limit errors. Thanks @steipete.
+
 ## 0.49.1 - 2026-09-04
 
 Users upgrading from v0.48.1 also receive the [v0.49.0 changes](https://github.com/openclaw/crabbox/blob/v0.49.0/CHANGELOG.md#0490---2026-09-03), previously available through the Go module release.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 - Report Azure Dynamic Sessions cleanup failures as failed runs, preserve primary Superserve/Azure errors through cleanup and timing output, and keep Superserve rollback bound to the originally created sandbox. [PR 1836](https://github.com/openclaw/crabbox/pull/1836). Thanks @steipete.
-- Local provider bridges: preserve cancellation and deadline causes when a POSIX child is interrupted, without masking completed command exits or output-limit errors. Thanks @steipete.
+- Local provider bridges: preserve cancellation and deadline causes when a POSIX child is interrupted, without masking completed command exits or output-limit errors. [PR 1862](https://github.com/openclaw/crabbox/pull/1862). Thanks @steipete.
 
 ## 0.49.1 - 2026-09-04
 
--- a/docs/provider-backends.md
+++ b/docs/provider-backends.md
@@ -482,6 +482,13 @@ envelopes, versions, identity checks, redaction, and provider error semantics in
 the adapter. Do not use it for noisy CLI output, streaming or NDJSON protocols,
 or commands with ambiguous side effects.
 
+The local command runner preserves caller cancellation/deadline causes when
+its context watcher interrupts a child that then exits by signal. It retains
+the underlying process error and does not relabel observed nonnegative exits,
+post-exit capture cleanup, or output-limit failures as cancellation. This is a
+POSIX signal-termination guarantee; Windows forced-termination codes remain
+unchanged. It does not prove that canceling a bridge stops its remote workload.
+
 Vanilla provider HTTP redirect policy also belongs in
 `internal/providers/shared`. `shared.SecureHTTPClient` clones an injected
 client, rejects destinations outside a trusted `shared.SameOrigin`, preserves
--- a/internal/cli/command_capture_unix_test.go
+++ b/internal/cli/command_capture_unix_test.go
@@ -23,22 +23,26 @@ import (
 func TestExecCommandRunnerFileCapture(t *testing.T) {
 	t.Setenv(controllerProcessTreeOwnedEnv, "")
 	for _, tc := range []struct {
-		mode      string
-		code      int
-		stdout    string
-		stderr    string
-		cancel    bool
-		waitDelay bool
-		overflow  bool
+		mode       string
+		code       int
+		stdout     string
+		stderr     string
+		cancel     bool
+		deadline   bool
+		lateCancel bool
+		waitDelay  bool
+		overflow   bool
 	}{
 		{mode: "normal", stdout: "out", stderr: "err"},
 		{mode: "nonzero", code: 7, stdout: "out", stderr: "err"},
 		{mode: "early-close", cancel: true},
 		{mode: "cancel", cancel: true},
+		{mode: "deadline", cancel: true, deadline: true},
 		{mode: "stdout-overflow", code: 5, stdout: strings.Repeat("x", 1024), overflow: true},
 		{mode: "stderr-overflow", code: 5, stderr: strings.Repeat("x", 1024), overflow: true},
 		{mode: "overflow-exit", code: 5, stdout: strings.Repeat("x", 1024), overflow: true},
 		{mode: "inherited-writer", waitDelay: true},
+		{mode: "completed-exit", code: 23, lateCancel: true, waitDelay: true},
 	} {
 		t.Run(tc.mode, func(t *testing.T) {
 			dir := t.TempDir()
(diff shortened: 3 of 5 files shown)


steipete:
## Native signal proof and maintainer disposition

The production change in source `44663031cec44b9ae763cdece7cf23ee58b3ea57` is unchanged in current head `e4af5a06b2ea5a14ccb5e18184771f9f4e2355fa`. A Linux arm64 binary composed from CodeSandbox lifecycle head `acc83e77fb2d6ca739b7e94ad1e18f87054b1fea` plus this executor file passes both actual SIGTERM cases. Without this executor change, both failed only cancellation classification (cleanup/retention themselves already matched expectations). The same Node24.20.0 image, synthetic SDK boundary and real Bash were used. No hosted workload or vendor-SDK certification is claimed.

| Case | Before | After | Automatic deletion |
| --- | --- | --- | --- |
| Default cleanup | exit1; failed/provider-error | exit1; canceled/canceled | original sandbox, once |
| Keep-on-failure | exit1; failed/provider-error | exit1; canceled/canceled | none; retained claim/resource, then explicit stop |

Every case verifies the exact target, resource/claim state, one timing record, final explicit cleanup and absent env files. The disposable, network-disabled container was removed. Full local CLI suite hit its package budget and is not a pass; the timeout investigation/hosted full CI are separate from this native evidence. Existing native subprocess tests also prove real deadline cause, code5 overflow precedence, graceful cancellation and observed exit23 during late cancellation; they pass ten focused repetitions.

As maintainer, I own the linked Unreleased note. This is an intentional POSIX error-chain correction, not a change to Windows exit codes or remote deletion authority.

### Reproduce the composed native check

In a disposable checkout of CodeSandbox PR https://github.com/openclaw/crabbox/pull/1843 at `acc83e77fb2d6ca739b7e94ad1e18f87054b1fea`, save `fixture.mjs` from the [complete synthetic SDK fixture](https://github.com/openclaw/crabbox/pull/1843#issuecomment-5547553934) into `proof/`. Save `cancellation.mjs` below there too. Fetch this PR's head, then use a Go overlay for the one executor source file:

```sh
git fetch origin codex/local-command-cancellation-c396
mkdir -p proof
git show e4af5a06b2ea5a14ccb5e18184771f9f4e2355fa:internal/cli/provider_backend.go > proof/provider_backend.go
python3 - <<'PYTHON'
import json
from pathlib import Path
root = Path.cwd()
(root / 'proof/overlay.json').write_text(json.dumps({'Replace': {
    str(root / 'internal/cli/provider_backend.go'): str(root / 'proof/provider_backend.go')
}}))
PYTHON
GOOS=linux GOARCH=arm64 CGO_ENABLED=0 go build -trimpath -overlay=proof/overlay.json -o proof/crabbox ./cmd/crabbox
docker run --rm --network none --read-only --cap-drop ALL \
  --security-opt no-new-privileges --pids-limit 128 --memory 512m \
  --tmpfs /tmp --tmpfs /project \
  --mount type=bind,source="$PWD/proof",target=/proof,readonly \
  node@sha256:be23f54a88d34e8824c741b19b91064094f92c1c97b194144bfc8b50d67258e2 \
  node /proof/cancellation.mjs /proof/crabbox
```

<details><summary>Complete synthetic signal driver</summary>

```js
import { mkdirSync, copyFileSync, writeFileSync, readFileSync, existsSync, readdirSync } from 'node:fs';
import { join } from 'node:path';
import { pathToFileURL } from 'node:url';
import { spawn, spawnSync } from 'node:child_process';
const binary=process.argv[2];
const records=[];
mkdirSync('/project/workspace',{recursive:true});
for(const keep of [false,true]){
 const root='/tmp/cancellation-'+keep;mkdirSync(root);const repo=join(root,'repo');mkdirSync(repo);
 const module=join(root,'fixture.mjs');copyFileSync('/proof/fixture.mjs',module);
 const env={HOME:join(root,'home'),XDG_STATE_HOME:join(root,'state'),XDG_CACHE_HOME:join(root,'cache'),PATH:'/usr/local/bin:/usr/bin:/bin',CRABBOX_CODESANDBOX_API_KEY:'synthetic-fixture-key'};
 mkdirSync(env.HOME);spawnSync('git',['init','-q',repo],{env});writeFileSync(join(root,'mode.json'),JSON.stringify({failDelete:false}));
 const common=['--provider','codesandbox','--codesandbox-sdk-package',pathToFileURL(module).href,'--codesandbox-bridge-command','node'];
 const leasePath=join(root,'lease.json');
 const args=['run',...common,'--no-sync','--timing-json','--lease-output',leasePath];
 if(keep)args.push('--keep-on-failure');
 args.push('--shell','--','printf started; sleep 30');
 const child=spawn(binary,args,{cwd:repo,env,stdio:['ignore','pipe','pipe']});
 let stdout='',stderr='',sent=false;
 child.stdout.on('data',data=>stdout+=data);child.stderr.on('data',data=>stderr+=data);
 const closed=new Promise((resolve,reject)=>{child.on('error',reject);child.on('close',(code,signal)=>resolve({code,signal}));});
 const timer=setInterval(()=>{
  const path=join(root,'calls.jsonl');if(!existsSync(path))return;
  const calls=readFileSync(path,'utf8').trim().split('\n').filter(Boolean).map(JSON.parse);
  if(!sent&&calls.some(c=>c.op==='command'&&c.command.includes('sleep 30'))){sent=true;child.kill('SIGTERM');}
 },10);
 const watchdog=setTimeout(()=>child.kill('SIGKILL'),15000);
 const result=await closed;clearInterval(timer);clearTimeout(watchdog);
 const calls=readFileSync(join(root,'calls.jsonl'),'utf8').trim().split('\n').map(JSON.parse);
 const lease=existsSync(leasePath)?JSON.parse(readFileSync(leasePath,'utf8')):null;
 const timing=stderr.split('\n').filter(line=>line.startsWith('{')).map(line=>JSON.parse(line));
 const state=JSON.parse(readFileSync(join(root,'state.json'),'utf8'));
 const errors=[];
 if(!sent||result.code!==1||result.signal!==null)errors.push('signal/exit');
 if(!lease||lease.kept!==keep)errors.push('retention');
 if(calls.filter(c=>c.op==='delete').length!==(keep?0:1))errors.push('automatic-delete-count');
 if(calls.some(c=>c.id!=='sb-fixture'))errors.push('wrong-target');
 if(state.deleted===keep)errors.push('resource-state');
 if(timing.length!==1||timing[0].runStatus!=='canceled'||timing[0].errorKind!=='canceled'||timing[0].exitCode!==1)errors.push('timing');
 const claimsDir=join(env.XDG_STATE_HOME,'crabbox','claims');
 const claims=()=>existsSync(claimsDir)?readdirSync(claimsDir).filter(n=>n.endsWith('.json')):[];
 if(claims().length!==(keep?1:0))errors.push('claim');
 if(!state.deleted){const stop=spawnSync(binary,['stop',...common,'--id','csbx_sb-fixture'],{cwd:repo,env,encoding:'utf8',timeout:30000});if(stop.status!==0)errors.push('explicit-cleanup');}
 if(!JSON.parse(readFileSync(join(root,'state.json'),'utf8')).deleted||claims().length)errors.push('cleanup-left');
 if(readdirSync('/project/workspace').some(n=>n.startsWith('.crabbox-env-')))errors.push('env-file-left');
 records.push({keepOnFailure:keep,pass:errors.length===0,errors,signalSent:sent,...result,stdout,stderr,lease,timing,automaticDeletes:calls.filter(c=>c.op==='delete').length});
}
console.log(JSON.stringify({binary,passed:records.filter(r=>r.pass).length,total:records.length,records},null,2));
process.exit(records.every(r=>r.pass)?0:1);

```

</details>

steipete:
Merged as `530d9b586aa3460c2d58e6d115ebaed0edc287da`. All five hosted workflows and23 substantive checks passed, including full Go tests, required Linux enforcement and build; exact-head review found no actionable issues. The actual merge tree matches the verified preview. Post-merge native local-runner/capture tests and shared/procjson/CodeSandbox race suites passed; visible main was fast-forwarded and clean.

Real signal testing exposed a common cause-loss boundary: os/exec reported an interrupted child as `signal: killed`, and adapters could not recover cancellation from that text. The local executor now preserves the caller cause only when its own context watcher interrupted a signal-terminated child. It retains the original process error, leaves completed exits and output-limit errors primary, and keeps post-Wait cleanup outside the cancellation-provenance recorder. Provider adapters do not need their own workaround.

Five native regressions fail before/pass after; focused cases passed ten repetitions. The composed CodeSandbox CLI/production Node-bridge/native Bash check passes2/2 cancellation cases and6/6 ordinary lifecycle cases. That composition includes the still-pending CodeSandbox lifecycle PR https://github.com/openclaw/crabbox/pull/1843; it is not a claim that its finalizer or hosted vendor certification is already on main. Windows forced-termination codes and remote workload cleanup authority remain unchanged.

The separate local full CLI suite reached its15-minute package budget with no preceding assertion failure. Its just-started retained-lease subtest passed alone in0.99s; the audit points to cumulative runtime, not a demonstrated hang. No timeout was raised and that local run is not recorded as a pass. Hosted full-suite success is separate evidence. The maintainer Unreleased note is included and both frozen release sections are unchanged.

PR #1870, opened 2026-09-05
ci: allow cold container bootstrap and stream smoke progress

steipete:
## Summary

Give only the local-container connector smoke a 30-minute job ceiling, retaining 15 minutes for other matrix rows. Reuse the existing live-capture helper for generic warmup and command execution so progress streams while lease, slug and run-ID extraction still works. Child failures preserve their original exit code. No readiness limit, lease policy, lifecycle assertion or cleanup authority is weakened.

If the credential-free hosted container bootstrap fails, capture bounded state and log diagnostics from only the uniquely identified smoke container. Selection requires one exact lease and one full Docker container ID with the expected labels. Missing or ambiguous identity skips diagnosis; no environment, argv, exec or deletion is added. Whole-job cancellation cannot run this diagnostic, and containers already removed by normal cleanup safely skip it.

## Why

The previous 15-minute job ceiling included 2–3 minutes of build time, while ordinary cold SSH bootstrap alone allows 20 minutes. One observed bootstrap completed after 12m14s but the job expired before use and cleanup. The ceiling needs bounded room for the lifecycle the application already permits.

Intermittent bootstrap failures are separate: initial head `59261e8844f31a7001080180ea00e209c17f3d9a` exhausted the existing 20-minute readiness allowance, but an exact-source retry passed. This PR does not claim a root-cause fix for those failures.

## Verification

- Existing real shell/pipe handshake fails against buffered baseline and passes with streamed provisioning output. Failure fixtures preserve exit 42 and clean only known leases.
- Existing workflow tests execute the actual inline diagnostic against missing, ambiguous and exact container identities. All six focused tests pass, including the new behavior test. The preceding full related script run passed 91 tests; a new broad local recheck encountered an unrelated existing XCP-ng fixture hang, which is being investigated separately.
- Actual Docker failure fixture: owned container exited 42, the diagnostic captured its State and bounded logs, then the fixture container was removed.
- Native local Docker lifecycle completed warmup, status, inspect, SSH, cache, sync, command and stop. This local macOS/ARM64 control is not substituted for hosted Linux proof.
- Hosted Linux Docker run https://github.com/openclaw/crabbox/actions/runs/33938746283 attempt 2 passed against exact head `59261e8844f31a7001080180ea00e209c17f3d9a`: build 1m59s, full native smoke 53s, total 3m16s. The diagnostic-only follow-up head is undergoing its own checks.
- Managed Codex review passed with no actionable findings at its configured P0 threshold; independent semantic review found no actionable issue. Diff and syntax checks passed.

This is CI/test-harness maintenance: no new dependency, credentials, provider flags or release mutation.

### DIFF
--- a/.github/workflows/connector-e2e-smokes.yml
+++ b/.github/workflows/connector-e2e-smokes.yml
@@ -54,10 +54,11 @@ jobs:
             # Cold bootstrap allows 20m; leave room for build, use and cleanup.
             timeout-minutes: 30
             smoke: |
+              set -o pipefail
               CRABBOX_LIVE=1 \
               CRABBOX_LIVE_PROVIDERS=local-container \
               CRABBOX_BIN="$PWD/bin/crabbox" \
-                scripts/live-smoke.sh
+                scripts/live-smoke.sh 2>&1 | tee "$RUNNER_TEMP/local-container-smoke.log"
 
           # Read-only doctor readiness contract. Hosted runners do not ship
           # Firecracker or CNI plugins, so the script's own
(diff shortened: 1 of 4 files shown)


steipete:
## Native gate diagnosis and local control on 59261e8844f31a7001080180ea00e209c17f3d9a

The hosted job now streams progress correctly and reaches the existing full 20-minute SSH-readiness deadline. It fails with exit5, not the new 30-minute Actions ceiling. Pull finished before readiness polling; the log never establishes successful SSH transport, so neither slow package installation nor an authentication defect is proved. All other applicable exact-head CI/release/CodeQL gates passed. The hosted connector gate remains unresolved.

A newly built CLI from this head completed both real local Docker paths with the same pinned Ubuntu image (`docker.io/library/ubuntu@sha256:2260313b31c8c011cd2eebe728008efac1b3982be73eb71348ea2648d2c0e09b`):

- Fixed-ID warmup reached ready in 41.741s, then synced and ran a synthetic checkout and stopped successfully. The fixed-ID terminal tombstone is retained by design; the native container and SSH key are gone.
- The actual `scripts/live-smoke.sh` ordinary-allocation path passed warmup, status/inspect/SSH/cache, sync/run (`crabbox-live-ok`) and native stop. Exact lease/container inventory is empty afterward.

The CLI SHA256 is `589e6861137f733a503d21cbb95cc0df0a5a34b97b96eda947a4fd4a3376ca2e`; ordinary-smoke log SHA256 is `abc1e95623bd17824a550fefa9040fc2886054044ffb2576ac0fba013f4a423d`. The checkout and credentials/state were isolated; workload data was synthetic. This is real local Docker proof on macOS/arm64, not a claim that the hosted Linux/amd64 failure is fixed.

A follow-up diagnostic patch is prepared but not pushed: on failure it derives the unique lease from this job's smoke output, selects only the exact labeled full container ID, and prints State plus a bounded bootstrap log. It does not dump Config.Env/argv, delete a container, alter readiness, or acquire another claim lock. The actual diagnostic shell was exercised against an owned native container that exited42, correctly reporting its state/log, then that fixture container was removed. All91 related script tests pass. Managed review is still pending before that patch is committed/pushed.

PR #1750, opened 2026-09-02
fix(daytona): bound control-plane HTTP requests

SebTardif:
## Summary

Rework the original shared-runtime timeout into a Daytona-owned control-plane default. Direct Daytona API requests now have a 60-second whole-request timeout, including response-body reads. Injected clients retain their policy; earlier caller deadlines still win. Shared runtime clients, toolbox command execution, and archive upload clients are unchanged.

The original global client deadline would also cap E2B streams and Daytona uploads. Keeping the default in the Daytona adapter fixes the finite-control-request gap without imposing that limit on data-plane operations. Existing exact ownership, bounded recovery/rollback, readiness, and snapshot barriers remain intact.

Includes maintainer-added Unreleased notes and Daytona lifecycle documentation. Thanks @SebTardif for the original report and timeout proposal.

## Verification

- Baseline RED: the real pinned SDK control request had no deadline. Fixed default/supplied/untimed/earlier-caller cases pass.
- Full Daytona race suite: `go test -race -timeout=5m ./internal/providers/daytona -count=1` — passed in 14.336s.
- E2B control/data-plane client and long-stream/upload regressions — passed under the race detector.
- Accepted-mutation timeout fixtures use a real HTTP server: one sandbox create or snapshot submission, exact identity recovery, confirmed deletion and claim removal on successful rollback, retained exact claim on delete failure, and no premature source restart for unresolved snapshot completion.
- Actual built CLI against direct Daytona, with no coordinator: created one small sandbox from the public `daytona-small` image, uploaded a synthetic file and verified its SHA-256, completed a 65.284-second command, propagated a separate workload exit 7, and explicitly stopped the lease. Independent inventory, exact native lookup (404), and local claim absence all confirmed cleanup. No snapshot or persistent provider configuration was created.

Observed direct-native output:

```text
sync candidate: 1 files, 9.8 KiB
sync complete in 2.649s
c1e04fd1695eae5308be1608a8f5cadaea2c10d2466df40228055c1e9f50ffb1  payload.txt
native-command-over-control-budget-ok
daytona run summary sync=2.649s command=1m5.285s total=1m8.543s exit=0
synthetic-failure-seven
daytona run summary sync=0s command=331ms total=1.773s exit=7
```

Observed built-CLI control timeout behavior (synthetic localhost API, actual 60-second default):

| Case | Exit | Observed elapsed | Result |
|---|---:|---:|---|
| Healthy inventory | 0 | 0.514s | `[]` |
[...]

PR #1371, opened 2026-08-16
fix(e2b): stalled API host can hang sandbox create and inspect

SebTardif:
## What problem this solves

When `Runtime.HTTP` is unset, E2B used one fallback HTTP client for two different ownership domains:

- finite lifecycle JSON (`create`, `connect`, `get`, `list`, and `delete`); and
- envd uploads and process response streams.

An unbounded fallback can hang a lifecycle call forever. The original `ResponseHeaderTimeout` proposal only bounded the wait for headers: it did not stop a lifecycle response that sent headers and then stalled its JSON body, and it imposed the provider's deadline on envd data-plane handshakes.

This is the same control/data ownership split established for CubeSandbox in https://github.com/openclaw/crabbox/pull/1229.

## What changed

- The fallback control client now has a 60-second whole-request timeout, covering connection setup, response headers, and complete finite JSON bodies.
- The fallback envd client has no whole-request timeout; upload and process-stream lifetime remains owned by the caller context.
- `doJSON` and every lifecycle method use the control client.
- `UploadFile` and `StartProcess` use the envd client.
- An injected `Runtime.HTTP` remains the exact same client object for both paths, preserving its transport, proxy, TLS, redirect, jar, and timeout behavior.
- Both private fallbacks leave `Transport` nil so Go continues to use the process-level default transport.

Thanks @SebTardif for identifying the hang and contributing the original patch.

## Source-blind loopback behavior proof

The E2B test artifact was compiled, then executed from an isolated directory containing only the binary and behavior contract—no source checkout. Short injected fixture deadlines avoid real-minute tests.

```console
=== RUN   TestE2BControlClientBoundsStalledResponseBody
    control response-body stall bounded: elapsed=41ms timeout=40ms
--- PASS: TestE2BControlClientBoundsStalledResponseBody (0.04s)
=== RUN   TestE2BControlClientBoundsWithheldHeaders
    control withheld headers bounded: elapsed=41ms timeout=40ms
--- PASS: TestE2BControlClientBoundsWithheldHeaders (0.04s)
=== RUN   TestE2BDataPlaneStreamOutlivesControlTimeout
    envd stream survived control deadline: elapsed=92ms control_timeout=30ms
--- PASS: TestE2BDataPlaneStreamOutlivesControlTimeout (0.09s)
=== RUN   TestE2BDataPlaneUploadOutlivesControlTimeout
    envd upload completed without truncation: bytes=12 elapsed=92ms control_timeout=30ms
[...]

PR #1229, opened 2026-08-03
fix(http): bound finite provider control calls

SebTardif:
## Summary

- Finite provider control calls no longer hang indefinitely when no runtime HTTP client is injected.
- Retains bounded fallback clients for Azure ARM, CubeSandbox management, FastAPI Cloud, GitHub Codespaces, Hostinger, Morph, Railway, RunPod, Sprites, and Vast.
- CubeSandbox keeps separate defaults: management calls are bounded, while file transfers and process streams remain caller-bounded.
- Upload, download, SSE, and long-execution clients keep caller-controlled lifetimes; Daytona remains unchanged.

## Verification

- Focused loopback behavior for CubeSandbox and FastAPI Cloud
- Full provider and CLI test suites
- Deadcode analysis

Thanks @SebTardif for identifying and implementing the original hardening.

Config/secrets: none.

PR #1881, opened 2026-09-05
fix(cli): give runtimeForApp a timed HTTP client

SebTardif:
## What Problem This Solves

Fixes an issue where users running `crabbox run`, `connect`, `status`, and other product commands against HTTP providers could hang forever when the provider control plane stopped responding. Those commands load backends through `runtimeForApp`, which left `Runtime.HTTP` unset. Providers such as opencomputer, crownest, superserve, opensandbox, unikraftcloud, and cloudrunsandbox then fell back to `http.DefaultClient`, which has no timeout.

The same hole exists on `RuntimeForProviderOperation`, used for provider lifecycle work that is not already bound to a configured backend.

## Why This Change Was Made

Give both constructors a shared 60-second `http.Client`, matching the finite-control clients already used in Azure and Hetzner helpers. A single `defaultRuntimeHTTP()` helper keeps the two paths from drifting.

This does not clone `http.DefaultTransport`. https://github.com/openclaw/crabbox/pull/1750 already bounded Daytona control-plane calls inside the Daytona adapter. This change covers the remaining unset `Runtime.HTTP` constructors on the product CLI path.

## User Impact

Stalled provider control-plane requests on the main CLI path now fail after 60 seconds instead of hanging until the process is killed. Operators get an error they can retry or diagnose. No config or flag changes.

## Evidence

Live `go` program calling the exported provider-operation runtime after the patch, plus the built `crabbox` binary from this tree:

```console
$ go run .
HTTP timeout: 1m0s
HTTP is DefaultClient: false
DefaultClient timeout: 0s
```

```console
$ crabbox --help
Crabbox leases remote test boxes, syncs your dirty checkout, runs commands, and cleans up.

Usage:
  crabbox <command> [flags]
  crabbox run [flags] -- <command...>
```

The same helper is used by unexported `runtimeForApp`, which is what `run`, `connect`, and `status` pass into `loadBackend`. Before the patch, `HTTP` was nil on both constructors, so those providers used `http.DefaultClient` (`Timeout` 0s, no deadline).

## Real behavior proof

- **Behavior or issue addressed:** Product CLI runtimes omitted HTTP, so provider control calls could hang forever on a stalled host via `http.DefaultClient`.
- **Real environment tested:** Windows 11, Go 1.27.0 windows/amd64, worktree at `fix/runtime-timed-http-client` based on origin/main [`6172f518`](https://github.com/openclaw/crabbox/commit/6172f518).
- **Exact steps or command run after this patch:** Built `crabbox` from this tree with `go build`, then ran a small `go` program in the same module that calls `cli.RuntimeForProviderOperation` and prints the client timeout next to `http.DefaultClient`.
- **Evidence after fix:** terminal output from the patched tree:

  ```console
  $ go run .
  HTTP timeout: 1m0s
  HTTP is DefaultClient: false
  DefaultClient timeout: 0s
  ```

- **Observed result after fix:** The exported provider-operation runtime now carries a 60-second HTTP client that is not `http.DefaultClient`. DefaultClient remains unbounded (0s), which is the hang that providers used when HTTP was nil.
- **What was not tested:** Live opencomputer, crownest, or superserve control-plane outage. Multi-minute Daytona archive uploads and E2B envd streams against a real host.

## Related

- https://github.com/openclaw/crabbox/pull/1229
- https://github.com/openclaw/crabbox/pull/1370
- https://github.com/openclaw/crabbox/pull/1371
- https://github.com/openclaw/crabbox/pull/1750
- Introduced on `runtimeForApp` in [`494f3a4d`](https://github.com/openclaw/crabbox/commit/494f3a4d) (2026-05-06)
- `RuntimeForProviderOperation` added in https://github.com/openclaw/crabbox/pull/1405

### DIFF
--- a/internal/cli/provider_backend.go
+++ b/internal/cli/provider_backend.go
@@ -1677,8 +1677,14 @@ func routeConfiguredProvider(cfg *Config) error {
 	return nil
 }
 
+const defaultRuntimeHTTPTimeout = 60 * time.Second
+
+func defaultRuntimeHTTP() *http.Client {
+	return &http.Client{Timeout: defaultRuntimeHTTPTimeout}
+}
+
 func runtimeForApp(a App) Runtime {
-	return Runtime{Stdout: a.Stdout, Stderr: a.Stderr, Clock: realClock{}, Exec: execCommandRunner{}}
+	return Runtime{Stdout: a.Stdout, Stderr: a.Stderr, Clock: realClock{}, HTTP: defaultRuntimeHTTP(), Exec: execCommandRunner{}}
 }
 
 const (
--- a/internal/cli/provider_exports.go
+++ b/internal/cli/provider_exports.go
@@ -25,7 +25,7 @@ func RuntimeForProviderOperation(stderr io.Writer) Runtime {
 	if stderr == nil {
 		stderr = io.Discard
 	}
-	return Runtime{Stdout: io.Discard, Stderr: stderr, Clock: realClock{}, Exec: execCommandRunner{}}
+	return Runtime{Stdout: io.Discard, Stderr: stderr, Clock: realClock{}, HTTP: defaultRuntimeHTTP(), Exec: execCommandRunner{}}
 }
 
 // ProviderSelectionIsAuthoritativeRoute reports whether cfg names an exact
--- a/internal/cli/runtime_http_test.go
+++ b/internal/cli/runtime_http_test.go
@@ -1,0 +1,27 @@
+package cli
+
+import (
+	"io"
+	"testing"
+	"time"
+)
+
+func TestRuntimeForAppSetsTimedHTTPClient(t *testing.T) {
+	rt := runtimeForApp(App{})
+	if rt.HTTP == nil {
+		t.Fatal("runtimeForApp HTTP is nil")
+	}
+	if got, want := rt.HTTP.Timeout, 60*time.Second; got != want {
+		t.Fatalf("runtimeForApp HTTP Timeout = %v, want %v", got, want)
+	}
+}
+
+func TestRuntimeForProviderOperationSetsTimedHTTPClient(t *testing.T) {
+	rt := RuntimeForProviderOperation(io.Discard)
+	if rt.HTTP == nil {
+		t.Fatal("RuntimeForProviderOperation HTTP is nil")
+	}
+	if got, want := rt.HTTP.Timeout, 60*time.Second; got != want {
+		t.Fatalf("RuntimeForProviderOperation HTTP Timeout = %v, want %v", got, want)
+	}
+}


steipete (changes requested):
Thanks for following through on the timeout work. The remaining issue is the scope of the default: a client installed on Runtime.HTTP is also an injected data-plane client, not only a control-plane client.

In current main, E2B's e2bHTTPClients returns the injected client for both control requests and envd operations. Daytona's archive upload and toolbox paths also inherit Runtime.HTTP. A global 60-second http.Client.Timeout therefore caps legitimate longer commands, response streams, and archive uploads even when their own operation budget is longer; that timeout includes reading the response body. This reverses the control/data separation preserved in https://github.com/openclaw/crabbox/pull/1750.

Please keep the runtime injection contract unchanged and narrow the fix to a demonstrated finite provider request that escapes its operation-specific bound. Nil Runtime.HTTP by itself is not evidence of an unbounded operation: adapters also impose request contexts and SDK deadlines. The named examples need to be checked at that boundary against current main.

The regression test should drive the production runtime/backend wiring: a stalled finite request must end, while a stream or upload that legitimately outlives the control timeout must finish, and caller cancellation must still work. Printing the constructor's timeout and running --help do not establish those behaviors.

I cannot land https://github.com/openclaw/crabbox/pull/1881 with this global timeout because of the long-operation regression. A focused adapter fix with that behavior proof remains welcome.

SebTardif:
@steipete

> I cannot land https://github.com/openclaw/crabbox/pull/1881 with this global timeout because of the long-operation regression. A focused adapter fix with that behavior proof remains welcome.

Agreed. Closing this PR.

A client installed on `Runtime.HTTP` is also the injected data-plane client (E2B envd, Daytona archive and toolbox). A 60s whole-request `Timeout` would cut a command such as `sleep 90` and long uploads, which is the contract [#1750](https://github.com/openclaw/crabbox/pull/1750) kept.

I rechecked the named nil-HTTP fallbacks on current `main`. Finite control calls already have adapter-owned bounds:

- opencomputer: 2m request context on JSON, separate 1h exec budget
- crownest: 2m request context except the event stream
- superserve, unikraftcloud, cloudrunsandbox: request-context deadlines
- daytona control: 60s client when `Runtime.HTTP` is nil
- islo: `ResponseHeaderTimeout` 30s and `Timeout` 0 so uploads and exec can still run

I did not find a finite control request that escapes those bounds. If one shows up later, I will send an adapter-only patch with stalled-request and long-operation HTTP proof.

SebTardif:
@steipete

Follow-up after the recheck: Nomad control JSON was still unbounded. `Regions()` dropped context, and job inspect / register used only the parent CLI context (cancel on SIGINT, no deadline) on `cleanhttp.DefaultPooledClient()`.

Adapter-only patch: https://github.com/openclaw/crabbox/pull/1916

PR #1931, opened 2026-09-06
perf: start commands without waiting for telemetry uploads

steipete:
## Summary

Commands synchronously waited for the first best-effort telemetry upload after capturing their resource baseline. A slow telemetry endpoint could therefore delay workload admission even though an upload failure did not prevent execution.

Keep baseline capture before work, but publish it through the existing sampler. Start the 15-second ticker before publication, derive each upload's timeout from the sampler context, and cancel and fully join the sampler before finish, failure, or lease replacement. This also removes the one-second wait that could let an upload outlive its owner. Terminal event handling, baseline retention, and signed receipt verification remain with their existing owners.

## Validation

- The blocked-upload admission regression fails on the original code and passes after the repair. Virtual-time coverage also proves the first periodic sample remains at 15 seconds after a slow initial upload.
- Race-tested owner coverage verifies cancellation while the transport still owns work, joins before finish/failure/replacement, duplicate starts, the retained terminal baseline, exact signed receipt commit/verification, and isolation of the next run's samples.
- `go vet ./...`, baseline/candidate builds with Go 1.26.5, and independent P0–P2 review pass.
- A real AWS comparison used one fixed `c7a.8xlarge` lease in `eu-west-1`, the same working directory/configuration and identical harmless no-sync commands. Both normal commands completed, both signed receipts verified, and both stored run summaries retained Linux baseline samples (32 CPUs; two samples each). Canonical Stop confirmed released/cleanup-complete; all command processes joined, SSH control sockets were absent, and operator configuration was unchanged.
- Observed command-preparation phase: 5.366s → 4.766s; runner total: 17.172s → 16.344s; process wall time: 18.323s → 18.158s. This single pair is noisy and is not a general speed benchmark. The controlled blocked-upload regression proves removal of the admission dependency.

Live source provenance: baseline `4e3df6dfc42a49bd665a92b2e2756b909122c8c3`, candidate with telemetry patch SHA-256 `7df3fbc3976486164ab4f4e1f038626519322b023ab8d00ef077dddabc0c2633`. Both were task-built development binaries, not signed releases. The later fixture-only commit does not alter production bytes.

## CI follow-through

[...]

PR #1937, opened 2026-09-07
perf: keep run diagnostics off workload admission

steipete:
## What Problem This Solves

Short remote commands wait for serial best-effort phase uploads before the workload can start. A real AWS diagnostic run spent 8–16 seconds on those phase calls. Output publication also used a timed wait that could return while its publisher was still running.

## Why This Change Was Made

Use one bounded, ordered publisher for phase and stream diagnostics. Existing lease bindings returned by run creation do not need another synchronous acknowledgement; initial and replacement bindings still do. Terminal recording drains diagnostics, cancels remaining requests and joins the actual publisher before committing the terminal result.

## User Impact

Workloads can start while diagnostic publication proceeds. Lease attribution, retained command logs and verified signed terminal receipts remain required; overflow and drain expiry produce a warning. This removes duplicate event state, the separate synchronous phase path and the detached wait goroutine.

## Evidence

- Real HTTP blocked-phase reproduction fails on the parent and passes with the repair.
- Original-order cancellation/join, existing binding, lease replacement, coordinator changes, output and terminal ordering checks pass under the race detector.
- Broader run lease/history/coordinator/receipt/failure race coverage passed in 213 seconds; focused final owner race passed in 5.413 seconds.
- Build passes; independent review is clean through P2.
- Real AWS ABBA comparison passed on one c7a.8xlarge lease with identical payload, configuration and workdir. Parent/candidate/candidate/parent wall times were 39.080 / 19.508 / 16.509 / 18.792 seconds; time to workload output was 22.714 / 14.480 / 8.155 / 11.247 seconds. Both candidate runs retained all nine events, complete logs and verified signed receipts, with zero warnings or diagnostic drops. The two samples per version show variance and are not a broad performance guarantee; the blocked-POST reproduction establishes the causal admission fix.
- All four runs retained the correct provider/lease/workload identity and telemetry baseline. The lease was released with cleanup complete, all owned processes joined, the SSH control directory disappeared, and operator configuration and the synthetic workdir remained unchanged.

Production delta: +45 lines for the acknowledged barrier and actual cancellation/join ownership; tests +287; docs/changelog +15. No new configuration, protocol or persistent-store fields. Related: https://github.com/openclaw/crabbox/pull/1931.

## Review follow-through

Required lease binding now drains/cancels and joins pending diagnostics before its own acknowledged HTTP request. Full diagnostic capacity and an 8-second prior diagnostic plus a healthy 3-second binding no longer reject admission; the request keeps its existing 10-second budget without sharing it with queue time.

The legacy missing-events concern is preserved where it can work: already-bound run creation remains usable without optional events, and sync-only explicitly continues with a warning when history binding is unavailable. Unbound or replacement command execution still requires a recorded lease/provider/slug tuple, as required by the signed terminal receipt contract shipped in v0.47; allowing that command would merely defer failure until receipt verification. New tests explicitly check these outcomes instead of discarding AttachLease errors. The required-binding queue findings and corresponding rank-up moves are addressed.

Final recorder/app-sync/terminal/receipt race coverage passed, broader lifecycle race coverage passed in 186.648s, and full-PR independent P2 review is clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Add `crabbox bench check` to enforce sample, failure, and p95 runner timing policy across every matched local benchmark group. [PR 1899](https://github.com/openclaw/crabbox/pull/1899). Thanks @vincentkoc.
 - Distinguish required-artifact, artifact-change, and artifact-schema validation failures from workload exits in run timing and failure digests, preserving exit 7 and artifact enforcement. [PR 1934](https://github.com/openclaw/crabbox/pull/1934), [Issue 1894](https://github.com/openclaw/crabbox/issues/1894). Thanks @coygeek.
 - Tailscale: keep provider diagnostics and OAuth credentials out of preflight and tag-ownership errors while preserving operation, HTTP status, and actionable tag guidance. [PR 1940](https://github.com/openclaw/crabbox/pull/1940).
+- Publish run diagnostics without delaying workload admission, while preserving ordered lease attribution and joining pending publication before terminal recording. [PR 1937](https://github.com/openclaw/crabbox/pull/1937).
 
 ## 0.51.0 - 2026-09-06
 
--- a/docs/features/history-logs.md
+++ b/docs/features/history-logs.md
@@ -39,7 +39,12 @@ not automatically redacted.
 
 Phase and stream diagnostics publish through one bounded queue while the workload
 continues. An existing lease's run creation already records its binding; a new
-or replacement lease binding is acknowledged before command admission. Before
+or replacement lease binding first drains and joins diagnostics, then gets its
+own acknowledged request before command admission. Missing diagnostic endpoints
+do not block an already bound run; a changed binding must be accepted because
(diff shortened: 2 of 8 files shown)


steipete:
Maintainer landing decision: retain the acknowledged binding barrier for new and replacement commands. A command whose required provider/lease/slug attribution cannot be recorded cannot satisfy the existing signed terminal receipt contract; stopping before workload execution gives it a correct, immediate outcome. Already-bound creation and sync-only retain their documented optional-diagnostics behavior. This accepts the earlier failure point identified by the review and addresses its remaining rank-up move.

The required binding keeps its own unchanged request budget after the diagnostic publisher is canceled and joined. Full-queue, delayed diagnostic, missing endpoint, replacement and terminal-order tests pass; independent P2 review is clean. Exact-head CI for 9bc07fb3f9fe3777a9f7d4dac7237f894c1ea1c6 is green, and the real AWS comparison preserved logs, events, signed receipts and confirmed lease cleanup.

PR #1756, opened 2026-09-03
fix(images): restore prior default after failed publication smoke

vincentkoc:
## What Problem This Solves

Fixes an issue where image publishers could leave a failed AWS Linux, Windows, or macOS image selectable after its normal promoted-image smoke failed. It also prevents explicitly retired AWS defaults from reappearing through an old rollback receipt when no catalog rows remain.

## Why This Change Was Made

Transactional AWS image promotion uses a dedicated fail-closed coordinator route, returns the exact prior default plus every alias the promotion overwrites, and restores only aliases still owned by the failed revision. The coordinator rejects `expectedCurrent` and rollback retirement fields on every non-transactional promotion route before provider dispatch, so legacy `/promote` remains unconditional and cannot weaken the compatibility boundary. Publisher rollback passes the captured receipt back to the coordinator.

An authorized rollback restores regional-key absence and cross-region fallback aliases, preserves any concurrent newer alias, and retires only the exact failed image ID and revision. Exact revision tombstones remain authoritative after same-image republication. Explicit retirement now records a cutoff even when the image survives only in a receipt: a global cutoff for regionless retirement, or the existing regional key for scoped retirement.

Receipt validation and publication share the effective maximum of global and regional cutoffs. One bounded global record per image also retains the last publication timestamp. Both timestamps participate in the existing catalog transaction: retirement includes every earlier publication, even when the clock is frozen or moves backward, while genuine later publication is stamped after the effective cutoff. No public receipt fields, config, schema version, TTL, or cloud API were added. The redundant affected-region discovery was removed; exact-revision tombstones and provider-deletion retry order remain intact.

The publisher wrappers arm rollback before invoking promotion and attempt recovery on every later error or signal while preserving the original exit. The transactional route shares the managed-checkpoint mutation fence and canonical coordinator lifecycle queue, so deleting or pending checkpoints are rejected and concurrent image lifecycle mutations remain serialized before promotion or Fast Snapshot Restore work. Catalog-only promotion rejects transactional expected-current and retirement flags at the CLI, coordinator route, and direct provider API, preserving its non-default contract. Non-AWS providers reject the transactional route before provider mutation.

The trusted qualification admission, adapter, execution, evidence, and matching docs landed first in prerequisite https://github.com/openclaw/crabbox/pull/1879 at `41f5039df911abe1061eeafb524752d9afad3640`. This PR retains that existing base without another rebase. No qualification or deployment surface changed in the receipt-only retirement repair.

## User Impact

[...]

PR #1956, opened 2026-09-07
fix(ci): admit qualification executor before candidate execution

vincentkoc:
## What Problem This Solves

Resolves a problem where image qualification cannot safely admit the hosted executor to its fixed SSH security group: the actual executor address is only known on the machine that later runs the candidate, while candidate code must never receive network-owner credentials.

Admission and teardown also need one durable owner so a delayed authorization response cannot recreate SSH ingress after finalization.

## Why This Change Was Made

Trusted preflight now registers the same executor's Cloudflare-observed IPv4 before downloading candidate bytes. Protected admission and credentialless execution start concurrently; a distinct run-bound capability records one immutable address and is consumed before the candidate proceeds.

The protected network owner persists intent, creates one tagged TCP 22 `/32` rule, verifies its exact readback, and durably arms execution in the existing authority Durable Object. Candidate and signer permissions do not gain security-group writes; `execute` receives no AWS or Cloudflare credentials, reusable controller capability, or privileged OIDC permission.

Cleanup independently attempts relay, compute, network, and candidate teardown. Recovery reconciles ambiguous authorization by canonical tags, persists an exact owned-rule receipt before revocation, validates the AWS revoke response, and durably records that acknowledgment before absence can clear the intent. A dispatched write without that proof remains pending regardless of elapsed time or empty inventory reads. The authority centrally stamps the trusted workflow attempt on compute/image/volume/key/snapshot creation and tagging; candidates cannot override it.

## User Impact

Operators get executor-specific SSH admission and durable cleanup evidence without world-accessible SSH or candidate network authority. This is a source-only prerequisite, not a cloud deployment or qualification run.

Before any apply or execution, the infrastructure owner must separately verify the dedicated sandbox envelope, durable state and recovery bindings, exact IAM policy/boundary, and unattended protected credential delivery. The network principal needs narrowly scoped tag-on-create permission, not standalone tagging. No private infrastructure repository or deployment target is changed here.

Hosted network credentials must be a short-lived STS session with its actual expiration and protected finalizer/reaper custody. Admission requires validity through run expiry plus 30 minutes. The maximum run is 120 minutes; a three-hour session leaves 30 minutes for provisioning. The reaper runs every 10 minutes with a 20-minute job bound, without treating scheduling as a cleanup guarantee. A local SSO login does not establish hosted custody.

## Evidence

- The original admission patch received independent source review. The follow-up unresolved-ingress repair follows an independently accepted plan and awaits renewed source review. Its cumulative 14-file SHA-256 manifest is `cbcf4fed2d9b2d0a209dab337a839e5ab49ee0123ffd41368ec03c596d49c236`; committed blobs are verified against the tested bytes.
- 124 focused tests passed on the repaired source: 78 Node workflow/network/reaper tests and 46 authority tests.
- Regression coverage includes a dispatch reply delayed until after finalization, insufficient remaining native-call time, a real three-hour credential lifetime, immutable/consumed executor registration, independent cleanup, and canonical attempt tags with candidate override rejection.
- The persisted dispatch fence is 60 seconds. Before an ingress write, the helper reserves its complete 25-second native-call budget inside both the absolute fence and run expiry.
- Test mocks capture 10 actual normalized authority requests and 16 native network requests for policy integration. Network creation still uses exactly six canonical string tags: run, attempt, owner, SHA, expiry, and `crabbox_qualification_network`. IAM permissions, credential duration and dispatch deadlines are unchanged.
- Worker typecheck passed: `npm run check --prefix worker`.
- Authority dry-run build passed: `npm run build:aws-qualification-authority --prefix worker`; Wrangler reported `--dry-run: exiting now`.
- Scoped lint, formatting, shell syntax, whitespace and privacy checks passed. No dependencies were installed or cloud resources changed. The two existing lint warnings in untouched controller-tool code remain outside this repair.

### Review Follow-up

The P1 in https://github.com/openclaw/crabbox/pull/1956#issuecomment-5568392938 was reproduced before repair: cleanup returned success after eight empty reads and zero revocations for an unresolved dispatch. It now fails closed and retains durable recovery.

New tests cover hidden ingress beyond every cleanup poll; later owned visibility with receipt persistence before deletion; failed persistence; successful revocation with a lost controller reply; ambiguous AWS revocation; foreign-rule preservation; and independent compute/candidate cleanup with controller and registry retention. An actual controller-handler, controller-entrypoint, Durable Object and network-helper chain verifies permitted effects and rejected token, deployment, run, intent, consumed-capability and finalization paths. The delayed-dispatch regression still proves no late AWS authorization and now requires retained recovery.

Prepared-but-never-dispatched intents may clear after fencing. Existing runs without network state retain the compute-only cleanup route. The added durable receipt uses the existing run Durable Object, with no new namespace.

These are deterministic source and mocked-provider boundary tests, not a live AWS/Cloudflare qualification claim. Authorized live boundary proof, coordinated rollout, state/recovery binding and independent hosted custody remain separate pre-merge/execution gates. No merge or deployment is requested here.

Focused test commands:

```sh
node --test scripts/image-qualification-network.test.js scripts/image-qualification-workflow.test.js scripts/image-qualification-reaper.test.js
npm test --prefix worker -- test/aws-qualification-authority.test.ts
bash -n scripts/image-qualification-execute.sh
git diff --check
```

Net change: 794 runtime lines, 20 workflow lines, 1,130 test lines, and 90 documentation lines. The follow-up repair adds 123 runtime lines to retain unresolved effects and persist verified cleanup facts. Runtime growth establishes the separate protected network owner and durable admission/cleanup invariant; it does not add a second authority namespace or broaden candidate privileges.

Related image work:
- https://github.com/openclaw/crabbox/pull/1756
- https://github.com/openclaw/crabbox/pull/1942


vincentkoc:
@clawsweeper re-review

The follow-up signed commit `9858cc093f68044d50c84d4cfda01439082df908` repairs unresolved ingress retirement and adds durable receipt, late-visibility, lost-reply and controller-to-authority boundary regressions. The PR body contains the updated 124-test proof and source manifest. Please review this exact head; no autofix, merge or deployment is requested.

steipete:
Closing after maintainer review. The gap is real only for dynamically assigned hosted executors; Crabbox's documented model is a deliberately pre-admitted fixed executor (docs/behavior/aws-image-qualification.md), and this proposal would require a dedicated AWS policy, short-lived credentials, independent reaper custody, real address provenance, and live admission/cleanup proof that do not exist today, on top of a 540-commit rebase against the current qualification code. The recovery protocol design (preserving unresolved ingress despite empty inventory reads) is sound and worth keeping in mind if we ever adopt dynamic executors. Thanks @vincentkoc for the thorough analysis.

Issue #1962, opened 2026-09-07
Run admission can time out and leave an unacknowledged history record


steipete:
## Problem

An ordinary command against an already-ready AWS lease can fail before execution because the coordinator run-create request exceeds its 10-second client deadline. The CLI correctly refuses execution without a coordinator run handle. Subsequent lease-filtered history can contain a running/starting record whose handle the caller did not receive, leaving outcome attribution unresolved.

## Observed evidence

This occurred during a verified image-fork setup using CLI 0.51.0, and again during one ordinary `uname -s` command using clean source `f650d3749227a6587d03967493a307edc6925072`. In the latter case AWS creation and warmup succeeded on the expected Worker version; the command exited 7 before execution when POST `/v1/runs` timed out. Canonical lease/instance/key cleanup completed. No command replay, anonymous run fallback, or manual history rewrite was attempted.

The source routes run admission through a shared lifecycle queue. Existing-image operations can hold that queue across provider I/O. This is a plausible wait boundary, not proof of which wait caused either incident. The current command-marker representation is insufficient to correlate all historical stdin-script attempts by command hash alone.

## Recovery design checkpoint

The CLI already has a caller-known execution identity before admission, but run creation uses an independently generated server identity. A repair should make the original create recoverable through authenticated exact-binding replay/readback, preserve live execution-context checks, and commit the initial run and event under one atomic owner. It must not replay the command, bypass required run tracking, blanket-remove event serialization, or merely increase the timeout. Persistent admission semantics need maintainer design acceptance before implementation.

Related: https://github.com/openclaw/crabbox/pull/1958 (a separate provisioning-latency change; it does not claim to fix this admission problem).

steipete:
Implemented and landed in https://github.com/openclaw/crabbox/pull/1970 as `f5dfb62977ce9ba4b87d5d68da37bd110f2d8d74`.

The CLI now retains a caller-known admission identity and original request; authenticated recovery returns that exact record. Run creation and its first event commit atomically. The client rejects canceled, mismatched, terminal, or already-progressed admissions and never replays remote execution as part of recovery. Existing POST clients remain supported; the new CLI requires coordinator-first deployment.

[...]

PR #1970, opened 2026-09-07
fix: recover lost run admission responses

steipete:
## What Problem This Solves

Fixes an issue where an ordinary command could lose its coordinator run identity when admission committed but the response was lost. The CLI could only fail without a handle, and later history could not reliably identify the exact attempt.

Related: https://github.com/openclaw/crabbox/issues/1962.

## Why This Change Was Made

The CLI now chooses a cryptographically random run ID before admission and uses a dedicated PUT route that supports recovery of the exact original request. The coordinator binds that request to its initiating owner and organization and commits the run and first event together. Recovery returns the existing record without appending another start event. A separate acknowledged event retains the existing lease-attribution and replacement behavior.

The CLI accepts only a matching, still-starting record within the live invocation. Admission recovery stays within the existing ten-second request budget and never retries remote execution. Cancellation, deterministic refusals, mismatched replies, terminal records, and coordinator-origin changes fail closed. Sync-only runs retain optional history.

## User Impact

Lost admission responses can be recovered without creating another run or replaying the command. The attempt ID is printed before admission for correlation. Current CLI builds require the new coordinator route; old coordinators reject it without anonymous creation. Existing clients retain POST support, and legacy run IDs remain readable. No configuration option, schema-version bump, dependency, provider policy, or retention change is introduced.

This does not promise ten-second cloud startup or remove the lifecycle queue. Existing historical records are not rewritten.

## Evidence

- Original-main regressions demonstrated the absent caller-known route, partial run persistence when initial-event storage failed, terminal/advanced record admission, and lost cancellation.
- Worker after refreshing over current main: 2,872 passed, three existing skips; Cloudflare and Node types, lint, formatting, deployment dry run, and docs build passed.
- Controlled client proof drops a committed admission response, then verifies identical identity/body recovery, one record, cancellation, unsupported coordinators, and a shared ten-second budget.
- Focused Go race checks and all six affected CLI integrations passed. Independent formal review found no actionable P0–P2 findings.
- Full local Go race run: 91 packages passed; CLI hit its unchanged twenty-minute package cap and reported five stale route/ID fixtures. All five were corrected and passed a focused race rerun. This is not a passing full local suite.
- Windows amd64 CLI test binary cross-compilation and `go vet ./...` passed.
- Live AWS on the reviewed candidate: ordinary Linux warmup passed (128.812s), `uname -s` returned one `Linux` line (12.685s for the run), and the same 128-bit ID matched acknowledgment, lease handle, verified signed receipt, and successful terminal history. There was one initial run event. Canonical Stop confirmed released/provider cleanup complete and removed owned keys. No workload was replayed.
- The first private proof harness read `runID` instead of the documented `runId`; read-only receipt/events/history checks recovered the proof without rerunning the command. The original harness failure is retained.
- Exact-head CI passed on `bdafeade5e76a73990709f574a45ebc9a0b32683`, including the full hosted Go race gate, coverage, all Go modules, Worker builds/tests, Windows checks, connector smokes, and the release check. The failed local full run remains documented above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Recover lost run-admission responses using a caller-known identity and an atomic initial history record, preserving authenticated request binding and refusing remote command replay. Current clients require a coordinator supporting the new admission route. [Issue 1962](https://github.com/openclaw/crabbox/issues/1962).
+
 ## 0.52.0 - 2026-09-07
 
 ### Highlights
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -233,6 +233,7 @@ Runs and observability:
 ```text
 GET  /v1/runs
 POST /v1/runs
+PUT  /v1/runs/{run-id}
 GET  /v1/runs/{run-id}
 GET  /v1/runs/{run-id}/logs
 POST /v1/runs/{run-id}/events
@@ -287,7 +288,9 @@ retries as admin.
 progress to the broker so the portal and `history`/`logs`/`events`/`results`
 commands can read it back:
 
-- `POST /v1/runs` creates a `RunRecord` in state `running`.
+- `PUT /v1/runs/{id}` atomically admits a caller-known run and its first event,
+  or returns the retained record for the same caller and original request.
+  Legacy `POST /v1/runs` creates a coordinator-issued `RunRecord` in state `running`.
 - `POST /v1/runs/{id}/events` streams phase-tagged events: `run.started`,
   `leasing.started`, `bootstrap.waiting`, `sync.started`/`finished`,
   `actions.hydrate.*`, `command.started`, stdout/stderr chunks,
--- a/docs/features/coordinator.md
+++ b/docs/features/coordinator.md
@@ -225,6 +225,7 @@ PATCH  /v1/checkpoints/{id}/retention
 POST   /v1/checkpoints/{id}/use
 DELETE /v1/checkpoints/{id}
 POST   /v1/runs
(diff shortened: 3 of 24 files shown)


steipete:
Maintainer decision: the caller-known admission/recovery design and coordinator-first rollout are approved. The reviewed coordinator was deployed through the supported workflow before running the new CLI. No CLI release or publication is part of this PR; existing clients retain the legacy POST route.

There is new live evidence after the review's captured PR body: the current candidate ran an ordinary AWS Linux command successfully. Its 128-bit identity matched the acknowledgment, retained lease handle, verified signed terminal receipt, and successful terminal history. There was one initial run event. Canonical cleanup confirmed the lease released, provider cleanup complete, and owned keys absent. This used the existing durable coordinator store after upgrade, and the metadata capture remained on the exact deployed version.

Response-loss injection is deliberately confined to controlled tests. The real HTTP client test drops a committed response and checks identical request recovery; coordinator tests cover original-request and actor binding, duplicate admission, terminal replay, and atomic rollback when the initial event write fails. The live test verifies the deployed storage/API/CLI/command contract without injecting failures into the shared coordinator. No historical records were rewritten. I accept this proof coverage for the repair; a combined production fault-injection transcript is not an additional merge requirement in the absence of a concrete defect.

Fresh-state cases use the coordinator storage test implementation; the live run used an existing durable namespace. No SQL schema migration or retention change is introduced. The PR body now records the live result, the corrected private-harness field-name error, and the failed local full-suite attempt separately from passing focused proof. Required exact-head CI still gates merge.

PR #1992, opened 2026-09-08
refactor(cua): share configured fallback defaults

steipete:
## What changed

CUA's typed configuration already declares its defaults, but the provider repeated eight configured values and Python repeated two import defaults. Runtime consumers now use the generated constants directly; the old local aliases and unused region default are deleted.

A whitespace-only `sdkFallbackImport` is accepted configuration: previously Go trimmed it to empty and Python selected `cua_sandbox`. Resolve that effective choice in Go before both JSON and environment transport, then remove the redundant Python terminal defaults. Preserve raw configuration, request-over-environment precedence, custom imports, and every other field's existing blank-before-trim behavior.

The fixed 15-second doctor budget and Python's version check for the actual `cua_sandbox` module are separate contracts and remain unchanged. Provisioning/deletion restrictions, authentication, endpoint selection, SDK loading and read-only inventory behavior are unchanged. No new config source, credential, dependency, or deployment is required. The landing maintainer has added the changelog entry and architecture documentation.

## Verification

- Baseline `b3bf78151061501e0f51f3dd557ef5ac419bda94` and candidate pass the same five focused compatibility tests: defaults/raw input, empty/whitespace/custom launcher and workdir, mixed-case kind, effective Python import selection, request/environment precedence, and the independent doctor budget. Baseline production files were checked against Git; no skips.
- Candidate-only assertions verify the effective fallback reaches both transport channels without changing the raw input.
- `go test -race -timeout=10m ./internal/providers/cua -count=1 -v`: passed, no skips, with network access denied. The existing read-only rejection tests passed too.
- `go vet ./internal/providers/cua`, `go build -trimpath ./cmd/crabbox`, docs build, and `git diff --check`: passed.
- Managed Codex review found no accepted/actionable P0 findings in the full candidate. Maintainer source review also checked the Go producer and Python consumer together.

The actual baseline and candidate binaries were separately run through `doctor --provider cua --json` with credentials excluded, isolated config/home/state, and networking denied. Their complete JSON results, stderr and exit codes matched for all three cases:

| Settings | Before and after |
| --- | --- |
| Defaults | Exit 1; SDK diagnostic names `cua` and `cua_sandbox` as missing. |
| Whitespace-only fallback import | Same complete result as defaults. |
| Custom preferred and fallback imports | Exit 1; SDK diagnostic preserves both configured module names. |

The CLI test used the unmodified embedded Python script, not an injected SDK fixture. The separate import-selection unit test intercepts SDK import deliberately. Neither proof installs or exercises the real SDK, verifies hosted credentials, or qualifies provider inventory/lifecycle. Missing SDK is an environment result, not a passing hosted smoke.

CI passed on `10be4c2ab9280accfdcbb388527a0bc038835b65`: main CI (including Go aggregate), release checks, connector smokes, docs proof and CodeQL all succeeded. [Main CI run](https://github.com/openclaw/crabbox/actions/runs/34226836085). The original watcher exited successfully without reruns. A superseded ClawSweeper dispatch was cancelled, but its exact-head review completed separately with no findings and sufficient terminal proof. This is not a hosted-SDK qualification, release or deployment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@
 
 - Describe OpenSandbox configuration bindings once, preserving CLI-only stale-claim cleanup, environment/flag-only API URL selection, defaults, and validation timing. [PR 1990](https://github.com/openclaw/crabbox/pull/1990). Thanks @steipete.
 
-- Keep CUA runtime and Python bridge defaults aligned with declared configuration, preserving custom SDK imports, whitespace fallback behavior, and read-only diagnostics. Thanks @steipete.
+- Keep CUA runtime and Python bridge defaults aligned with declared configuration, preserving custom SDK imports, whitespace fallback behavior, and read-only diagnostics. [PR 1992](https://github.com/openclaw/crabbox/pull/1992). Thanks @steipete.
 
 ## 0.52.0 - 2026-09-07
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@
 
 - Describe OpenSandbox configuration bindings once, preserving CLI-only stale-claim cleanup, environment/flag-only API URL selection, defaults, and validation timing. [PR 1990](https://github.com/openclaw/crabbox/pull/1990). Thanks @steipete.
 
+- Keep CUA runtime and Python bridge defaults aligned with declared configuration, preserving custom SDK imports, whitespace fallback behavior, and read-only diagnostics. Thanks @steipete.
+
 ## 0.52.0 - 2026-09-07
 
 ### Highlights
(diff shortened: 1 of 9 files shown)


steipete:
Landed as `cca8b490735524610f2b527a31906e8124378e07` through the ordinary protected squash merge. The raw merge parent is `b3bf78151061501e0f51f3dd557ef5ac419bda94`, and tree `290e717dfedccd5d78c260108833cd00e5660037` exactly matches the reviewed head.

CUA now takes its configured runtime defaults from one declaration. The Go bridge preserves whitespace-only fallback-import behavior before passing the effective value through JSON and environment, so Python no longer needs a second default. Custom imports, request precedence, the fixed doctor budget and read-only restrictions remain unchanged.

After syncing main, the complete CUA race suite passed again with networking denied (`go test -race -timeout=10m ./internal/providers/cua -count=1`, 2.481s). The rebuilt main binary has SHA-256 `0d9daab930d3b5ac90a2316dcd1ec30b3562308ab33572e5959d5c2a7e162823`, identical to the candidate used for the three actual CLI/Python diagnostic comparisons. This is local diagnostic proof, not hosted SDK readiness. All substantive CI workflows passed; the superseded review dispatch cancellation did not prevent the exact-head review from finishing with no findings. Main is clean.

PR #2046, opened 2026-09-09
fix(aws): avoid slow fallback when an instance type has no capacity

steipete:
## What Problem This Solves

Fixes an issue where users creating a public AWS runner would wait through repeated requests to an unavailable instance type before an already configured alternative was tried. A live allocation returned `InsufficientInstanceCapacity` for its preferred Spot type; the transport recorded twelve signing attempts across two instance-creation operations, taking 43.2 seconds in that stage.

## Why This Change Was Made

The AWS HTTP client retries server errors before the provisioning owner can inspect their AWS error codes. This change hands a complete, exact `RunInstances` capacity rejection back to that owner when another permitted type or market passes the existing quota preflight. It preserves the existing candidate selection, client token and resource-outcome handling.

The capacity probe is limited to 64 KiB and one second. Opaque, malformed, oversized, stalled or unreadable bodies keep normal retry handling. Generic server errors, rate limits, transport failures, exact types, macOS, private workspaces and final candidates retain their existing behavior.

## User Impact

Eligible allocations reach a configured fallback sooner instead of repeatedly requesting unavailable capacity. A permitted alternative may therefore win earlier; no new fallback, configuration option or dependency is introduced.

## Evidence

- Real local HTTP plus the production AWS signer/client and full candidate fallback reproduced eleven requests to the unavailable type before the next candidate. The repaired HTTP 500 and 503 capacity cases make one request per candidate.
- 191 focused tests passed, including retry preservation, exact/pinned/private launches, candidate and market boundaries, signing/network failures, and bounded response probes that leave the original body readable.
- Worker lint, formatting, typechecking and Wrangler dry-run build passed.
- Two fresh cloud desktop runs passed Computer input while WebVNC remained connected. Each had 150 health samples with the node continuously present and no OOM kills; both were reclaimed and provider cleanup verified. These establish the live baseline, not deployed performance of this patch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Changes
 
+- AWS: hand definitive instance-capacity rejections directly to an already configured type or market fallback, avoiding repeated requests to the unavailable capacity while retaining exact-type, macOS, private-workspace, and transient-error retries.
 - Describe Scaleway configuration bindings once, sharing configured defaults while preserving explicit SDK location overrides and distinct file, environment, and flag list behavior. [PR 2036](https://github.com/openclaw/crabbox/pull/2036). Thanks @steipete.
 - Preserve recorded broker network diagnostics in inspect/status JSON, including SSH source CIDRs and AWS placement fields, without changing the resolved network mode or adding provider requests. https://github.com/openclaw/crabbox/pull/2039. Thanks @vincentkoc.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Changes
 
-- AWS: hand definitive instance-capacity rejections directly to an already configured type or market fallback, avoiding repeated requests to the unavailable capacity while retaining exact-type, macOS, private-workspace, and transient-error retries.
+- AWS: hand definitive instance-capacity rejections directly to an already configured type or market fallback, avoiding repeated requests to the unavailable capacity while retaining exact-type, macOS, private-workspace, and transient-error retries. [PR 2046](https://github.com/openclaw/crabbox/pull/2046). Thanks @steipete.
 - Describe Scaleway configuration bindings once, sharing configured defaults while preserving explicit SDK location overrides and distinct file, environment, and flag list behavior. [PR 2036](https://github.com/openclaw/crabbox/pull/2036). Thanks @steipete.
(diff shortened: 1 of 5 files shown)


steipete:
Maintainer disposition for the changelog finding: this is maintainer-authored work, and I am retaining the release note as the landing maintainer. The root AGENTS.md explicitly asks maintainers and agents to add user-visible fixes to Unreleased as work lands. Removing this entry would undo that required maintenance. All entries inherited from main are preserved, and the note links this PR.

The refreshed head resolves the earlier base conflict. The review confirms the capacity handoff is correct and identifies no production-code or proof blocker. The current-head Worker check and all connector smokes have passed; the remaining CI and required workflow must finish before normal merge.

steipete:
Deployment and live-validation follow-up: the automatic [coordinator deployment](https://github.com/openclaw/crabbox/actions/runs/34380079684) completed, and the deployed Worker version was verified at 100% traffic. Both PR CI and [CI on the merged main commit](https://github.com/openclaw/crabbox/actions/runs/34380079542) passed.

Seven additional cases passed in local workerd with the production compatibility settings, using the unchanged production AWS signer, transport, and fallback owner against a synthetic HTTP service. They covered success, complete HTTP 500/503 capacity responses, generic retries, dripping and oversized bodies, and UTF-8 decoding. Original response bodies remained readable after probe cancellation.

The first deployed AWS checkpoint-fork smoke returned Cloudflare error 1101 before readiness. Its provider cleanup was subsequently confirmed through the existing recovery owner, and a final normal stop completed local cleanup. The exception did not reproduce in the bounded workerd diagnostic and remains unexplained.

After that cleanup completed, one fresh fixed-ID AWS checkpoint fork passed readiness and normal cleanup on the same deployed Worker. The fork became ready in 155.031 seconds. Provider diagnostics reported 8.285 seconds for provisioning and 1.612 seconds for the single instance-creation request. All 77 captured invocations matched the deployed Worker version; none contained an exception (76 were OK, and one GET near cleanup was canceled).

The fresh attempt obtained its preferred instance type immediately, so it does not establish live capacity-handoff savings. The controlled HTTP and workerd cases prove that branch. This provider-only smoke also does not measure full cloud desktop or model-turn startup. Both test leases are fully cleaned up, and all diagnostic helpers have exited.

PR #2052, opened 2026-09-09
refactor(config): unify Linode bindings and configured defaults

steipete:
## Summary

Describe all five Linode file/environment settings once, preserving the flagless interface, raw nonempty file lists, environment parsing and explicit image/type markers. The typed initializer takes core's already-resolved image—including an empty result—instead of repeating OS resolution. Configured fallback values remain separate from native class selection and validation order.

All five file fields use the shared value-storage mode from https://github.com/openclaw/crabbox/pull/2053, retaining the original omission of ignored empty strings and lists during user-config writes. Empty provider blocks remain `{}`. Provider operations are unchanged; provider/refactor docs and the Unreleased note accompany the change.

## Verification

- Original binding characterization passed before and after the refactor. The writer follow-up first reproduced retained empty Linode keys, then passed with unchanged regression fixtures after the five-field correction.
- Eight focused configuration/writer tests, all 80 generator tests (including 32 freshness checks), vet, static analysis, docs build and independent review passed on the prepared follow-up.
- After merging actual main, eight focused tests, 32 freshness checks, vet and build passed again. The guide/test conflicts were resolved using the exact reviewed Linode additions; source and merge state stayed unchanged during the checks.
- The integrated binary passed 17 frozen public writer cases: seven intended omissions and ten unchanged cases. All 58 original CLI rendering/configuration cases also matched, including four preserved parser errors. No fixtures or expected outputs were adapted to a candidate result.

The original public-head CI attempt failed during dependency installation in three jobs because of a Google Chrome apt package-index hash mismatch. Those results remain recorded; checksum checks were not weakened. This updated head requires fresh normal CI before landing.

The existing PR history is retained through ordinary merges and a fast-forward update, without force-pushing. Local checks used Go 1.26.5, cleared inherited environment, network denial and synthetic configuration paths. This is bounded local configuration verification, not live Linode provisioning or account qualification.

PR #2053, opened 2026-09-09
fix(config): preserve provider file storage semantics

steipete:
## Summary

Restore the saved-YAML behavior of 147 historically value-backed settings across 26 providers. Generated pointer fields had started retaining ignored empty strings, empty lists and numeric zero values when ordinary commands such as `config set-broker` rewrote user configuration. Effective overlays were unchanged for these fields, but the persisted representation drifted.

Add explicit, per-field `fileStorage:"value"` metadata. It is admitted only with an existing zero-ignoring file rule; the normal pointer representation remains the default. The generator emits the original concrete file type and equivalent raw-value predicate—without a custom serializer, duplicated field roster or changes to source trust, environment handling, flags, runtime defaults or native behavior.

The cohort was checked against original migration parents. All 78 originally pointer-backed fields stay pointers. Modal's empty secrets list also remains presence-preserving so a save does not discard an explicit clear. Empty provider blocks remain `{}`, and ignored negative numeric values remain stored verbatim. Five pending Linode fields will adopt the shared repair after this lands: https://github.com/openclaw/crabbox/pull/2052.

## Verification

- The actual writer regression tests failed before adoption and passed afterward with unchanged fixtures, including the `config set-broker` handler against synthetic configuration.
- All 79 generator tests passed, including 31 freshness checks and generated fixture tests. Before adoption, the new engine reproduced all 31 existing outputs exactly.
- Exactly 161 existing configuration tests passed: 92 core tests and 69 provider tests. Native, authentication and lifecycle tests were not included in this local configuration gate.
- Scoped vet, static dead-code analysis, reproducible build, docs build and independent Codex/source review passed. Independent reconstruction verified the exact field cohort, preserved pointer fields and unchanged non-file generated code.
- Ten frozen public CLI writer cases passed on both the initial and actual-parent binaries. Outputs, exits and timeouts matched the baseline; saved files matched eight unchanged controls and precisely two intended field-omission changes. Explicit empty blocks and meaningful clear/false/zero settings remain intact.

The first generator-test baseline exposed three int64 fixture setup omissions; those fixtures were corrected and requalified before production changed. The black-box oracle initially misread empty blocks as removable; this was corrected and frozen before any candidate was supplied, with the original oracle and all captures retained. No assertion was adapted to a candidate result.

The branch is based on the actual retained-lease/Mac-host merge. Fresh actual-parent runs passed the 161 configuration tests, writer regressions, vet, static analysis, build and CLI cases; the original evidence remains separate. Final public-head CI is still required before landing.

Local Go checks used Go 1.26.5 with inherited environment cleared and networking denied. CLI cases used synthetic paths and no credentials or provider requests. This is configuration/persistence proof, not hosted-provider qualification. Changelog and the typed-binding guide are updated.


steipete:
Landed at https://github.com/openclaw/crabbox/commit/5334a7b6cad0c4f14a76eb344be1c035f0923860. The actual merge matches the tested combined tree and expected parent. Post-merge writer regressions and all 31 generated-file freshness checks passed; the rebuilt CLI matches the tested combined candidate byte-for-byte (SHA256 `6428354093b73f46063aaac83fe7347d652e5970f5459ca498c7cf08134c046e`). Source and index stayed unchanged during these checks.

The repair restores ignored-empty/zero omission for the 147 audited fields while retaining explicit empty blocks, meaningful pointer values and Modal's list clear. Earlier validation remains scoped and separate: 79 generator tests, 161 existing configuration tests, vet, static analysis, docs build, independent review, and ten actual local CLI writer cases. Those CLI cases preserve eight controls and make exactly two intended field-omission changes; they are not hosted-provider qualification.

Exact final-head CI completed with 24 successful checks and one skipped. Automated review supplied no verdict after an input-safety rejection; that was not treated as approval or retried. The independent review and normal repository merge checks were used. Earlier baseline-fixture/oracle corrections and the stopped no-Go comparison against an older documentation snapshot remain recorded; none was relabeled as a passing execution.

The changelog and binding guide are included. Linode's five-field follow-up remains separate in https://github.com/openclaw/crabbox/pull/2052.

PR #2104, opened 2026-09-11
fix(config): share passive provider display fields

steipete:
## Summary

Fix missing Freestyle and Crownest values in offline config inspection and establish a shared provider-owned display path. Each provider returns ordered, explicitly selected and already-redacted fields for JSON and text. Core collects and renders those records without inferring visibility from runtime structs or configuration-input tags.

Both new sections appear even when another provider is selected. Freestyle preserves relative workdirs and zero sizing, with loaded API-key presence only. Crownest preserves empty project IDs, zero timeout and false cleanup preferences without looking up credentials. URL redaction reuses the existing owner. No Configure, Doctor, SDK, CLI, file or network operation is added by the presenters.

The collector rejects duplicate section keys, text labels, field names and canonical coverage. JSON insertion checks existing top-level keys before mutation; the text path reserves legacy line names without evaluating JSON values or adding environment reads. A source-derived test keeps those names aligned with the actual formatters. Text write failures and short writes now propagate through the command. Existing legacy lines keep their order, and the two new lines appear immediately before the unchanged inspection/status block.

## Verification

- 55 config-show and generic projection tests, with 101 subtests, passed under the race detector. The two actual provider presenter tests passed separately.
- Twelve actual built-CLI comparisons passed across defaults, selected/unselected providers, trusted file and environment input, raw/empty values, zero/false, synthetic key presence and URL query/fragment redaction. The baseline correctly has neither new section. After removing exactly the two approved JSON members or text lines, every remaining output byte matches baseline, including provider status and legacy line order. JSON values were not reserialized to hide differences.
- The text insertion assertion was clarified before candidate execution to place the new lines before the existing inspection notice; original preparation and baseline evidence were retained. A concurrent test-only edit was recorded separately from the immutable binary source copies.
- Go vet, 249 Markdown link checks, whitespace checks and managed Codex review passed at the requested P0 threshold.

## Remaining coverage

This is the first stage, not a complete provider-display migration. The full target remains 80 distinct sections for 81 canonical providers, preserving Apple sharing. After these two additions, 27 providers still lack value sections and three lack text sections; the existing legacy projections must also migrate without changing field names, types, redaction or text layout. The migration contract is documented in docs/refactor/provider-config-display.md.

No native-provider, authentication or readiness proof is claimed. No real credential, hosted API, deployment or release operation was performed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
 - Validate sparse-checkout and skip-worktree sync scope before SSH lease work, explain how to materialize or exclude missing tracked files, and rebuild the final manifest after acquisition. [PR 2097](https://github.com/openclaw/crabbox/pull/2097), [Issue 1567](https://github.com/openclaw/crabbox/issues/1567). Thanks @coygeek.
 - Include the resolved `workroot` in SSH-backed inspect/status JSON, including native Windows and WSL2 leases. [PR 2069](https://github.com/openclaw/crabbox/pull/2069). Thanks @steipete.
 - Show Phala settings in offline configuration text and JSON, preserving unset versus explicit attestation settings without invoking the provider CLI. [PR 2099](https://github.com/openclaw/crabbox/pull/2099). Thanks @steipete.
+- Show Freestyle and Crownest settings through a shared offline JSON/text display path, preserving zero and false values while redacting URLs and reporting only loaded key presence.
 
 ### Maintenance
 
--- a/docs/commands/config.md
+++ b/docs/commands/config.md
@@ -64,6 +64,14 @@ means no explicit override; `true` and `false` remain distinct. This is a
 configuration value, not evidence that remote attestation has run or passed.
 Inspection does not read Phala's stored credentials or invoke its CLI.
 
+Freestyle and Crownest also have value sections in both formats, including
+when unselected. Freestyle shows the loaded URL, relative workdir, CPU/memory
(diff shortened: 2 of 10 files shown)


steipete:
Merged and verified at `0f1d21619281c9a533b81d981aa70cfd1baa587e`.

The provider-owned display capability now supplies the Freestyle and Crownest value sections without making the core renderer discover credentials, apply defaults, or contact providers. Existing inspection/status output and legacy provider output remain unchanged. This is the foundation for moving the remaining display definitions to their providers, not a claim that the whole migration is complete.

Exact-merge verification passed 57 focused tests with 101 subtests, full Go vet, and 12 built-CLI display cases. The CLI checks validated the ten new field values and confirmed unchanged legacy bytes after removing only the two expected new sections/lines. These are passive configuration-display checks, not native provider lifecycle proof. Managed review passed at P0 scope.

Post-merge CI initially failed while fetching the Daytona toolbox client ZIP from proxy.golang.org with an HTTP/2 INTERNAL_ERROR during Vet; there was no source-level vet diagnostic. After preserving and reviewing the original full logs, one failed-jobs retry passed: Go test and its aggregate reran successfully, while nine other successes were carried forward. Connector E2E, CodeQL and Pages also passed separately. Unrelated cancelled dispatch checks were not treated as successes or review approval.

CI: https://github.com/openclaw/crabbox/actions/runs/34627084405

PR #2112, opened 2026-09-11
fix(config): show delegated provider settings offline

steipete:
## Summary

Expose all 34 loaded settings for OpenComputer, OpenSandbox and CUA through their own passive display sections. JSON uses `openComputer`, `openSandbox` and `cua`; text uses the corresponding lowercase labels. Sections remain visible when the provider is unselected.

The shared collector/renderers consume explicitly chosen provider fields, replacing the need for separate future core JSON and text implementations. API URLs use existing redaction; integer zero, explicit false, empty strings and raw resource-limit strings retain their configured meaning. No credentials, external CLI config, SDK installation or provider resources are discovered. CUA bridge/package/import fields are unresolved references, not readiness evidence, and do not enable provisioning.

Docs and Unreleased notes describe the new values and their limits. This adds three missing sections; 24 missing value sections and three missing text sections still remain, alongside migration of existing legacy display owners.

## Verification

The same three actual-provider tests first failed because the display capability was absent, then passed with the complete 34-field contract. Tests cover explicit output keys/types/order, raw values, selected/unselected input, zero/false and shared URL redaction. Broader configuration race tests, Go vet and docs checks passed. Managed Codex review passed at P0 scope.

Eight real built-CLI calls used frozen synthetic inputs, an isolated HOME, empty PATH, denied network access and bounded execution. Four scenarios were run in both JSON and text:

| Scenario | Outcome |
| --- | --- |
| Loaded defaults | All 34 values/types and three sections matched the prewritten contract. |
| Nondefault file settings and API URL environment aliases | Custom fields were shown; URL query/fragment content was removed. |
| Explicit zero, false and empty inputs | Raw configuration distinctions were retained. |
| Whitespace-bearing environment strings | Non-URL strings were preserved without normalization. |

All eight calls exited zero with no timeouts. After removing only the three independently validated new JSON members/text rows, every legacy output byte matched the baseline; JSON was not reserialized for comparison. The flag-only `ForgetMissing` fields were observed as default false in this config-show-only CLI corpus; no nondefault CLI claim is made.

Baseline source: `0f1d21619281c9a533b81d981aa70cfd1baa587e`; baseline binary SHA-256: `cb754b63f71dd263f8e5a69c61aa2b97c86db88eee26fe72fc17c43a37828ef3`; initial candidate binary: `dae0fd3fd8b110188a190ba2e6d1971895c0fa2b3780aa65a0abd220e28b3dfc`. Later integration retains the exact presenter implementation. These are configuration-display checks, not native provider or authentication proof.


steipete:
Merged and verified at `82d8a6064d4ca1acb25758be8a4e4960570e7a62`.

All 34 loaded OpenComputer, OpenSandbox and CUA settings now have provider-owned offline value sections. The shared display path preserves configured zero/false/raw values and redacts URLs without discovering credentials, resolving external defaults or executing SDK bridges. The independent Daytona and checksum-retry changes were retained during integration.

Exact-main verification passed 62 config-show tests and 101 subtests with the race detector, plus all eight frozen built-CLI cases. The latter validate the three new sections and their values, then confirm unchanged legacy bytes after removing only those approved additions. The original baseline was not rerun. Flag-only ForgetMissing remains default-false coverage in that CLI corpus; no native or authentication proof is claimed.

All eleven post-merge CI jobs, five connector checks, four CodeQL analyses and both Pages jobs passed without reruns. Managed reviews passed at P0 scope. The separate automated review produced no verdict and was not treated as approval or retriggered.

CI: https://github.com/openclaw/crabbox/actions/runs/34638071950

PR #2114, opened 2026-09-11
fix: stop labeling Daytona snapshot forks with the default class

### LINKED ISSUE #2109: daytona: fork lease labels record the default class instead of the snapshot's actual resources; plus /proc reports host CPU/memory to workloads
Two small findings from a Daytona provider test pass. Both are low severity; filed together since neither warrants its own issue.

## 1. Fork lease labels record `class=beast` while the sandbox runs the snapshot's resources

A `checkpoint fork` created without an explicit `--class` records the CLI default (`beast`) in the lease labels, even though the forked sandbox inherits the snapshot's actual resources (here: `daytona-small`, 1 vCPU / 1 GiB).

### Reproduction

```bash
# fork a checkpoint captured from a small-class sandbox, without passing --class
/tmp/crabbox checkpoint fork chk_<id> --slug cbx-fork-test

# inspect the resulting claim
cat ~/Library/Application\ Support/*/crabbox/claims/cbx_<fork-lease>.json
```

### Observed

Claim labels for the forked lease:

```json
"labels": {
  "class": "beast",
  "crabbox": "true",
  "lease": "cbx_<fork-lease>",
  ...
}
```

while the sandbox actually enforces the small tier (verified inside it):

```
cgroup cpu.max:    100000 100000     → 1 CPU
cgroup memory.max: 1073741824        → 1 GiB
```

Note that an *explicit* `--class` is validated correctly and rejected on mismatch, with a good error message:

```
Daytona snapshot <snapshot-id> must be active container with 4 CPU, 8 GiB memory, 10 GiB disk and no GPU for class=beast
```

So only the implicit-default labeling path is affected. Suggested fix: label forked leases from the resolved snapshot resources (or omit `class`) rather than from the CLI default, since the label feeds cost accounting and later class validation.

## 2. `/proc` exposes host CPU/memory inside sandboxes, so build tools oversubscribe

Inside a `small` sandbox the cgroup limits are applied correctly, but `/proc` is not virtualized:

```
cgroup cpu.max:     100000 100000          → 1 CPU   (correct)
cgroup memory.max:  1073741824             → 1 GiB   (correct)
nproc:              64                     ← host
/proc/meminfo:      MemTotal 791870644 kB  ← host (~755 GiB)
```

This is a property of the sandbox isolation layer rather than a configuration issue, and is not expected to change, so it needs handling on the consumer side.

### Why it matters for Crabbox users

Anything that sizes itself from `nproc` / `os.cpus()` / `/proc/meminfo` will oversubscribe a 1-core, 1 GiB box by 64x: `make -j$(nproc)`, Node-based test runners (jest/vitest, pnpm worker defaults), and Rust/Go build parallelism. That leads to thrashing or OOM for exactly the workloads people run on a leased box.

### Suggested mitigation

Crabbox already knows the requested class and therefore the real limits, so it could export sensible defaults into the run environment, e.g.:

```
GOMAXPROCS=<vcpu>
MAKEFLAGS=-j<vcpu>
UV_THREADPOOL_SIZE=<vcpu>
NODE_OPTIONS=--max-old-space-size=<~75% of memory MiB>
```

Happy to open a PR for either item.

Patrick-Erichsen:
## What changed

Fixes the misleading Daytona snapshot-fork labels reported in https://github.com/openclaw/crabbox/issues/2109. Snapshot-driven allocations omit an inherited, unverified class; an actual class selection or validation keeps its label. Resource counts do not uniquely identify a canonical class, so the fix does not infer one or resize the snapshot.

This maintainer update integrates current main while preserving the contributor implementation and current provider/core boundaries. It also strengthens the existing explicit custom-snapshot regression to check create, sandbox, and persisted claim labels. Fixed-ID replay keeps the original allocation, fingerprint, and ownership intent. Thanks @Patrick-Erichsen.

The documentation explains caller-controlled concurrency and memory profiles. Crabbox does not inject workload tuning, change host virtualization, or install anything. There are no new dependencies, credential fields, or configuration/fingerprint formats.

## Verification completed

The same focused tests fail on current-main production code in the four ordinary/fixed inherited/custom fork-label cases and pass on the integrated candidate. The baseline uses a two-production-file Go overlay with unchanged candidate tests; it is not a full old-release binary run.

- Focused race tests: 3 top-level tests and 6 subtests passed, including preserved explicit-class controls.
- Full `go test -race -count=1 -timeout=20m ./internal/providers/daytona`: 93 top-level tests and 215 subtests passed, with no skips.
- `go vet ./internal/providers/daytona`, documentation links, and `node scripts/build-docs-site.mjs`: passed.
- Independent managed Codex review of the complete six-file diff against main `9de175962514381677cdace4935c90340115ee4c`: no accepted findings at its selected P0 threshold.

These are source and HTTP-backed provider tests, not live Daytona allocation evidence.

## Still required before landing

Live direct-provider checkpoint capture and ordinary/fixed fork qualification has not run. It must verify omitted unvalidated labels, preserved explicit-class labels, unchanged fixed replay, harmless workload execution, and deletion of every test sandbox and snapshot. A successful coordinator-only doctor or broker smoke does not cover that direct checkpoint path.

This PR remains unmerged pending that native proof and final-head CI. The maintainer changelog entry is retained under Unreleased; published release history is unchanged.

Closes https://github.com/openclaw/crabbox/issues/2109 after the complete fix is verified and landed.


steipete:
Maintainer native-validation update: this is **not ready to land yet**. I refreshed the existing PR head ce247273 onto main a9fd4b5 in an isolated, uncommitted integration (tree ba66dcd76dc65a48a18fbc91e49d48e48d46ecd9); that refresh has not been pushed. Focused regression tests pass on the candidate and fail in the four intended omission cases on main; independent managed review found no actionable P0 findings (P0-only scope).

A real direct-provider run now confirms native checkpoint capture, ordinary and fixed snapshot forks, preserved CPU/memory/disk and captured marker content, absent unvalidated class labels, and harmless command execution. However, both normal stop forms returned 404 for the newly created fixed lease while a fresh exact native read still showed its sandbox running. The native matrix stopped there: explicit-class fork cases, baseline-upgrade replay, and the complete nested intent comparison remain unqualified.

The fixed-stop code is unchanged by the class-label patch. Source inspection points to its pre-delete inventory call through the deprecated `/sandbox/paginated` endpoint. Daytona documents that endpoint's [June 25 retirement](https://www.daytona.io/dotfiles/updates/daytona-is-moving-to-cursor-based-pagination), and the current served schema omits it. This is a strong compatibility lead, not a captured request-level trace of the failing stop. The replacement cursor inventory is documented as eventually consistent, so a URL-only replacement must not silently weaken deletion/absence confirmation.

All four test sandboxes and both sequential test snapshots are now independently confirmed absent by exact-ID HTTP 404 reads. One fixed sandbox needed explicitly scoped, normal provider-native deletion after the CLI failure; that fallback is cleanup, not a passing CLI-stop result. The failed local claim remains private and untouched as evidence. The first snapshot was replaced after a validator marker-sequencing mistake, with its fork and snapshot confirmed deleted before replacement; that procedural correction was not a product failure.

I am retaining the prepared label change and its passing local proof, but withholding push/landing until the ordinary fixed-stop compatibility issue and remaining native acceptance are resolved. No stale-claim recovery behavior, account policy, deployment, or unrelated resource was changed.

steipete:
Superseded by https://github.com/openclaw/crabbox/pull/2421, which landed this fix rebased onto current `main` with the same implementation, your co-author credit on the commit, and a changelog entry thanking you. Verified with the full race suite on a remote Crabbox AWS gate plus CI. Thanks @Patrick-Erichsen!

PR #2111, opened 2026-09-11
fix(daytona): release fixed leases with API keys and reconcile native expiry

steipete:
Fixed Daytona leases could be created with an API key but could not be released: cleanup called an organization endpoint that accepts browser OAuth credentials only. Native TTL expiry also left an acquired fixed lease permanently unresolved after Daytona removed its sandbox.

Use the authenticated organization returned by the current-key endpoint for API-key cleanup and empty-account acquisition. Preserve resource-backed acquisition for older Daytona servers that omit that field. Reconcile an acquired fixed lease through the existing release owner only after checking its original organization, exact native UUID, and complete failure-inclusive database inventory. Inspection records the released tombstone without issuing a provider deletion; ambiguous creates and incomplete observations retain their claims.

Validation:
- Reproduced both failures with released Crabbox 0.56.0.
- Real Daytona CLI proof with the repaired build: fixed acquisition with an API key, explicit and repeated stop, native one-minute TTL expiry, released inspection, and rejection of replaying the retired fixed ID.
- `go test -race -timeout=10m ./internal/providers/daytona -count=1` passed (153.825s); scoped vet and docs generation passed.
- Independent Codex review found no actionable P0–P2 findings.
- The OpenClaw registered-backend live scenario also passed against this build: shared lease creation, token refresh, native pause/resume, preserved remote edits, reopen, recreate, and full cleanup.
- Extended existing HTTP-backed lifecycle coverage for JWT-only organization endpoints, old-server identity compatibility, changed or malformed identity, incomplete inventory, and expired-lease inspection.

This unblocks the lease-backed sandbox integration in https://github.com/openclaw/openclaw/pull/144454. No new configuration, dependency version, or storage format is introduced. Older servers without current-key organization metadata still require an OAuth organization profile for fixed cleanup.

### Inspectable live CLI evidence

Repeated against the final production source tree in `ca25cc1c02cd15e066813524e893c24a2cd27ede`, using an isolated HOME/workspace, the real Daytona API, and an API key. The projection below includes only lease identity, state, readiness, and command exit status; credential-bearing warmup output is omitted. No mocks were used. Both leases were released, and repeated stop also returned zero.

```text
$ crabbox warmup --provider daytona --class small --lease-id cbx_08ef6fc01684 --slug lifecycle-proof --keep --ttl 5m
exit 0
$ crabbox stop cbx_08ef6fc01684
exit 0
$ crabbox inspect --id cbx_08ef6fc01684 --json
[...]

PR #2115, opened 2026-09-11
fix: allow recovery of stale Daytona claims after sandbox deletion

### LINKED ISSUE #2108: daytona: a claim bound to a deleted sandbox can never be released (no stale-claim cleanup path)
## Summary

When the sandbox a claim points at no longer exists, `crabbox stop` hard-errors and refuses to release the claim — permanently. Unlike other providers, the daytona provider exposes no flag to clear a stale claim, so the only remedy is hand-deleting JSON from the Crabbox state directory.

This is not an edge case: the daytona provider **requires** a TTL on every sandbox (`validateDaytonaCreateConfig`: *"provider=daytona requires positive TTL and idle timeout"*), so Daytona destroying the sandbox out from under Crabbox is the normal end-of-life path.

## Root cause

`internal/providers/daytona/backend_ssh.go:462-471` — sandbox resolution, which `stop` goes through:

```go
if cloudID != "" {
    sandbox, getErr := client.GetSandbox(ctx, cloudID)
    if getErr != nil {
        if daytonaIsNotFoundError(getErr) {
            return nil, "", exit(4, "daytona claim %s is bound to missing sandbox %s", claim.LeaseID, cloudID)
        }
        ...
    }
    if sandbox == nil || strings.TrimSpace(sandbox.GetId()) == "" {
        return nil, "", exit(4, "daytona claim %s is bound to missing sandbox %s", claim.LeaseID, cloudID)
    }
```

The behavior is asserted as intended in `internal/providers/daytona/stop_absence_test.go:29-31` for `missing resource`, `different endpoint`, and `different account`. The fail-closed posture is reasonable; the gap is the absence of an operator recovery path.

Other providers expose exactly that, visible in `crabbox stop --help`:

- *"remove the local claim when stop gets 404 (explicit stale-claim cleanup)"*
- *"remove the local claim when stop sees a missing Kubernetes claim"*
- *"remove local claim when the MicroVM is already missing"*

There is no daytona equivalent. The generic `-confirmed-absent-local-cleanup` is marked internal and refuses with: `confirmed-absence local cleanup requires explicit provider, scope, coordinator binding, and complete expected identity set`.

## Reproduction

```bash
# Build from main and authenticate as in the companion issue, then:
mkdir -p /tmp/cbx-repo && cd /tmp/cbx-repo && git init -q
echo hello > marker.txt && git add -A && git -c user.email=t@t -c user.name=t commit -qm init

/tmp/crabbox warmup --provider daytona --class small          # note the slug + sandbox UUID
SANDBOX=<sandbox-uuid-from-output>

# Simulate what Daytona's TTL/auto-delete does routinely
curl -s -X DELETE -H "Authorization: Bearer $DAYTONA_API_KEY" \
     "https://app.daytona.io/api/sandbox/$SANDBOX"

/tmp/crabbox stop --provider daytona <slug>                   # fails
/tmp/crabbox stop --provider daytona <slug>                   # fails identically, deterministic

# the claim is still there
ls ~/Library/Application\ Support/*/crabbox/claims/            # macOS
```

### Observed

```
daytona claim cbx_7f449afb4fff is bound to missing sandbox <sandbox-uuid>
```

Repeated on every retry. The claim file `crabbox/claims/cbx_7f449afb4fff.json` persists with `cloudID` pointing at the deleted sandbox, and had to be removed manually.

## Impact

- Stale claims accumulate for every TTL-expired lease, with no CLI path to clear them.
- Compounds with the fixed-lease 401 issue into a dead end: `stop` first fails with 401, then once TTL destroys the sandbox `stop` fails with "bound to missing sandbox" forever.
- Operators must edit files under the Crabbox state directory by hand.

## Suggested fix

Either:

1. Add a daytona flag matching the sibling providers, e.g. `--daytona-missing-local-cleanup`, that removes the local claim when the bound sandbox returns a well-formed 404; or
2. Treat an application-level 404 for a claim-bound UUID as confirmed absence and drop the claim automatically.

For (2), Daytona's 404 is distinguishable from an intermediary-synthesized one: it always returns `application/json` with a structured envelope, e.g.

```json
{"path":"/api/sandbox/<id>","timestamp":"...","statusCode":404,"error":"Not Found","message":"Sandbox with ID or name <id> not found"}
```

so checking that shape rules out proxy/CDN 404s while still satisfying the fail-closed requirement.

Patrick-Erichsen:
Closes https://github.com/openclaw/crabbox/issues/2108

## What Problem This Solves

Fixes an issue where users stopping an ordinary Daytona lease after native TTL or external deletion had no supported way to remove its stale local claim. Fixed-ID release and expiry reconciliation were repaired separately in https://github.com/openclaw/crabbox/pull/2111.

## Why This Change Was Made

Add command-only `crabbox stop --provider daytona --daytona-forget-missing <lease-id-or-local-slug>`. It checks only the exact sandbox ID bound in the local claim and requires a structured Daytona 404 before removing the unchanged claim under its lock. The operation sends no DELETE and reports local forgetting rather than remote release. Normal stop remains fail closed. The recovery flag is registered only for `stop` (including its `release` alias); it is absent from other commands rather than accepted and rejected later. Brokered mode and other provider selections reject this flag before backend execution.

Fixed replay records, checkpoint-held claims and coordinator/adapter registrations retain their existing reconciliation owners. Live resources, malformed responses, auth failures and transport failures keep the local record. No credentials, dependencies or claim format change; the flag cannot be enabled through YAML or environment variables.

## User Impact

Operators can explicitly recover an ordinary stale claim after checking the original Daytona account and endpoint. A 404 can also indicate lost access; this local-only operation does not prove remote deletion or stop billing. Fixed claims continue to use normal stop and retain terminal replay protection.

## Evidence

Ran the real CLI/provider code against an HTTP fixture with isolated local claim storage, using identical recovery input before and after. No live Daytona resources were allocated.

| Observation | Baseline `2ead1eee` | Candidate |
| --- | --- | --- |
| Recovery command | `flag provided but not defined: -daytona-forget-missing` | Exit 0 after exact-ID structured 404 |
| Local claim | Remains on disk | Removed |
| Provider deletion requests | 0 | 0 |
| Output | No recovery path | `forgot local Daytona claim lease=cbx_123456abcdef sandbox=sandbox-original; provider deletion was not verified` |

- `go test -race ./internal/providers/daytona -run '^TestDaytonaForgetMissing' -count=1 -v` passes. The baseline portable recovery test fails as shown above.
- Brokered CLI regression: the initial candidate incorrectly attempted coordinator release (6 HTTP requests); the final candidate rejects the command before any request. Other command/provider flag combinations are rejected.
- Default-stop preservation tests pass for missing resources, changed endpoint/account and inventory errors.
- Regression cases preserve claims on malformed/incomplete/conflicting 404, 401/403/503, connection failure, live and empty-success responses, absent IDs, checkpoint custody, coordinator/adapter registration and acquired/released fixed IDs.
- Complete provider gate: `go test -race -timeout=10m ./internal/providers/daytona -count=1` passes (108.200s on the final implementation); latest added guard cases also separately pass with race detection.
- `go vet ./internal/providers/daytona ./internal/cli` and `node scripts/build-docs-site.mjs` pass.

- Follow-up command-scope proof on `8c999281`: compiled CLI help shows the recovery flag for `stop`/`release` only, absent from `run`, `warmup`, `status`, `ssh`, and `doctor`. `run --daytona-forget-missing` fails during parsing. Targeted race tests, scoped vet, CLI build, and independent re-review pass.

Independent Standards and Spec reviews are clear after addressing the brokered-route finding in the stale-claim change.

PR intentionally left unmerged. The branch is in the upstream repository, so maintainers can edit it directly (GitHub’s fork-collaboration toggle applies only to cross-repository PRs).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Daytona: add explicit local recovery for ordinary missing-sandbox claims while retaining fixed replay records and rejecting brokered cleanup. [PR 2115](https://github.com/openclaw/crabbox/pull/2115). Thanks @Patrick-Erichsen.
 - Daytona: attest API-key cleanup through current-key organization metadata and retire acquired fixed leases after verified native TTL or external deletion; inspection records a terminal tombstone without issuing deletion or reusing the fixed ID. [PR 2111](https://github.com/openclaw/crabbox/pull/2111). Thanks @steipete.
 - Show Freestyle and Crownest settings through a shared offline JSON/text display path, preserving zero and false values while redacting URLs and reporting only loaded key presence. [PR 2104](https://github.com/openclaw/crabbox/pull/2104). Thanks @steipete.
 - Retain recognized workspace-owner protocol states in child and phase-witness inspection errors, improving diagnostics without changing ownership decisions or retry behavior. [PR 2106](https://github.com/openclaw/crabbox/pull/2106). Thanks @steipete.
--- a/docs/commands/stop.md
+++ b/docs/commands/stop.md
@@ -101,6 +101,11 @@ Crabbox lease ID and local slug:
   the caller's cancellation and deadline. The CLI remains signal-cancelable;
   the automatic-cleanup timeout does not shorten that lifetime. Non-cancelable
   callers, including detached job cleanup, retain the 30-second fallback.
+  For an ordinary claim whose sandbox is already missing, explicit
+  `--daytona-forget-missing` queries the exact bound ID and removes only local
+  state after a structured 404. Check the original account and endpoint first:
+  this does not verify remote deletion or stop billing. Live sandboxes and fixed
+  replay-protected claims cannot be forgotten.
 - `coder` — stops the Coder workspace by default and removes the local claim.
   Set `coder.deleteOnRelease` or pass `--coder-delete-on-release` to delete the
   workspace instead.
--- a/docs/providers/daytona.md
+++ b/docs/providers/daytona.md
@@ -177,6 +177,7 @@ Provider flags:
 --daytona-work-root
 --daytona-ssh-gateway-host
 --daytona-ssh-access-minutes
+--daytona-forget-missing # stop only; ordinary local claims only
 ```
 
 The non-auth settings can also be set through environment variables:
@@ -195,6 +196,26 @@ require operator resolution. See [managed Daytona cleanup](../features/lifecycle
 for lifetime, key-rotation, and recovery behavior.
 
 ## Direct lifecycle
+
+### Forgetting an ordinary stale claim
+
+If native TTL or an external deletion removes an ordinary sandbox, normal `stop`
+preserves its claim when the sandbox cannot be resolved. After checking the
+original Daytona account and API endpoint, explicitly forget that local record:
+
+```bash
+crabbox stop --provider daytona --daytona-forget-missing <lease-id-or-local-slug>
+```
+
+This command queries only the claim's exact sandbox ID and removes local state
+only after a structured Daytona 404. It refuses live sandboxes, authorization
+errors, malformed responses, checkpoint-held claims, and coordinator/adapter
+registrations that require normal release reconciliation. It never
+sends DELETE or reports remote release: a 404 can also mean the current credentials
+cannot access the original account, so any remaining sandbox may still be billed.
+The flag is command-only and cannot be enabled in a profile or environment variable.
+Fixed claims cannot be forgotten; use ordinary `stop` to reconcile them and retain
+their terminal replay protection as described below.
 
 ### Fixed operation IDs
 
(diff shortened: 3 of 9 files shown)


steipete:
Superseded by https://github.com/openclaw/crabbox/pull/2454 (merged as fcbc92e8b). The maintainer decision was to make stale-claim recovery a core capability behind the existing targeted `stop --force` rather than a Daytona-only flag: adapters only attest that the exact bound resource is absent (structured not-found for the exact ID plus, where the provider can list, a complete inventory that omits it), and core removes the unchanged local claim under its lock without any native mutation. Your PR shaped the Daytona evidence rules that landed; new ordinary Daytona claims now also capture their endpoint/organization binding so this recovery is scoped correctly. Changelog credits you and @mislavivanda. Thanks @Patrick-Erichsen!

PR #2119, opened 2026-09-11
feat: retain repository ownership during remote command execution

steipete:
## What Problem This Solves

Callers that manage their own remote workspace need command execution and cleanup to remain bound to Crabbox's current repository claim. Preparing raw SSH credentials and executing them later cannot preserve that ownership boundary after an operator transfers the lease.

## Why This Change Was Made

Add `crabbox exec --id … -- COMMAND ARG…` as a provider-owned command transport. It resolves fresh access under the existing shared claim fence and retains that fence through process exit, cancellation, and private transport cleanup. It preserves non-PTY stdin/stdout/stderr and command exit status, without syncing files or hydrating a workspace. Credentials remain in private SSH configuration.

Add `stop --current-repo --id …` for fixed-ID cleanup with the repository check inside the provider's exclusive release fence. Daytona implements this capability; ordinary administrative stop remains unchanged. An offline `exec --check` reports effective execution and scoped-cleanup capabilities so integrations can reject unsupported routes before allocation. Daytona provides a separate exec admission contract that rejects ordinary and incomplete claims before provider access; only acquired fixed leases have the required release fence.

## User Impact

Integrations can execute commands on an acquired fixed-ID Daytona lease while retaining repository ownership for the entire local operation, and safely refuse cleanup after an operator transfers that lease.

## Evidence

- Entry-point command, exact stream/argument/exit-code, cancellation, claim-lock, and scoped cleanup tests, including real loopback OpenSSH authentication failure without credential diagnostics.
- Race tests for CLI and Daytona; vet for the four changed packages; docs build and diff checks.
- Real Daytona proof passed in 140.515 seconds using the final source-built binary: fixed-ID acquisition and recovery, streamed workspace/file operations, retained execution, registry reopen, recreate, and cleanup. A transfer between two synthetic repositories after command preparation rejected old-owner execution, file access, and scoped stop; the current owner could execute and release the lease. All three test leases were released. An earlier equivalent Daytona transport run also verified access beyond its configured 60-second SSH token lifetime.

No credential configuration, schema, or dependency changes. Direct Daytona is the initial supported provider. Other providers must add the matching execution activity lifecycle before advertising support. No release is part of this PR.


steipete:
Fixed the ordinary-lease admission finding in c5a3b45256a53ad4677a31848f2138105599c3f1. Core now requires an exec-specific provider resolver, and Daytona admits only acquired fixed-ID claims before native access. Its ordinary run resolver keeps its existing contract. The entry-point regression reproduced the old acceptance, then passed for a fixed lease and rejected ordinary, malformed ordinary, and incomplete fixed claims with zero provider requests.

The exact source-built CLI also passed a real Daytona integration run in 140.515 seconds. The harness prepared execution, transferred the lease between two synthetic repositories before process admission, verified rejection of the old owner's command/file access/scoped stop, and verified successful execution and release by the current owner. Relevant captured output:

```text
Crabbox live: recovered pre-submission reservation cbx_6b2a131aa4ec
Crabbox live: concurrent creators share cbx_157787907cd3
Crabbox live: adopted cbx_157787907cd3; recreated cbx_243d58295807
Crabbox live: prepared exec, file access, and cleanup respect repository transfer; current owner succeeds
Crabbox live: all registered leases released
Test Files  1 passed (1)
Tests       1 passed (1)
```

The live run used the registered backend integration being completed in https://github.com/openclaw/openclaw/pull/144454. All leases above were released. The updated CLI candidate also passed fresh independent P2 review, focused race tests, and Windows test-package cross-compilation.

PR #2122, opened 2026-09-11
refactor(http): share provider JSON request construction

steipete:
## Summary

Share the complete context-bound JSON request-envelope constructor across DigitalOcean, Linode and Vultr. Remove the three duplicated encoding/request prefixes without introducing a common cloud client.

The helper preserves nil versus typed-nil bodies, Encoder's newline and HTML escaping, encoding-error precedence, context association, content length and replay via GetBody. Each provider keeps its existing URL concatenation, headers, credentials, transport, retries and response interpretation. Linode still omits Content-Type for a nil body; DigitalOcean and Vultr still set it. Vultr constructs and encodes again inside each retry attempt.

## Verification

- Pinned-original client implementations passed the new focused regressions through a Go test overlay, including existing loopback HTTP request-shape tests. The same candidate tests passed after fixing two missing shared imports caught by the first build.
- Broader client and helper regressions passed 49 tests and 30 subtests across four packages with Go 1.26.5 and the race detector, without failures or skips. Inert transports check exact request bytes and headers; existing local HTTP handlers check real request dispatch. The Vultr retry test verifies a fresh encoding on the second attempt without real delay or cloud traffic.
- Independent managed Codex review passed at P0 scope. Docs generation and whitespace checks passed.
- Reversing only the three prefix substitutions and necessary import changes restores the original client files byte-for-byte; all header, retry and response code remains unchanged.

This is a behavior-preserving internal refactor. It adds no new provider credentials, network destinations or public CLI behavior. Tests use synthetic inputs and local/inert transports, not live cloud accounts.

## Observed production-client HTTP trace

Source: `f66ca6080c79e3233cc2a4957e41e9472976020d` (clean before and after). Three additive, uncommitted test overlays exercised the production DigitalOcean, Linode, and Vultr request clients through real HTTP transport and loopback `httptest` servers. No production file was replaced.

Race-enabled targeted proof passed: 3 tests, 10 subcases, 11 actual HTTP requests; 13.73 seconds total. Authorization was present for every request; only presence was recorded. Caller-context association was verified in a wrapper that delegated to the real transport.

| Provider | Case | Method / path | Content-Type | Content-Length | Actual body (JSON string notation) | Server response |
|---|---|---|---|---:|---|---:|
| digitalocean | nil | `POST /proof/records?limit=2` | `application/json` | 0 | `""` | 204 |
| digitalocean | typed-nil | `POST /proof/records?limit=2` | `application/json` | 5 | `"null\n"` | 204 |
[...]

PR #2123, opened 2026-09-11
refactor(linode): select paginated resource types at call sites

steipete:
## Summary

Let Linode's five resource-list methods select their result types directly. Replace the private endpoint-to-type switch, union of unrelated slices and callback plumbing with one Linode-local typed page collector.

The collector preserves the anonymous raw metadata envelope, including Page and Results validation even though those values do not control traversal. Resource data is still decoded separately with the same error wrapper. Later failures return earlier pages but never partially decoded current-page data. Nil results, ordering, duplicates, query construction and the existing page-count stop rule remain unchanged.

The old unsupported-path branch belonged only to the private dispatcher; the five resource methods now select their concrete types and unchanged paths. No public fallback or compatibility shim is added. Authentication, request construction, response handling and other providers' pagination policies are untouched.

Builds on https://github.com/openclaw/crabbox/pull/2122. The landed dependency and unrelated upstream XFCE changes were integrated without modifying them; the resulting Linode patch is byte-identical to the reviewed and tested original change.

## Verification

- Original implementation and candidate passed the same frozen pagination tests: five resource types across 17 scenarios each, plus the preserved original loopback HTTP case.
- Broader Linode client regressions passed 7 tests and 95 subtests with Go 1.26.5 and the race detector, both before and after integrating current main. No failures or skips.
- Cases cover empty/null pages, nonpositive page counts, ordering/duplicates, partial results after HTTP/read/data/metadata errors, and malformed metadata retaining its validation boundary. Tests also assert exact request paths, context and headers.
- Independent managed Codex review passed at P0 scope for the original and integrated patch. Docs generation and whitespace checks passed.

Tests use synthetic data with local or inert HTTP transports, not live cloud credentials or resources. This is an internal behavior-preserving refactor, not a new user-visible feature.

## Observed production pagination over HTTP

Source: `e4415bad02d310a9a7e5138c278cc79ebb583c4b`; clean before and after. A single additive Go test overlay exercised all five production Linode list methods through real HTTP transport to loopback `httptest` servers. No production file was replaced.

Targeted `-race` proof passed: six cases, 12 HTTP requests, 8.72 seconds total. Every server-observed request was GET with `page_size=500`; authorization was present, with no value recorded.

| Production method / case | Server-observed path | Pages | Returned typed IDs | ID Go type | Result |
|---|---|---|---|---|---|
| ListLinodes | `/linode/instances` | 1 → 2 | `[101, 102, 102, 103]` | `int64` | success; order and duplicate retained |
| ListTypes | `/linode/types` | 1 → 2 | `["fixture-a", "fixture-b", "fixture-b", "fixture-c"]` | `string` | success; order and duplicate retained |
| ListImages | `/images` | 1 → 2 | `["fixture-a", "fixture-b", "fixture-b", "fixture-c"]` | `string` | success; order and duplicate retained |
| ListRegions | `/regions` | 1 → 2 | `["fixture-a", "fixture-b", "fixture-b", "fixture-c"]` | `string` | success; order and duplicate retained |
| ListFirewalls | `/networking/firewalls` | 1 → 2 | `[101, 102, 102, 103]` | `int64` | success; order and duplicate retained |
| ListLinodes-later-data-error | `/linode/instances` | 1 → 2 | `[71]` | `int64` | decode error; prior page retained |

The failure server sent page one with ID `71`, then page two containing ID `72` followed by a string where an integer ID was required. The actual returned slice was `[71]`: the earlier page remained, and the current partially decoded page was discarded. Actual error:

```text
linode /linode/instances?page=2&page_size=500 decode data: json: cannot unmarshal string into Go struct field linodeInstance.id of type int64
```

Limits: synthetic loopback responses only; no cloud account, live provider compatibility, credential validity, native tooling, or provisioning was exercised. The optional null-first-page case was not run. This is a runtime transport trace, not a mock-transport result.


steipete:
Merged and verified on actual main commit `9801716716753b7c9fbeef79234259a5d717fc88`.

The five Linode list methods now select a typed pagination loop instead of routing through a private union and endpoint callbacks. The two-stage envelope/data decoding remains intact: later failures preserve earlier pages, partial current-page decoding is discarded, ordering and duplicates stay unchanged, and only the existing page-count rule stops traversal.

Verification passed 7 focused tests and 95 subtests with the race detector. The actual merged source also passed six loopback HTTP cases across all five list methods, producing 12 real requests with observations identical to the pre-merge proof. This is local transport proof, not a live cloud lifecycle claim. The independent mandatory Codex review found no P0 findings within its selected scope.

Post-merge [CI](https://github.com/openclaw/crabbox/actions/runs/34660901379) passed all 11 jobs, [Connector E2E Smokes](https://github.com/openclaw/crabbox/actions/runs/34660901360) passed all five jobs, and [CodeQL](https://github.com/openclaw/crabbox/actions/runs/34660900546) passed all four jobs. No reruns were requested. The unrelated tag-codec work that reached main before this merge was preserved.

PR #2235, opened 2026-09-13
fix(freestyle): reject negative decoded VM sizing

steipete:
## Summary

Freestyle's mechanical flag bindings preserve signed integer inputs, but the provider had no later sizing validator. A negative override could silently omit custom sizing when neither field was positive, or reach the request with a positive companion field.

Use the existing provider validation hook and validate again when configuring the provider directly. Decoded negative vCPU or memory values now return exit 2 before constructing a backend. Keep zero/default and partial-default sizing unchanged, and continue passing positive values through for the service to validate.

Raw flag binding, generated YAML/environment decoding, request shape and workspace operations are unchanged. In particular, this does not make every raw negative YAML/environment value an error, add power-of-two validation, or migrate the legacy v1 API. Provider docs and the Unreleased changelog are updated. The workspace files being refactored in https://github.com/openclaw/crabbox/pull/2232 are untouched.

## Verification

The original provider fails all five new negative Configure cases; the four default/positive controls and existing raw-binding/request-shape controls pass. The candidate passes five Freestyle tests with nine subtests, plus two existing core decode tests with eight subtests, under the race detector with no skips. Provider vet, CLI build, docs link checks and site generation pass.

The built public CLI was also exercised in five negative-flag cases against an inert loopback endpoint using synthetic-only configuration. Every case returned the exact exit-2 diagnostic and the endpoint recorded no requests. The temporary fixture and server were cleaned up.

Independent managed review found no P0 issues in the four-file patch. No live Freestyle account or VM was used; this is local validation/CLI proof, not a claim about current server support for the legacy API or allowed positive sizes.

## Current-main integration

The automatic merge preview of final head `9d3d9dd9a42f664b686c6dc2fdf2be231dacecf6` onto main `7e018529d5784021ac9dbe209acd8d16c4d17884` is conflict-free and has tree `79e2df289172667b9d8f1606e587d71e2b5f226f`. It preserves the landed workspace refactor and passes the focused race/decode tests, vet, CLI build and full docs gate. A separate managed review of this combined four-file patch found no P0 issues.

The upstream Go changes produced a different CLI binary, so all five public-CLI negative-flag cases were rerun against this new artifact: exact exit-2 diagnostics and no requests recorded by the inert loopback endpoint. The earlier-base preview and binary proof remain historical evidence, not a substitute for this run. This is a retained uncommitted automatic merge preview, not a rewrite of the PR branch.


steipete:
Merged as `c6740e55bcd26bfe7b9d0465944907e5a3641674`, on top of the landed workspace refactor. The actual merge tree exactly matches the reviewed and tested current-main preview.

The merged source passed the focused Freestyle race tests (5 top-level tests, 9 subtests), existing decoding tests (2 top-level tests, 8 subtests), vet, CLI build and the full docs gate, with no test skips. Its metadata-neutral CLI binary is byte-identical to the refreshed preview binary exercised in five public-CLI negative-sizing cases; each returned the expected exit 2 and the inert endpoint recorded no requests. The earlier-base binary proof was not substituted for that refreshed run.

Decoded negative sizing now fails before backend construction. Raw decoding, zero/default and positive-value handling remain unchanged; this adds no positive-tier restriction or legacy API migration. Docs and the Unreleased changelog are included.

All four post-merge validation workflows passed on the first attempt: [CI (12/12 jobs)](https://github.com/openclaw/crabbox/actions/runs/34773315488), [Connector Lifecycles](https://github.com/openclaw/crabbox/actions/runs/34773315500), [CodeQL](https://github.com/openclaw/crabbox/actions/runs/34773315364), and [Pages](https://github.com/openclaw/crabbox/actions/runs/34773315504). Connector desktop-resize was skipped; no validation-job reruns were needed. Task fixtures were cleaned up. This is local validation/CLI proof, not a live Freestyle VM claim.

PR #2237, opened 2026-09-13
fix(opencomputer): reject negative decoded sizing

steipete:
## Summary

OpenComputer preserved decoded negative CPU and memory values, but request construction copied only positive values. Negative sizing therefore silently became service-default or unintended partial sizing.

Add the existing selected-provider validation hook and use it when configuring the provider directly. Decoded negative CPU or memory now returns exit 2 before backend construction. Preserve zero/default sizing, partial positive values, and service-owned positive tier validation.

Mechanical YAML/environment/flag decoding, aliases, timeout fallbacks, request construction and workspace/lifecycle code are unchanged. The docs distinguish preserved raw decoding from the new selected-provider check, and the Unreleased changelog is updated. This addresses the same lost-input pattern as https://github.com/openclaw/crabbox/pull/2235 without imposing Freestyle's decoding policy or adding shared tier assumptions.

## Verification

Original production fails exactly five new negative-sizing subcases; six existing/default/positive subcase controls pass. The candidate passes three provider tests with eleven subtests and two existing decoding tests with sixteen subtests under the race detector, with no skips. Vet, CLI build and full documentation checks also pass.

The built public CLI passed eleven negative-sizing cases covering explicit flags, environment values, repository YAML and both provider aliases. Each returned exit 2 with the expected diagnostic, and the inert loopback endpoint recorded no requests. Tests used synthetic-only configuration; task fixtures and the server were cleaned up.

Independent managed review found no P0 issues in the four-file patch. No live OpenComputer account or sandbox was used, and this does not certify service-side positive tiers or change authentication behavior.

## Integrated main

The PR now includes main `082b1aa81d098d3f15be34bd65c0d924d86b53b5` at head `0315d88c1dd73eac17de19d0c231a2c5ace48e91`, tree `e907d141cdefe96aea0b113aae5d10d56d03d33d`. Both integrations conflicted only at changelog insertions; all upstream files and both notes are preserved, and the OpenComputer guard, regression tests and provider docs are unchanged. No force push was used.

The combined source passes the focused provider/decoding races, vet, build, full docs and independent P0 review. The eleven public-CLI cases were repeated after the command refactors changed the binary. The final rebuild with the image update matches that tested binary exactly. Earlier source/CI receipts remain historical; CI for this integrated head is the landing gate.


steipete:
Merged as https://github.com/openclaw/crabbox/commit/42b9d9234e20bc564376b0e9e2709f059e1141ee. Negative decoded CPU and memory values now return a configuration error before backend construction instead of silently falling back to service defaults. Zero/default, positive and partial sizing behavior remain unchanged; the provider owns validation through the shared hook.

Verified the actual merge tree against the qualified candidate, then repeated the focused provider and decoding race tests (5 top-level tests, 27 subtests; no skips), vet, CLI build, and full documentation checks. The rebuilt binary exactly matches the one exercised by eleven local CLI flag/environment/YAML/alias cases, all returning the expected exit-2 diagnostics with zero requests to the inert endpoint. This is configuration-path proof, not a live cloud lifecycle claim.

All actual-merge workflows passed on their first attempt: [CI](https://github.com/openclaw/crabbox/actions/runs/34778421646) (12 jobs), [CodeQL](https://github.com/openclaw/crabbox/actions/runs/34778421473) (4), [Pages](https://github.com/openclaw/crabbox/actions/runs/34778421678) (2), and [Connector E2E Smokes](https://github.com/openclaw/crabbox/actions/runs/34778421655) (5 passed; desktop-resize skipped). Changelog and provider docs are included. All task-owned temporary proof resources are cleaned up.

PR #2242, opened 2026-09-13
fix(hyperv): reject negative decoded resource sizes

steipete:
## Summary

Reject decoded negative Hyper-V CPU and memory values during provider configuration validation. Previously, early defaults replaced those invalid values with 4 CPUs and 8192 MB before validation could see them. Only zero now selects those defaults; positive values remain unchanged.

The provider owns this policy through the existing shared validation hook. Direct Configure keeps its target, Windows-mode, and Tailscale checks before sizing validation. Signed environment/flag decoding, positive-only YAML overlays, target overrides, and unselected flag behavior are preserved. No guest, PowerShell, lease lifecycle, image, or authentication code changes.

## Verification

The original implementation fails the focused regression. Candidate race tests passed: 15 provider tests with 10 subtests and 5 decoding tests with 3 subtests, with no skips. Existing target override cases are included; negative errors assert the exact exit code 2 and diagnostic. Focused vet, CLI build, and the full documentation gate passed.

Seven isolated local CLI checks passed: five negative flag/environment cases produced the expected diagnostics, and two passive config-show controls accepted zero and positive input. Native-command stubs recorded zero calls. Exact zero/default and positive sizing values are covered by pure regression tests; this is not native Windows or VM lifecycle proof.

The first CLI fixture incorrectly expected a Hyper-V sizing section in config-show JSON; that section is not currently exposed. Corrected the fixture expectation without changing production code and retained the original failure evidence.

Validated on main f8e969616b3f7aa8dc8e75ab094f88ccbe410bee, with tracked source bytes, modes, and index bound to the gates. Managed review found no actionable blocking findings. Provider docs and changelog updated; no credential changes.


steipete:
I accepted the operation-scope finding after this PR merged: putting creation sizing checks in shared backend configuration could prevent existing-lease maintenance. That placement issue is now repaired by merged https://github.com/openclaw/crabbox/pull/2244 at `7a830c90b1e3168f0afba6304b91e96e43f8ecf3`, including the same pattern in Multipass, OpenComputer and Freestyle.

Creation still rejects negative sizing. Valid synthetic owned-lease regression cases now reach stop/cleanup/reuse under irrelevant negative creation settings, retaining their original ownership and state assertions. The paired real CLI commands and observed diagnostics are inspectable in the follow-up PR body; they establish admission and passive configuration behavior, not real VM destruction.

The original merge’s local and all four main workflows passed, but did not cover this operation-scope regression. The corrective merge now separately passes 34 top-level race tests and 119 subtests, vet, a fresh CLI build matching the 40-command proof binary, full docs, and all four actual-main workflows. No outstanding finding from this sizing-placement issue remains.

PR #2259, opened 2026-09-14
test(cloudflare): make cleanup deadline fixture deterministic

steipete:
## Summary

Repair the Cloudflare bounded-cleanup fixture exposed by the post-merge CI run at 3f93287cb8c260fe92eb657d5210102cd48fe03e. Its 20 ms deadline covered local claim admission, a real loopback connection and server-handler scheduling. The test could see DeadlineExceeded with a correctly retained recovery claim before its server-side arrival signal was observed, then fail with "destroy was not attempted".

Use the existing synthetic HTTP transport and Go's standard virtual-time test harness to observe the request and its real context cancellation deterministically. Keep the 20 ms deadline, require a positive bounded request deadline, and retain the error, failed result, kept session, command-count and recovery-claim assertions. No production implementation, timeout, provider call or cleanup policy changes.

## Verification

- The repaired test passes with the race detector.
- Focused cleanup/keep/finalization tests pass: three top-level tests and two subtests, with no skips.
- Provider vet and diff checks pass.
- Managed Codex review found no blocking findings in the selected P0 scope.

The failed job's original log is retained; no blind workflow rerun was used. This PR addresses the Cloudflare fixture only. The separately failed Windows architecture smoke is being investigated independently.

## Maintainer proof assessment

The automated review found no actionable code defect and verified repository-admin authorship. This is a test-only fixture repair, so executing the changed fixture and its neighboring real-loopback tests is direct proof of the change; a live Cloudflare sandbox would not exercise this test harness. No live Cloudflare service claim is made. Normal required GitHub checks still apply.

Captured local race-test output:
```text
--- PASS: TestCloudflareRunCleanupDestroyUsesBoundedContext (0.04s)
--- PASS: TestCloudflareRunKeepReturnsSessionHandle (0.03s)
--- PASS: TestCloudflareRunFinalizesAfterFailedCleanup/0 (0.03s)
--- PASS: TestCloudflareRunFinalizesAfterFailedCleanup/42 (0.03s)
--- PASS: TestCloudflareRunFinalizesAfterFailedCleanup (0.07s)
PASS
```

The combined Cloudflare-fixture and Windows-diagnostics tree also passed the selected five top-level tests/five subtests under race, both-package vet, and Windows amd64 compilation. Compilation is not native Windows proof.

PR #2262, opened 2026-09-14
test(opencomputer): make cleanup deadline fixture deterministic

steipete:
## Summary

Apply the same deterministic deadline-fixture approach as https://github.com/openclaw/crabbox/pull/2259 to OpenComputer's cleanup test. The existing fixture gives a real HTTP handler 20 ms to record a DELETE, so an early deadline can be mistaken for a missing attempt. This equivalent source pattern was found during the follow-up audit; no OpenComputer flake was reproduced or claimed.

Reuse the unchanged fake API response handler through the existing RoundTripFunc and Go's standard virtual-time test harness. Clone the incoming request before adapting its body and close the consumed body. Preserve the 20 ms timeout, actual cancellation cause, failed result, kept session, elapsed bound and exact DELETE count; additionally assert a positive bounded request deadline.

Only one test file changes. Production cleanup, request construction, identity, timeouts and API behavior are unchanged. Neighboring real-HTTP tests remain in place.

## Verification

- Three focused tests pass under the race detector: ephemeral create/execute/cleanup, bounded cleanup, and reused-session behavior.
- Provider vet and diff checks pass.
- Managed Codex review found no blocking findings in the selected P0 scope.

This is test-harness proof, not a live OpenComputer service smoke. The candidate is based on current main d59a1eb65536296f212cbd1dec4e8db40be390b6; the provider's baseline source is unchanged across the earlier base advance.


steipete:
Landed in b29070336fe86ced4baeb6098ab4cbfc26f02cd1. The OpenComputer cleanup fixture now uses virtual test time and its existing response handler through an in-process transport. It preserves the 20 ms deadline, cancellation cause, failed result, kept session and exact DELETE count. Production behavior is unchanged; this equivalent fragile fixture pattern was found by source audit, not by a reproduced OpenComputer failure.

The separate managed review and original tests completed before the later automated input-screening notice was observed. That automation produced no verdict. The rejected packet and workflow log were not opened, rerouted or retried, and the candidate head/body were not changed to seek another review. After main advanced, mechanical identity checks confirmed the unchanged candidate and provider source; the same three focused race tests and vet passed on the integrated tree. The normal guarded merge preserved the other maintainer's changes without administrative override.

Post-merge verification passed all 12 [CI jobs](https://github.com/openclaw/crabbox/actions/runs/34835293263), [Connector E2E](https://github.com/openclaw/crabbox/actions/runs/34835293265) (five passed, desktop-resize skipped), and all four [CodeQL analyses](https://github.com/openclaw/crabbox/actions/runs/34835292216) on the exact merge. Pages was excluded by the test-only path filter; peer checks were not counted. A transient stale cached status resolved through the original watcher without a workflow rerun or access-route override. No live OpenComputer service proof is claimed.

PR #2267, opened 2026-09-14
fix(sync): compare managed-state scope during snapshot acceptance

steipete:
## Summary

Snapshot acceptance compared ordinary exclusion rules but omitted their effective managed-state subtree. Identical ignore patterns therefore compared equal even when that protected subtree changed or disappeared.

Include the managed subtree in the shared equality helper instead of adding separate comparisons at each acceptance call site. Selection rules and cleanup behavior are unchanged. Documentation and the Unreleased changelog are updated.

## Verification

- The pure equality regression fails on the base for changed and removed managed scopes; the same-scope control passes.
- Production acceptance-loop proof uses synthetic Git checkouts and the existing phase hook to make one sequential `XDG_STATE_HOME` configuration change after the first copy. Payload files stay unchanged.
- On base `1559392055497f9da1539316dfc1d173db42662e`, the overlay behavior test fails with `attempts=[1] accepted scope="state-b/crabbox"`: it accepts without rebuilding. The fixed overlay passes with `snapshot attempts=[1 2]`, the new managed subtree, and identical payload content.
- The full local-Git snapshot path also rebuilds on attempt two. It passes on both base and candidate because its fingerprint already includes the managed subtree. This is a cross-platform control, not claimed as a failing-before regression.
- `go test -race ./internal/cli -run '^(TestManagedStateSnapshotRevalidatesConfigurationChange|TestGitOverlaySnapshotRevalidatesManagedConfigurationChange)$' -count=1 -v` passes on the candidate in 6.849s. The same tests on the baseline fail only the overlay case in 6.184s.
- Earlier focused equality/stable snapshot/staged-index races passed in 9.830s.
- The first platform run failed the Windows overlay fixture at `core_autocrlf`, before its behavioral assertion. The test now separates the POSIX overlay regression from the native-Windows-compatible local-Git control; no production guard is weakened or Git setting forced to claim unsupported overlay support.
- Independent managed Codex review: scoped-clean at P0 for implementation and genuine test changes.
- Exact-head [CI passed](https://github.com/openclaw/crabbox/actions/runs/34903055458) at `e0bfdcff07f97fa1c0809b4f6634d1ca32b02d2d`. Native Windows and macOS logs confirm the equality cases and local-Git control with `[1 2]`, the new scope, and preserved content. The separate overlay regression is qualified by the local before/after run above, not those native job logs.

No filesystem mutation race or secret transfer was used for the configuration-transition proof.


steipete:
Merged and verified in https://github.com/openclaw/crabbox/commit/ba467447de8343894acdf884f3888d8ba9d6e29e. The merged tree exactly matches the reviewed and tested candidate. The shared exclusion comparison now includes managed-state scope; ordinary selection and cleanup policy remain unchanged.

The overlay regression fails on the baseline and rebuilds on attempt two with the fix. Native Windows and macOS passed the equality cases and portable local-Git control. The initial Windows fixture failed at overlay configuration validation; coverage was corrected without weakening that guard. The separate overlay regression is supported by local before/after proof, not claimed as an execution in those native contract jobs.

Post-merge focused race tests passed in 4.993s. All post-merge workflows succeeded: [CI](https://github.com/openclaw/crabbox/actions/runs/34905035258), [CodeQL](https://github.com/openclaw/crabbox/actions/runs/34905035478), [Connector](https://github.com/openclaw/crabbox/actions/runs/34905035251), and [Pages](https://github.com/openclaw/crabbox/actions/runs/34905035296). Connector's desktop-resize job was skipped. Changelog and sync documentation are included, and the landing checkout is clean.

PR #2276, opened 2026-09-15
fix(crownest): share context-bound status waits

steipete:
## Summary

Use the shared status-wait owner for Crownest's request deadline, observation loop, and cancellation-aware delay. Crownest retains lease resolution, its 250 ms interval, readiness aliases, terminal-state decisions, and provider diagnostics.

The new context-only constructor shares initialization without introducing an adapter clock. Existing clock-aware providers keep their current deadline precedence, and Freestyle's separate in-flight request contract is unchanged.

This also fixes inconsistent cancellation classification: when the caller cancels after the child deadline has expired, Crownest's former sleep path returned `DeadlineExceeded`, while its API-error path returned `Canceled`. Both boundaries now preserve caller cancellation. Ready and terminal observations retain their existing precedence.

## Verification

- The new regression fails on the unchanged implementation only for parent cancellation after child expiry: `context deadline exceeded`, expected `context canceled`.
- Shared, Crownest, and OpenComputer race suites pass, including existing clock-precedence tests; targeted vet and docs build pass.
- A compiled CLI exercised warmup, one-shot status, polling to readiness, a terminal state, a blocked-request timeout, caller cancellation, stop, and claim removal against an owned loopback HTTP fixture. The 150 ms request timeout completed in about 183 ms with exit 5; cancellation remained a cancellation.
- Managed review reported no blocking findings in the changed scope.

The HTTP proof uses the real CLI and client, but a synthetic local API—not a live Crownest cloud account. The live doctor check found no configured Crownest API key. No credentials, provider configuration, or resources are changed by this PR.

### Final-commit runtime evidence

The clean checkout was verified before and after building at commit `21e25274af3216c9f6369ab93e3ace27316c924d`, tree `0368e43461a2a2eb794a71d4c5b68d2a97686ef6`; Go package resolution pointed at that checkout. Binary SHA-256: `1f9caf0ff0bcce27e408b5713c12fc46ec8db68886b75ce36ad15778dfb8926b`. Go emitted no embedded VCS fields, so this binding is based on those external readbacks, not an inferred embedded revision. The complete HTTP fixture was rerun with these captured results:

```json
{
  "single-status": {
    "state": "starting",
    "ready": false
  },
  "poll-ready": {
    "state": "running",
    "ready": true
  },
  "terminal": {
    "exit": 5,
    "statusGETs": 1,
    "stderr": "crownest sandbox sbx_wait_fixture entered terminal state \"failed\" before becoming ready"
  },
  "request-timeout": {
    "exit": 5,
    "statusGETs": 1,
    "stderr": "timed out waiting for cnsbx_sbx_wait_fixture to become ready"
  },
  "caller-cancel": {
    "exit": 1,
    "statusGETs": 1,
    "stderr": "context canceled"
  },
  "cleanup": {
    "deleted": true,
    "claimAbsent": true
  }
}
```

The fixture issued two status GETs before readiness and one bounded request for each timeout/cancellation case. This remains local API-fixture proof, not a Crownest cloud-account claim.


steipete:
Merged as https://github.com/openclaw/crabbox/commit/0a3632e5551efa54d906af64d4ab0733565572d9, with the same tree as the reviewed PR head. Crownest now shares request deadlines, polling, and cancellation classification while retaining its own readiness, terminal-state, and provider-error policy. Existing clock-aware adapters keep their prior deadline precedence.

The regression demonstrates the former inconsistency: after child expiry followed by caller cancellation, the polling-delay path returned `DeadlineExceeded` instead of `Canceled`. The new implementation passes that case. Shared, Crownest, and OpenComputer race suites pass on clean main, and the compiled-CLI loopback HTTP proof covers status polling, bounded requests, cancellation, and cleanup. Its actual results and build binding are in the PR body; no live Crownest cloud-account test is claimed.

All post-merge checks passed for the exact merge commit, including the full CI run https://github.com/openclaw/crabbox/actions/runs/34983510113, connector smokes, CodeQL, and Pages. Docs and the Unreleased changelog entry are included. No release was published by this change.

PR #2298, opened 2026-09-17
fix: report effective Linode type in doctor

steipete:
## Summary

Linode doctor reported the generic class default even when `linode.type` selected a different machine. Provisioning already resolved that configuration correctly; diagnostics read the raw defaulted field instead.

Use the existing effective-type resolver in Doctor so diagnostics and provisioning share the same precedence. No machine selection or cloud mutation behavior changes. Add a concise changelog entry and extend the existing fake-client test suite.

## Verification

- Regression overlay against the unchanged backend fails specifically for the native configured type (`g6-standard-1` reported instead of `g6-nanode-1`).
- Focused Go 1.26.5 tests pass for native configured type, explicit generic override, class fallback, existing defaults, and runtime bindings.
- Managed independent review: no actionable findings at P0 scope.
- `git diff --check` passes.

No live cloud operations or new credentials/configuration. This is a diagnostic-only fix with local behavioral proof.


steipete:
Merged as 0c9c08f46e0ddc7d0e6aa84fc77b844bfe9c43da. Doctor now uses the same effective machine-type selector as acquisition, so its diagnostic agrees with provider-native configuration, explicit generic overrides, and class defaults without changing provisioning.

The configured-native-type regression failed against the previous implementation and all three selection cases passed after the fix. This was controlled fake-client behavior proof, not a live Linode API run. Post-merge CI passed all 12 jobs: https://github.com/openclaw/crabbox/actions/runs/35176255330. All four CodeQL analyses passed: https://github.com/openclaw/crabbox/actions/runs/35176254872. Connector smokes passed with the existing desktop-resize skip: https://github.com/openclaw/crabbox/actions/runs/35176255472.

The changelog entry is a maintainer-authored landing update, as required by this repository. It is intentionally retained; the automated contributor-changelog objection does not apply to this maintainer change.

PR #1745, opened 2026-09-02
fix(parallels): macOS leases connect when Tools cannot report the guest

saariuslystoned:
## What Problem This Solves

Lets operators acquire and drive Apple-silicon macOS guests through the Parallels provider when Tools cannot report a guest IP or run guest-prep, then reuse the saved SSH port after Tools is cleared, and take an account-authenticated screenshot or type into an already-owned clone after a normal guest Remote Management grant.

Native acquire of an owned clone does not stamp `desktop=true`, so `03de7896` died in the capability gate before any RFB session. This PR keeps the earlier Parallels macOS bootstrap, MAC-matched DHCP discovery, saved-port reuse, password-boundary, and help-contract work, then adds a fail-closed owned-clone desktop exception and an RFB type-path settle so Screen Sharing works without forging `--desktop`.

## Why This Change Was Made

Parallels Tools remains the preferred discovery and preparation path. For macOS only, operators can configure a trusted host-side bootstrap identity. When Tools publishes no IP, Crabbox matches the clone's exact normalized NIC MAC against an unexpired Parallels DHCP lease, rejects missing, malformed, stale, or duplicate-ambiguous results, verifies the configured SSH port, and streams the existing preparation script over SSH. Later sync and run traffic uses Crabbox's generated per-lease SSH key. A reuse invocation restores the claim's saved SSH port before discovery so a generic config port (2222) that is closed does not block a live saved port (22).

The Parallels provider uses the same account-password contract as other macOS providers. The guest account password is accepted only from trusted user configuration or `CRABBOX_PARALLELS_PASSWORD`. There is no password CLI flag; repository configuration cannot select it; help text states that boundary. The password is removed from target-facing, host, bootstrap, direct `prlctl`, and doctor child environments, and is never written to the guest or passed on argv by Crabbox. Generated SSH lease credentials are unchanged.

[...]

PR #2301, opened 2026-09-17
fix(parallels): preserve IP discovery results

steipete:
## Problem and fix

Parallels Resolve discarded IP-discovery errors and replaced the observed VM with the failed call's empty result. A normal connection could therefore succeed in resolution without an endpoint; best-effort status and cleanup could lose the VM identity needed for guest metadata. Successful discovery had a second inconsistency: the SSH target used the discovered IP, but the returned Server retained the empty inventory address, so status could show an empty host beside a valid SSH host.

Return the original discovery error for connection requests. Keep status and release best-effort, retaining the original observed VM when discovery fails. Assign the reported server IP once, after discovery, so its value and the SSH endpoint agree.

This is a narrow repair of existing main behavior. It adds no flags, dependencies, credentials, bootstrap, DHCP-discovery, desktop or source-VM mutations. The larger feature at https://github.com/openclaw/crabbox/pull/1745 remains separate, including its untouched-template and independent desktop-proof requirements. Thanks @saariuslystoned for the related address-refresh correction in that PR.

## Verification

- Baseline main `de1e442e` failed all four controlled discovery-error cases: connection, release, status, and readiness status.
- Before the address refresh, the successful-discovery regression returned SSH host `192.0.2.10` with an empty Server IP; the new regression requires both fields to agree.
- Focused Resolve/Release/Cleanup race tests passed on Go 1.26.5 after integration with current main.
- Independent review of the complete candidate against current main found no actionable P0 defects; diff hygiene passed.

```sh
go test -race -count=1 -timeout=2m -run '^(TestResolve|TestRelease|TestCleanup)' ./internal/providers/parallels
```

These are controlled fake-runner regressions, not live VM proof. No VM was started, changed or deleted for this repair. The maintainer changelog records the user-visible correction.


steipete:
Landed as https://github.com/openclaw/crabbox/commit/74dd94d304fae007b27dd0bb7268718023b0ace6. Parallels connection resolution now reports discovery errors instead of replacing the observed VM with an empty value. Status and cleanup retain best-effort behavior and the original VM identity. The reported address is assigned after discovery, so displayed host and SSH endpoint stay consistent.

The controlled baseline failed all four discovery-error modes, and the separate success regression showed a valid SSH host beside an empty server address. The corrected focused race tests passed, independent scoped P0 review passed, and final-head CI, CodeQL, release checks and connector smokes passed: https://github.com/openclaw/crabbox/actions/runs/35184224410.

Main was pulled and verified clean with the exact tested Git tree. Post-merge CI passed all 12 jobs: https://github.com/openclaw/crabbox/actions/runs/35185702441. All four post-merge CodeQL analyses and connector smokes passed as well. This is fake-runner and source/CI evidence, not live VM proof; no VM was started, changed or deleted. Thanks @saariuslystoned for the related address-refresh correction in https://github.com/openclaw/crabbox/pull/1745. That larger bootstrap/desktop feature remains open with its separate untouched-template and native-proof requirements.

PR #2302, opened 2026-09-17
fix: account for bare-metal CPUs in AWS quota checks

vincentkoc:
## What Problem This Solves

AWS instance names do not reliably encode their vCPU count. Bare-metal types could be treated as costing zero vCPUs, allowing impossible quota recommendations and misleading capacity checks.

## User Impact

Provisioning, readiness, and the CLI capacity doctor use EC2 instance metadata for quota calculations. Missing or denied metadata remains unknown and cannot produce an affordable recommendation. Existing market fallback and private workspace limits remain intact.

## Why This Change Was Made

Reuse the existing DescribeInstanceTypes reader for candidate types and both market checks. Add the read-only ec2:DescribeInstanceTypes permission to the generated baseline IAM policy and the existing tightly scoped image-qualification requests.

## Evidence

- Reproduced both Worker quota/readiness failures and all six CLI metadata cases on the original implementation; the corrected implementation passes.
- 297 AWS Worker tests and 276 selected AWS fleet tests passed.
- 100 CLI AWS tests and 84 AWS provider tests passed, including their subcases.
- Worker and Node.js typechecks, formatting, lint, and scoped independent review passed. Exact-source results from unchanged sibling tests were retained.
- EC2 allocation and hardware comparisons remain unrun; these checks used isolated provider fixtures.

Punchcard-Session: clear-meadow-workshop-fv


steipete:
Maintainer input qualification for head `a45a0d4cc25cf0e9e97b8aed42de4a8086318885` against merge base `74dd94d304fae007b27dd0bb7268718023b0ace6`:

I checked all 14 changed head files and their 14 exact base blobs, validating head contents against Git blob IDs, and scanned those inputs plus the PR diff locally with TruffleHog 3.97.4, network access denied, verification disabled, and every result class included.

The six observed offline matches are three unchanged source patterns on both sides:

- `docs/infrastructure.md:116`: a PostgreSQL documentation example with a reserved documentation hostname and replacement password.
- `worker/src/aws.ts` (base line 3913, head line 3969): a Box-detector false positive on an exported TypeScript function identifier, not a credential literal.
- `worker/test/fleet.test.ts:47673`: a synthetic URI inside a literal error thrown by a mocked provider, using a reserved documentation hostname.

These exact occurrences are qualified as documentation/test fixtures or a source identifier. No detected values or raw scanner output are reproduced here. The diff produced no findings.

This is not a blanket file exemption, a passing secret scan, a code-review verdict, or permission to merge. The original verification-enabled diagnostic reports four findings but exposes metadata only for finding zero; this offline six-match result does not certify the complete historical finding set or qualify unseen material. The initial local intake also caught and corrected the Contents API's omitted content for the large fleet test file before the complete scan and blob validation above.

steipete:
Maintainer validation at `a45a0d4cc25cf0e9e97b8aed42de4a8086318885`:

- Seven CLI quota/doctor tests passed: `go test ./internal/cli -run 'TestAWS(CapacityDoctor|RecommendedClassForQuota)' -count=1 -timeout=5m`.
- All 297 Worker tests in `aws.test.ts`, `aws-private.test.ts`, `aws-qualification-authority.test.ts`, and `aws-fetch-client.test.ts` passed.
- These runs used an explicit credential-free environment and fresh HOME. Managed independent review completed at the configured P0-only threshold with no findings; that is not an all-priority correctness certificate.

The source remains unchanged. Real AWS catalog/quota and create/use/destroy validation are not yet established: the available direct AWS credential chain could not authenticate. No resources were allocated, and running against an already-deployed broker would not test this candidate's Worker implementation. The PR remains unmerged pending that live verification; passing fixture tests is not being substituted for it.

steipete:
Superseded by https://github.com/openclaw/crabbox/pull/2435, which landed this change rebased onto current `main` with your co-author credit, plus two follow-up guards found in review (non-Standard instance families no longer compared against Standard quotas; `im4gn`/`is4gen` covered). Verified by the full race suite on a remote Crabbox AWS gate and CI. Thanks @vincentkoc!

PR #1789, opened 2026-09-04
fix: report bounded ASCII Box cleanup progress

steipete:
## Summary

Fix https://github.com/openclaw/crabbox/issues/1730: show periodic progress while ASCII Box cleanup waits on native commands, and preserve the deletion phase, exact accepted operation, and last validated status when a status lookup times out.

The reporter uses a cleanup-scoped cadence, so both a stalled command and repeated fast polls produce output. Each command joins its reporter before returning, keeping it out of the guarded remote-cleanup callback's writer. Remaining time comes from the actual command context; no new whole-Stop deadline is claimed.

Centralize native status/config execution through the existing prepared-command path and cap capture at 8 MiB per stream. Runner/capture errors remain errors even if their output looks like valid empty inventory or a completed operation. Deletion authority, claim bindings, retry rules, native mutation sequencing, and completion/absence gates remain intact. Documentation and Unreleased notes credit @shunkakinoki for the report.

## Verification

- Red regression: a pending → blocked → stalled native deletion-status lookup returned only `context deadline exceeded` on previous source.
- Provider race suite passed after the fix, including held-command progress, fast-poll cadence, reporter join, partial-capture rejection, accepted-operation retention, and canceled inventory confirmation with both successful and failed responses.
- Existing real-subprocess output-limit, inherited-pipe, and controller-owned process-group tests passed under the race detector.
- Vet, docs-link checks, diff checks, independent semantic review, and managed review passed. The independent review caught two diagnostic gaps (snapshot phase labeling and cancellation precedence during failed inventory); both were corrected and retested.

Built live pending-retry proof and exact-head hosted CI are pending. The task's earlier ASCII Box is natively absent but its exact deletion operation is still `blocked`, so completed cleanup has not been proved. This draft does not claim successful destruction or permission to remove the retained claim.

## Source-bound native pending-retry result

Candidate `966a19453879c019d947ebd3af1daf18db64957a`, built CLI SHA-256 `d36ac7c8a587c0bda8877060043321e811df3f3609504639432bf6c9bf7ec150`, was run against the same task-owned retained claim using the official native ASCII CLI. It returned the expected exit 2 in 3.425 seconds:

```text
ascii-box cleanup phase=deletion-operation elapsed=0s remaining=30s; waiting for native CLI
ascii-box cleanup phase=deletion-operation operation=bdop_53e2ade7b3f34fdc83113cd1cbc6a4d0 last_observed_status=blocked; retaining claim
```

[...]

PR #2303, opened 2026-09-17
provider(ascii-box): support the Boat rebrand end to end

zozo123:
ASCII renamed Box to Boat and changed several CLI contracts at the same time. Current Crabbox cannot decode the new response envelopes, may select a missing SSH key or executable, rejects renamed deletion operations, and fails to redact the new API-key prefix. This update repairs those paths while preserving `ascii-box` as the canonical provider identity, so existing configuration, claims, and leases remain usable.

The adapter accepts `boat` as an alias, discovers the renamed executable when the legacy default is unavailable, reads both old and new JSON envelopes, recognizes both deletion kinds, and redacts both API-key prefixes. Mixed inventory envelopes are reconciled so an empty or partial envelope cannot hide a still-present resource. Existing explicit executable paths and existing legacy SSH keys retain precedence. Per-resource ownership, pagination, deletion completion, and unknown deletion-kind checks remain in force.

The provider documentation, generated provider metadata, and maintainer changelog are updated. Thanks @zozo123 for the implementation and native service verification.

### Verification

The contributor exercised the real Boat service with the renamed CLI and API key: doctor and inventory, fresh acquisition, repository sync and command execution on both advertised SSH endpoints and IP-only backends, retained-lease reuse, and deletion of every test sandbox with no sandboxes remaining. Retained reuse kept the same sandbox and credential. Each renamed surface also has focused regression coverage; contradictory mixed envelopes are tested hermetically because the service currently emits only the new envelope.

The maintainer candidate integrates current main at `7ff98c244932ae36b2b25478508f212ad9575221`. No product implementation was changed from the contributor's verified revision; maintainer edits add release notes and correct an obsolete test comment. Validation of the integrated candidate:

- `go test -race ./internal/providers/asciibox ./internal/providers/all ./internal/providers/shared -count=1 -timeout=3m`
- `go test ./internal/cli -run Ascii -count=1 -timeout=3m`
- `node scripts/generate-provider-matrix.mjs --check`
- `bash scripts/check-docs.sh` (command references, links, skill/integration consistency, generated site, and 20 site tests)
- Independent managed Codex review: no accepted/actionable findings at the configured P0 threshold. This is scoped review, not an all-priority correctness certificate.

The contributor's earlier validation also included `go vet ./...`. Hosted checks must pass on the updated candidate before merge.

### Boundaries

Recycled-IP host trust remains tracked separately in https://github.com/openclaw/crabbox/issues/1748 and https://github.com/openclaw/crabbox/pull/1785. A vendor deletion operation can outlast the release deadline; this update preserves the claim until completion is established. Cleanup progress remains separate in https://github.com/openclaw/crabbox/pull/1789. These proposals are distinct, not duplicates of the rename repair.

No new credential storage, token scopes, or deployments are introduced. Existing `asciiBox.*` configuration and `ASCII_BOX_API_KEY` remain valid. If both CLI-owned key files exist, the legacy key keeps precedence because the CLI exposes no authoritative way to choose another credential.


zozo123:
@clawsweeper re-review

Both findings from the last review are fixed at `2c03845c`, and the PR body now carries the requested behavior proof.

**P1 — select the SSH key used by the configured native CLI.** `boxSSHKey` no longer chooses by file name. It follows the key the configured CLI last wrote: `PrepareSSH` runs that CLI immediately before the target is used, so its key is the live credential, and a tie keeps `ascii_box_ed25519` (the pre-rename name). Covered in both directions by `TestBoxSSHKeyFollowsTheActiveCLI` — upgraded home selects the renamed key, pinned legacy CLI selects the legacy key — and proven live with both key files on disk, for a fresh acquisition and for reuse of a `--keep` lease by id.

**P2 / security — preserve every reported resource when both envelopes are present.** `decodeBoxes` no longer picks one envelope. It reconciles both by identity, preserving order and merging the two reports of the same Box. For `{"sandboxes":[{"id":"bx_other"}],"boxes":[{"id":"bx_target"}]}` it previously returned only `bx_other`; it now returns both, so contradictory inventory can no longer let stored deletion evidence authorize removing the claim of a Box the service still reports. Covered by `TestDecodeBoxesReconcilesBothEnvelopes` and `TestDecodeBoxesMergesFieldsAcrossEnvelopes`.

Both tests were verified to fail against the previous head `bf0804e6`.

One proof item I could not produce live, stated plainly: the contradictory-envelope case needs a CLI that emits both `sandboxes` and `boxes` with different contents, and the current service emits only `sandboxes`. That case is covered hermetically instead.

Also note the earlier readiness change was removed entirely rather than refined — measurement showed the endpoint is already present on the first `info` call (0.3s) and that `Acquire` re-fetches before building the target, so requiring it bought nothing and cost ip-only boxes about 60s. `boxReadyForSSH`, `waitForBoxReady` and `CreateBox` are byte-identical to pre-PR.

zozo123:
@clawsweeper re-review

Head is now `110ce4df`. The remaining P1 is addressed by removing the behavior rather than proving it.

**Backward compatibility of SSH key selection.** You were right that the timestamp-based choice was unprovable, and it was also unsound: a pinned legacy CLI whose valid key predates an unused renamed key would have been handed the wrong credential, and the CLI exposes no signal saying which key it authorized. So there is no longer any selection between two present keys. `boxSSHKey` falls forward to `ascii_sandbox_ed25519` **only when `ascii_box_ed25519` is absent**; whenever the legacy key exists it is presented exactly as before this PR. No existing setup can change which credential Crabbox offers, so the backward-compatibility case needs no native trace — the code path is unchanged from main. `TestBoxSSHKeyPrefersTheLegacyKeyWhenPresent` covers the case you named: a legacy key ninety days older than an unused renamed key still wins.

I have also retracted the earlier "both key files present selects the renamed key" trace from the PR body, because it described the heuristic that is now gone and would have contradicted the code.

Live re-verification on `110ce4df`: `ssh=user@213.32.26.240:19036`, sync 13.426s, `BOAT_E2E_OK`, `lease=6.232s`, exit 0.

One honest note on proof hygiene: an earlier attempt on this build failed with `SSH host-key verification failed`. That was https://github.com/openclaw/crabbox/issues/1748, not this change — sixteen verification acquisitions had left twelve pins in the provider-wide `known_hosts` and the service recycled an IP with a new host key. Clearing the stale pin let the run succeed. The PR body now records that reproduction.

steipete:
Merged as `f7c01047bf711a6af2d96ac5399046fa12f783e0`. The final reviewed head was `7d0c8c7ff6f643c5067b2af7ae2eb6322d976199`.

Local verification passed: Go 1.26.5 race tests for `./internal/providers/asciibox ./internal/providers/all ./internal/providers/shared`, focused CLI ASCII tests, provider-matrix consistency, and `bash scripts/check-docs.sh` including the generated-site tests. The contributor's real Boat acquisition/sync/run/reuse and zero-sandbox cleanup evidence is retained in the body; this maintainer pass did not repeat that live service run. Managed independent review found no accepted/actionable findings at the configured P0 threshold.

Final hosted CI https://github.com/openclaw/crabbox/actions/runs/35485039600 passed. All 22 checks on the exact final head were terminal: 20 successful and two optional skips, with no pending or failing checks. An unrelated main-branch Tart fix required a changelog-only integration refresh and a fresh CI run.

Existing lease host-trust and asynchronous-deletion/progress work remain separate; this merge does not close those issues. Contributor credit and release notes are preserved.

PR #2304, opened 2026-09-17
test(macos): isolate Linux installer temporary fixtures

steipete:
## Summary

Keep the Linux toolchain installer test fixtures portable on macOS without changing the production installer. Darwin's bare `mktemp -d` chooses its native temporary directory, while the fixture's Go stub requires the cache to remain under its private `TMPDIR`. That mismatch makes an otherwise valid fixture fail its version check.

On Darwin only, the fixture now supplies an explicit private-directory template for the bare `mktemp -d` invocation. Every other argument shape delegates to the real utility unchanged. Linux execution, the production installer, and all existing assertions are unchanged.

## Verification

The representative failure was reproduced on clean main, and the test and installer bytes matched the failing checkout. After the fixture correction, `node --test scripts/linux-toolchain-archives.test.js` passed 100 tests with zero failures and five existing optional skips on macOS. `git diff --check` passed. Independent structured review found no actionable issue at its requested P0 scope.

This is a test-only portability fix. It does not alter installed toolchains, credentials, provider resources, or release behavior; no user-facing changelog entry is needed.


steipete:
Landed as [09df29cc](https://github.com/openclaw/crabbox/commit/09df29ccc90bef484257756b2a322cb698761918). Darwin's bare temporary-directory command placed fixture caches outside the expected private test root; the Darwin-only adapter now supplies that root explicitly and delegates every other argument shape unchanged. The production installer and all existing assertions are unchanged.

Verification: the representative failure reproduced on clean main, then `node --test scripts/linux-toolchain-archives.test.js` passed 100 tests with zero failures and five existing optional skips on macOS. `git diff --check` and independent scoped review passed. All required PR checks passed on exact head `11418d9725a6054eed31410477f194a4bc44e710`, including [CI](https://github.com/openclaw/crabbox/actions/runs/35216445669) and the [snapshot release check](https://github.com/openclaw/crabbox/actions/runs/35216445600).

After the squash merge, the merged tree `527c708d22c597b5a537022dab5954cef4b773a9` exactly matched the tested candidate. The main checkout was fast-forwarded and verified clean. Post-merge main CI is a separate run, not claimed complete here. This is test-only maintenance; no release, production deployment, or provider resources were involved.

PR #1556, opened 2026-08-27
feat: share native filesystem runner and explicit copy recovery

steipete:
## Landed

Merged as `1171182f926d157534cc79e87e04ab7d4a9781f1`, with all pre-merge checks green and the exact tested combined tree verified on clean main. The [maintainer disposition](https://github.com/openclaw/crabbox/pull/1556#issuecomment-5715773532) explicitly accepts the remaining live changed-binding coverage gap; it is not claimed as tested. Post-merge CI passed all 12 jobs, and connector smokes passed five jobs with the desktop-resize job skipped.

## Summary

Replace SSH archive-copy and JUnit shell/PowerShell implementations with one Go filesystem runner over the existing authenticated transport. The CLI keeps provider routing, credentials, lease ownership, result policy, and local destination selection. The runner owns bounded reads, report discovery, framing, staging, and publication; the old remote algorithms are removed.

Copy and recovery now share physical-parent resolution. In particular, a symlink parent followed by `..` is resolved before lexical cleanup, retaining the caller-selected destination. Publication requires complete framing, the expected companion identity, and successful transport completion. Caller cancellation causes and owned-process retirement evidence remain distinct.

Runtime artifact discovery is shared with the existing supervisor. Complete installations carry Linux, macOS, and Windows filesystem companions for amd64 and arm64, with capability-aware manifests and versioned provenance. Explicit invalid operator packs never silently fall back to compilation; CLI-only development builds may compile the embedded helper locally. Native Windows bootstrap uses tar and a streaming .NET digest.

Native Windows archive copy remains unsupported. WSL2 and Windows operator copy routes retain their existing safe-rsync requirement; this does not add archive fallback there. Workspace-sync publication, process supervision, and provider-native/API-session adoption remain separate integrations.

## Recovery and upgrade contract

The implementation follows the [maintainer recovery decision](https://github.com/openclaw/crabbox/pull/1556#issuecomment-5709162058). New transactions use an external destination-bound durable journal. Legacy state requires an explicit `cp --recover keep-destination|restore-backup` choice, retaining unselected data rather than treating old marker contents as authorization. Normal copy refuses unresolved legacy state with exact next-action guidance.

[...]

PR #2305, opened 2026-09-17
test(macos): preserve native bootstrap in dash fixture

steipete:
## Summary

Keep the optional dash signal fixture faithful to the production launcher on macOS. Its blanket shell replacements also rewrote Darwin's native descriptor-cleanup bootstrap to dash; that synthetic combination fails when a busy test process passes a multi-digit descriptor.

The test now runs the outer shell and workload under dash while leaving the generated bootstrap unchanged. Eight owned null descriptors make the condition deterministic. All existing signal, exact-output, and cleanup assertions remain, and no production code changes.

## Verification

The original fixture on clean main reproduced all six failures with owned null descriptors. After the correction, `go test -race -count=3 -timeout=2m ./internal/cli -run '^TestWorkspaceOwnerDashSignalHandoff$'` passed all 18 subcase executions on native macOS with Go 1.26.5. `git diff --check` and independent scoped P0 review passed.

This preexisting fixture issue was found during the broader validation of https://github.com/openclaw/crabbox/pull/1556. It is kept separate from that runtime change. No live provider, production deployment, release, or user-facing changelog change is involved.


steipete:
Merged as `d3edc2ad464370b5c7768086b5206ba7aea9a781`. The merge tree exactly matches the tested candidate, and the maintainer main checkout is fast-forwarded and clean.

The fixture previously rewrote the native Darwin descriptor-cleanup shell to dash, creating an artificial failure when the child inherited a multi-digit descriptor. The correction retains the production bootstrap while testing the outer shell and payload under dash. Eight owned null descriptors reproduce the original six failures; the corrected six-case matrix passed three repetitions with race detection, retaining the signal, exact-output, and owner-release assertions.

The required independent P0-scoped review and all pre-merge CI checks passed. Production code is unchanged. Post-merge main CI is being verified separately.

PR #2341, opened 2026-09-18
fix(tenki): authenticate gateway certificates with the current CLI

francoluxor:
Tenki gateways present different host certificates signed by the same CA. Crabbox pinned a replica's leaf key, so subsequent connections could fail when another replica answered. The released Tenki CLI v1.7.0 also omits `known_hosts_file` and rejects the old `--idle-timeout` creation flag.

This change verifies gateway host certificates using authoritative provider trust. A CLI-reported `known_hosts_file` takes priority. Otherwise, the Tenki adapter uses the native CLI's workspace key and API endpoint to request the CA and discover the session's gateway identity, then writes a private authority file alongside the native certificate. It binds the exact WebSocket session route to the authenticated gateway ID used in the host certificate. Native private keys and certificates remain owned by Tenki.

The shared, provider-neutral SSH mode requires strict verification, excludes alternate trust sources and leaf-key enrollment, preserves certificate negotiation with a gateway alias, and disables multiplexing so each new connection checks current trust. Discovery failure never falls back to stale trust or TOFU. Existing established connections are not retroactively revoked.

Kept sessions now use only `--sticky`; non-kept sessions use `--max-duration` from TTL. The removed idle-timeout flag is unsupported by the current CLI. Docs explain that idle metadata is bookkeeping, sticky sessions have no duration limit, and the native maximum duration pauses a session rather than destroying it.

### Compatibility and dependencies

The authenticated fallback needs a workspace API key from `tenki onboard`, `TENKI_AUTH_TOKEN`, `TENKI_API_KEY`, or native CLI config. It follows native config/credential/endpoint precedence, requires HTTPS, rejects cross-origin redirects, and suppresses server error bodies that could echo credentials. No new Crabbox config fields or repository secrets are required.

Adds `connectrpc.com/connect` v1.20.0 for HTTP/2 gRPC using the existing HTTP-client boundary, plus a small MIT-licensed protobuf subset from the public Tenki Go SDK v1.2.0. `scripts/gen-tenki-proto.sh` regenerates the messages. All provider-specific discovery and credential behavior stays in the Tenki adapter.

### Verification

Fresh live qualification used official Tenki CLI v1.7.0 and a newly provisioned test workspace:

- Candidate doctor, non-sticky warmup, readiness, repository sync, remote command, no-sync reuse, and SSH command generation passed. The provider API key was not forwarded into the remote command environment.
- Candidate `run` rejected wrong, empty, and missing CA trust. A task-local CLI proxy wrapper changed only the candidate's authority file immediately before the SSH handshake; session identity, provider discovery, native credentials, and gateway route remained real. A subsequent valid candidate run confirmed the negative command's remote marker was absent. Failed handshakes did not enroll keys or repair the altered file.
- Strict native OpenSSH executed a command through both the WebSocket gateway and its authenticated public TCP endpoint using the candidate-generated CA file. Those routes presented two distinct host certificates accepted by the same authority. This is distinct-certificate proof across routes, not a claim of forced live replica rotation through the WebSocket route.
- Candidate stop succeeded, the exact session reached `TERMINATED`, provider active inventory was empty, and local claims were removed. Every sandbox created during qualification was cleaned up.

Automated coverage includes real OpenSSH trusted leaf rotation and wrong/removed/missing CA rejection through argument and generated-config transports, with and without gateway aliases. Provider regressions cover native credential precedence, authenticated gRPC wire fields, exact gateway binding, CA validation, redirects, cancellation, redacted errors, private atomic authority replacement, and unchanged native key/certificate files.

Passed locally with Go 1.26.5 and Node 24: full Tenki provider race suite, the broader core SSH race suite, all 20 native certificate scenarios and real forwarding tests without skips, provider/core vet, all six Tenki smoke-harness tests, and the documentation gate. Managed Codex review completed with no accepted/actionable findings at its configured P0 threshold. All substantive current-head CI checks passed: https://github.com/openclaw/crabbox/actions/runs/35498513771. Merged as `cdd3a00b0a1ca1b6180f9fc1fc6b05357bc11ef9`.

Thanks @francoluxor for the original fix and real-service rotation report. Includes the maintainer changelog entry for https://github.com/openclaw/crabbox/pull/2341.


steipete:
The repaired candidate `5c240e2d026d88bd7fe1af4c18fca3e650148999` has passed hosted CI: https://github.com/openclaw/crabbox/actions/runs/35484275219. All 27 exact-head checks are terminal (23 successful, four optional/dispatch skips). The ordinary fork workflow approvals were granted after maintainer source, native OpenSSH, and managed review.

Native certificate regressions and the Tenki provider race suite also passed explicitly with Go 1.26.5. The six previously failing negative SSH cases now reject before remote authentication; both trusted certificate cases still work on both transports.

The PR remains draft only for its documented real Tenki lifecycle qualification. This is a pushed repair with green CI, not a claim that local synthetic SSH proof substitutes for the provider gate.

PR #2357, opened 2026-09-19
fix(gcp): bound readiness and finalize lease SSH artifacts

steipete:
## Summary

Move GCP public-IP readiness into the provider adapter and reuse the shared polling owner. The two-minute budget now covers running observations as well as delays; an already-cancelled caller does not start another request. Immediate API failures, zero results on failure, the timeout diagnostic, and the later SSH-readiness policy remain intact.

The real-cloud qualification also exposed an existing cleanup omission: GCP deleted the instance and local claim but left the generated lease SSH key and `known_hosts`. Terminal cleanup now shares one guarded sequence: successful provider deletion (or confirmed absence), successful SSH connection-artifact cleanup, then claim removal. The unchanged-claim lock covers that sequence; provider or artifact failures retain the claim for retry. Dry-runs remain read-only. Stale records without a cloud resource identity preserve their existing claim-only pruning behavior. Azure's Delete/Resume cleanup decisions are unchanged.

## Why

The old client-side wait checked elapsed time only between requests, so a running observation could exceed the readiness budget. Moving the readiness decision behind the adapter also removes its duplicate polling loop and the obsolete internal client method/interface entry.

Claim removal did not imply SSH-artifact removal. Reusing an explicit shared finalizer closes that gap without deleting local recovery material before remote cleanup succeeds, or allowing a changed claim to authorize cleanup.

## Verification

- Regression tests reproduced the old release and confirmed-absent cleanup leaving SSH files behind. The updated cases pass, with additional coverage for dry-runs, provider/artifact failures, changed claims, absent release, and stale-claim cleanup.
- `go test -mod=readonly -race -count=1 -timeout=180s -json ./internal/providers/gcp ./internal/providers/shared ./internal/providers/azure`
- `go test -mod=readonly -race -count=1 -timeout=180s -json ./internal/cli -run '^TestGCP'`
- `go vet -mod=readonly ./internal/providers/gcp ./internal/providers/shared ./internal/providers/azure ./internal/cli`
- `go build -mod=readonly -trimpath -buildvcs=false -o crabbox ./cmd/crabbox`
- Independent Codex P0/P1 review: no actionable findings.

### Native GCP before/after

The first real one-shot run at `503c1cb3242308920626056eafcc5448afd92d2a` successfully created, bootstrapped, executed on, and deleted an `e2-micro` VM. Independent inventory confirmed VM/disk absence, but local inspection found the generated private/public SSH key and host-trust files still present. Those task-only files were recorded and manually removed after cloud absence was confirmed; this was not counted as automatic local-cleanup success.

The fixed build at `0d53db8866e3d537bf6bc7e06ced89c10a64d3eb` then passed a second direct, non-kept run in `us-central1-a`, with a 20 GiB boot disk, no repository sync, no guest service account, and no capacity fallback. The runtime command was:

```sh
crabbox run --provider gcp --type e2-micro --market on-demand --no-sync \
  --ttl 20m --idle-timeout 5m --slug gcp-n578-10f19223f2 --timing-json -- \
  sh -c 'echo gcp-native-readiness-ok; uname -m; sleep 8'
```

Captured guest output:

```text
gcp-native-readiness-ok
x86_64
```

The complete fixed run exited 0 in 190.022 seconds. Separate `gcloud` inventory captured the VM and attached boot-disk identities, verified the platform's 1,200-second `maxRunDuration` with `DELETE`, and confirmed zero test VMs or disks afterward. The test-only firewall was removed by the harness after VM absence; all four pre-existing firewall identities were unchanged. The generated lease SSH directory, SSH control directory, and exact claim were automatically absent after the run. No rescue stop, forced termination, or manual key cleanup was needed for the fixed build. The temporary cloud credential file was removed and the credential session closed.

Native source tree: `5773b846ea82e7c584223143e0496f7c71c35259`.
Built CLI SHA-256: `ae2545a5dcef4e91f1f8e8d377d819044ed8738683fa374e19190f28529a72c9`.

The follow-up head `39b96fdde0d58d7d3f5510bb9ff8631d283cf788` adds only changelog references and documentation wording. Rebuilding that candidate produces the identical CLI SHA-256 above; runtime source is unchanged from the native-qualified build.

This cloud run proves normal authenticated acquisition, readiness, execution and release. Stalled-observation deadline/cancellation behavior is covered separately by deterministic tests and the existing real local TLS/Google-SDK cancellation test; no vendor outage or authorization bypass was simulated.

## Configuration and release notes

No new configuration, credentials, dependencies, or provider permissions are introduced by the patch. The provider documentation and Unreleased changelog describe both user-visible fixes. Normal CI is required before landing.


steipete:
Merged as https://github.com/openclaw/crabbox/commit/73f24086b8abb05e3d288f217af7d9e52f6d9300 after the reviewed-head CI passed.

The GCP adapter now uses the shared polling owner for its two-minute public-IP budget, including in-flight observations. Terminal cleanup uses one unchanged-claim guard around provider deletion or confirmed absence, local SSH-artifact removal, and claim retirement. Failed provider or local-artifact cleanup preserves the retry claim; dry-runs remain read-only.

The real GCP before/after run documented above reproduced retained SSH files, then verified automatic cleanup of the fixed lease's keys, host-trust files, control directory and claim, alongside independent VM/disk absence. The merged commit has the exact reviewed tree. Focused merged-source race tests, vet and build passed, and its CLI SHA-256 is byte-identical to the native-qualified artifact: `ae2545a5dcef4e91f1f8e8d377d819044ed8738683fa374e19190f28529a72c9`.

Post-merge CI passed: https://github.com/openclaw/crabbox/actions/runs/35476608086. Changelog and provider documentation are on main; the task-owned main checkout is clean.

PR #2369, opened 2026-09-20
refactor(worker): share implicit machine candidate policy

steipete:
## Summary

Give GCP and Hetzner one owner for implicit machine-candidate resolution: unsupported canonical classes remain unsupported without a concrete stored type, custom classes remain literal, and the stored type keeps precedence with stable deduplication. Each adapter still selects its own profile table and supported target/architecture. Their deliberately different explicit-override guards stay local: GCP preserves a truthy whitespace value, while Hetzner requires nonblank text. AWS and Azure retain their additional provider-specific rules unchanged.

The existing config suite now covers the previously uncovered empty/whitespace explicit-override distinction. No machine mapping, allocation fallback, provider request, persisted format or public configuration contract changes.

## Verification and test-budget correction

Node 24 validation passed formatting, lint, Worker and Node typechecks, the complete default Worker suite, the Worker dry-run build and the Node build. The successful suite reports **3,258 passed, zero failed, 15 skipped**: three existing live opt-in checks and twelve Linux-only cases on this macOS host. The separate PostgreSQL recovery suite remains outside the default suite and is covered by its dedicated CI job. Independent Codex P0/P1 review of the full candidate found no actionable findings.

Initial full validation exposed five-second timeout sensitivity in the 64-group AWS signing stress matrix, followed by a second fixture mismatch. A fresh focused run also timed out; an unchanged single case subsequently completed every assertion in 13.84 seconds under a bounded diagnostic allowance. The patch gives only that existing matrix a 30-second per-case budget. All eight rows and every assertion remain; the global timeout and AWS production code are unchanged. In the passing full run, these cases took approximately 4.1–7.7 seconds each.

The timeout adjustment does not claim to fix unfinished-work/global-fetch teardown isolation. The final full validation ran in a credential-free environment behind an independently verified outbound-network block, with loopback and Unix sockets permitted. No native cloud lifecycle is claimed. The initial failed runs are retained as evidence rather than described as passing.

```sh
npm ci --prefix worker --ignore-scripts --no-audit --no-fund
npm run format:check --prefix worker
npm run lint --prefix worker
npm run check --prefix worker
npm run check:node --prefix worker
npm test --prefix worker -- --maxWorkers=4
npm run build --prefix worker
npm run build:node --prefix worker
```

[...]

PR #2370, opened 2026-09-20
test(worker): isolate AWS reconciliation stress transports

steipete:
## Summary

Isolate the AWS reconciliation stress matrix from mutable global fetch state. A timed-out case can leave asynchronous signing/reconciliation unfinished; restoring global fetch or installing the next case's fixture must not change where that older work dispatches requests.

The existing eight cases and 64 groups move to a dedicated test file, retaining every rule-set assertion and the scoped 30-second budget. The file-scoped SDK mock keeps real inherited signing but dispatches only through the originating asynchronous context. It rejects missing or closed contexts before signing and checks closure again after signing. There is no global or real-fetch fallback.

Teardown closes contexts synchronously. This does not cancel WebCrypto already running or drain fixture callbacks that have already started; those callbacks retain only their original fixture. The matrix exercises reconciliation, signing and synthetic successful responses, not SDK response-retry behavior. Other test files keep their existing transport paths. No production API or transport implementation changes.

## Verification

A gated-signing regression starts a request in one context, closes that context, activates another fixture and resumes signing. The old request rejects without either fixture receiving it. The same regression also checks already-closed and missing contexts. All eight stress rows passed alongside this regression.

Final Node 24 validation passed formatting, lint, both typechecks, the complete default Worker suite, the Worker dry-run build and the Node build: **3,259 tests passed, zero failed, 15 existing environment-gated skips** (three live opt-in checks and twelve Linux-only cases on this macOS host). All execution used a credential-free environment with independent outbound denial; loopback and Unix sockets remained available. No live cloud lifecycle is claimed. The initial lint findings—typed mocks and async assertion placement—were corrected before rerunning the full gate. Independent Codex P0/P1 review of the final candidate found no actionable findings.

```sh
npm run lint --prefix worker
npm run check --prefix worker
npm run build --prefix worker
```

The new file is needed to scope the SDK mock independently from the large fleet suite. It reuses the existing provisioning-storage fixture and uses neutral lease data. No dependency, credential, production behavior or release changes; no user-visible changelog entry is needed. The existing Worker-path deploy-on-main workflow remains active, including for test-only changes.

## Current-main integration

The dependency https://github.com/openclaw/crabbox/pull/2369 has landed. This PR targets main and includes `508c100aaaf7f4a9875ccb1096eb53654c6a5aeb`. The one squash-history conflict was resolved only after verifying that main's complete Worker subtree matched the reviewed parent. The final Worker subtree and main-to-candidate patch are identical to the passing candidate. Current-main independent review with provider, SDK, storage and test-configuration context found no actionable P0/P1 findings.

The Worker and Node executable bundles and source maps are byte-identical to the parent build; the generated README differs. This is not a claim that every build artifact is identical. Normal current-head CI and protected main-only landing remain required.


steipete:
Additional actual-runner validation on the exact candidate: a private append-only diagnostic deliberately timed out during gated signing. Vitest executed the candidate’s unchanged afterEach, and the next diagnostic verified that the old context was closed before releasing signing. The abandoned request rejected without reaching either fixture; a new-context signed request succeeded.

This was an expected-negative child run: exit 1, exactly one intentional 1,000 ms timeout, one passing follow-up diagnostic, and nine original cases filtered. No unexpected errors were observed. The unchanged original test-file prefix was hash-bound to the candidate. Node 24 execution used no credentials and independent external-network denial. The intentionally failing diagnostics are not part of this PR. This supplements the already-passing full default suite and current-head CI; it is not a live AWS lifecycle claim.

steipete:
Post-merge verification completed for `f8e1799e4f6287f25c27ef2979d5f67906ba92c0`: the exact parent-relative patch and complete Worker subtree match the reviewed candidate. An unrelated Tenki change landed during the merge window and was retained; this is not a claim that the entire candidate and merge trees are identical.

The merged checkout passed format, lint, both typechecks, 3,259 default-suite tests (zero failures, 15 documented environment skips) and both builds under credential-free external-network denial. Main CI and all substantive commit checks passed, including the separate PostgreSQL integration lane. The private actual-timeout diagnostic is described above and remains expected-negative evidence, not a green suite.

The existing automatic coordinator deployment succeeded, and authenticated application reachability was verified using the existing configured CLI destination and credentials. No login, access-policy change, manual deployment or native AWS lifecycle was performed. This closes the test-transport isolation work.

PR #2373, opened 2026-09-20
refactor(node): share asynchronous operation tracking

steipete:
## Summary

Move the existing Node HTTP operation tracker into a dedicated Node-local module and reuse it for socket and maintenance promises. Remove duplicate promise-set registration, settlement cleanup and repeated draining. Each work category retains its own tracker; this does not combine their lifecycle policies.

The new track method returns the same already-started promise. The existing run method keeps synchronous callback invocation and its original rejection/settlement scheduling. Socket FIFO, admission and shutdown order remain unchanged: sockets, alarms, provisioning, maintenance, bounded wake hints, jobs, then storage. Tracking neither cancels work nor closes admission.

## Verification

Added regressions for promise identity and rejection, synchronous callback throws, follow-up work registered during draining, and maintenance blocking jobs/storage shutdown after socket work completes. Formatting, lint, both typechecks, the full default Worker suite, Worker dry-run build and Node build passed on current main: **3,262 passed, zero failed, 15 existing skips** (three live opt-ins and twelve Linux-only cases on macOS). Tests ran on Node 24 with a credential-free environment and independent external-network denial. Independent Codex P0/P1 review found no actionable findings.

The first check caught two missing explicit returns in new test callbacks; those were fixed. A later full run hit the unchanged bootstrap readiness test’s five-second aggregate timeout. That test passed unchanged in isolation, and the subsequent complete gate passed. Its independent scenario-budget issue is being addressed separately rather than changing runtime behavior or raising the suite timeout.

```sh
npm run lint --prefix worker
npm run check --prefix worker
npm run build --prefix worker
```

No dependencies, credentials, public API or user-visible behavior change; no changelog entry is needed. This is local runtime bookkeeping proof, not a native provider lifecycle or PostgreSQL integration claim; the separate CI integration lane remains required. The existing Worker-path automatic deployment workflow remains active, although the production runtime change is Node-only.


steipete:
Post-merge verification completed for `314460f3fd312775fdfe3b8cc0034a9ce96c0222`. The actual parent and tree match the reviewed preflight composite; the introduced patch is unchanged.

The merged checkout passed formatting, lint, both typechecks, 3,262 default-suite tests (zero failures, 15 documented environment skips) and both builds under credential-free external-network denial. Main CI and all substantive commit checks passed, including PostgreSQL integration and the executed Linux fixtures. The cancelled check is a ClawSweeper metadata dispatch, not a source validation failure.

The automatic coordinator deployment completed its actual Deploy step, followed by successful authenticated reachability using the existing configured CLI destination and credentials. No manual deployment or native provider lifecycle is claimed. Separate HTTP/socket/maintenance ownership, FIFO and shutdown order are preserved; this closes the tracking consolidation.

PR #2376, opened 2026-09-20
refactor(gcp): share image observation normalization

steipete:
## Summary

Centralize GCP machine-image and disk-snapshot observation normalization inside the GCP adapter. Both API responses share the same optional identity, state and checkpoint-provenance fields; replace the duplicate interfaces and field mapping with one provider-local owner.

Callers still select the exact image kind and resource reference. Only snapshots add the snapshots list. Preserve nullish name/self-link defaults, empty strings, missing-state readiness, truthy immutable identifiers/provenance and split-token reconstruction. Decode the checkpoint token once. API routing, not-found handling, deletion and other providers are unchanged.

## Verification

Extend the existing machine-image and snapshot route tests with populated, missing, null, empty and partial-token observations. Complete output assertions cover optional-field omission and the snapshot-only list while preserving exact request-path assertions.

Node 24 validation passed formatting, lint, both typechecks, the full default suite and both Worker/Node builds: **3,267 passed, zero failed, 15 existing skips** (three live opt-ins and twelve Linux-only cases on macOS). All 98 GCP cases passed. Execution used a credential-free environment with independent external-network denial. Independent Codex P0/P1 review found no actionable findings.

```sh
npm run lint --prefix worker
npm run check --prefix worker
npm run build --prefix worker
```

This is a high-confidence behavior-preserving response-mapping extraction, not native GCP lifecycle evidence. No credentials, dependencies, configuration or user-visible behavior change; no changelog entry is needed. The existing Worker-path automatic deployment remains active. Normal current-head CI and protected landing remain required.


steipete:
Post-merge verification completed for `5922c4cec7e9af36d3301c76b5d39ea5cb0e7590`. The actual parent and tree match the reviewed preflight composite, preserving the exact introduced patch.

The merged checkout passed formatting, lint, both typechecks, 3,288 default-suite tests (zero failures, 15 documented environment skips) and both builds under credential-free external-network denial. Main CI and all substantive commit checks passed. The response tests preserve both image-kind routes, nullish/empty defaults, optional identity and checkpoint provenance, and the snapshot-only resource list.

The automatic coordinator deployment completed its actual Deploy step, followed by successful authenticated reachability using the existing configured CLI destination and credentials. This establishes application reachability, not native GCP image or provisioning proof. No manual deployment or access-policy change was made. This closes the provider-local normalization consolidation.

PR #2387, opened 2026-09-20
fix(parallels): install the macOS Node baseline on the Parallels path

### LINKED ISSUE #2381: Parallels macOS leases hang at readiness because nothing installs the `node`/`npm` baseline the readiness probe requires
A Parallels macOS lease whose template meets every documented requirement still never
becomes ready. Warmup clones, starts, discovers the guest IP, installs the per-lease key,
runs guest preparation, and then loops until the bootstrap deadline:

```text
waiting for 10.0.0.0:2222 bootstrap ssh-transport... elapsed=3m27s remaining=16m33s ports=proxy
...
timed out waiting for SSH on 10.0.0.0 during bootstrap probe=readiness cause=deadline_exceeded authentication=unknown
```

The guest is reachable the whole time. Logging in with the key Crabbox just installed works:

```sh
ssh -i <per-lease key> -o UserKnownHostsFile=<per-lease known_hosts> \
    -o ProxyCommand="ssh -W %h:%p <parallels host>" <guest user>@<guest ip> 'id -un'
# -> <guest user>
```

The actual cause is the readiness command, not the transport.

### Why

`sshReadyCommand` (`internal/cli/ssh.go`) adds a Node requirement for **every** macOS
target, on top of the generic `crabbox-ready` check:

```go
if target.TargetOS == targetMacOS {
    return sshReadyCommand(SSHTarget{}) + " && " + remotePortableShellInvocation("node --version >/dev/null && npm --version >/dev/null", nil)
}
```

In the guest:

```text
$ test -x /usr/local/bin/crabbox-ready && echo READY_SCRIPT_OK
READY_SCRIPT_OK
$ node --version
zsh:1: command not found: node
$ npm --version
zsh:1: command not found: npm
```

So `crabbox-ready` passes and the lease still fails readiness.

Crabbox has a macOS Node baseline installer — `sharedMacOSNodeInstall()`
(`internal/cli/bootstrap_generated.go`), checksum-pinned Node 24.19.0, idempotent, exits 0
when `node` and `npm` already work. It is only reachable from the AWS path:

- `bootstrapManagedMacOS` (`internal/cli/macos_bootstrap.go`) — called only from
  `internal/cli/aws_windows_bootstrap.go:44`
- `macOSUserData` (`internal/cli/bootstrap.go`) — reached only through `awsUserData`,
  `internal/cli/aws.go:636`

The Parallels macOS path instead runs `parallelsPOSIXEnsureReadyScript`
(`internal/cli/parallels.go`), which creates `/var/cache/crabbox/npm` and
`/var/cache/crabbox/pnpm` but never installs Node.

Eight providers declare a macOS target (`anthropicsandboxruntime`, `aws`, `lume`, `cua`,
`external`, `ssh`, `parallels`, `tart`). The readiness gate applies to all of them; the
Node install is AWS-only. I qualified this on Parallels; I have not tested the others.

### Also undocumented

`docs/providers/aws.md` states the contract for AWS macOS:

> Bootstrap installs Node 24.19.0 … Readiness requires both commands to execute
> successfully.

`docs/providers/parallels.md` "Template requirements" lists Parallels Tools, a stable guest
SSH user, an OpenSSH server, `git`/`rsync`/shell, and a power-off snapshot — but not Node.
An operator following that list builds a template that cannot reach ready.

### Suggested fix

Any one of these would remove the trap; the first two are the substantive ones:

1. Run the existing `sharedMacOSNodeInstall()` on the Parallels macOS preparation path, the
   way the AWS path already does. It is idempotent, so templates that ship Node are
   unaffected.
2. Make the readiness failure name the missing command instead of surfacing as an SSH
   deadline. Right now the message points at the transport and authentication, both of
   which are fine, which sends operators to debug networking.
3. At minimum, add `node` and `npm` to the macOS template requirements in
   `docs/providers/parallels.md`.

### Reproduced on

Current `main` (`033bcb47`), Parallels Desktop on Apple silicon, macOS 26.5 guest, remote
Parallels host over SSH. The same template completes clone → start → IP → Parallels Tools
guest exec → key install → guest preparation, and is SSH-reachable throughout.

saariuslystoned:
## Problem

macOS readiness requires Node, but only the AWS path ever installed it.

`sshReadyCommand()` asks a macOS target for **both** halves:

```go
if target.TargetOS == targetMacOS {
    return sshReadyCommand(SSHTarget{}) + " && " + remotePortableShellInvocation("node --version >/dev/null && npm --version >/dev/null", nil)
}
```

Node reaches macOS guests through `sharedMacOSNodeInstall()`, which is wired only into `macOSUserData` → `awsUserData`. The Parallels path runs `parallelsPOSIXEnsureReadyScript()`, which creates `/var/cache/crabbox/{npm,pnpm}` but never installs a runtime. So `crabbox-ready` passed, the Node half failed, and warmup sat at the gate until it timed out with a transport-shaped message that sends you looking in the wrong place:

```
timed out waiting for SSH on <guest ip> during bootstrap probe=readiness cause=deadline_exceeded authentication=unknown
```

The transport was fine throughout — the guest was SSH-reachable with the per-lease key. Fixes https://github.com/openclaw/crabbox/issues/2381.

## What this does

Settle Node before the readiness gate, resolving it the way the probe does:

1. **Already on the standard PATH** → nothing to do.
2. **The guest user's login environment has it** (Homebrew, nvm, asdf) → link into `/usr/local/bin` and download nothing.
3. **Neither** → install via the existing `sharedMacOSNodeInstall()`, checksum-pinned, same layout as AWS.

Then **verify rather than assume**: preservation is only claimed once the linked result actually runs on the fixed PATH `crabbox-ready` uses. Anything still needing the guest user's login context falls through to the installer instead of leaving a helper that fails while the probe passes.

`crabbox-ready` for macOS now asserts the same contract the probe enforces, on an explicit PATH.

Linux guests are untouched. No generated file is edited — `bootstrap_generated.go` is `DO NOT EDIT`; this only calls into it.

## Decisions a reviewer will reasonably question

**Why the Node work runs before the readiness gate.** The gate exits early when the guest's existing `crabbox-ready` passes. A guest carrying a `crabbox-ready` that predates the Node checks would pass it, exit early, and skip the install on *every* warmup — the fix would silently no-op on exactly the hosts that need it.

**Why `su - "$user" -c 'bash -lc ...'` rather than the user's login shell.** `remotePortableShellInvocation` execs `bash -lc`, and zsh reads different rc files, so `su - user -c 'command -v node'` can miss a runtime the probe resolves. Dropping to the user first also means their login setup is never sourced as root.

**Why `node -p process.execPath` rather than `command -v node`.** An asdf-style shim re-execs through its manager, which is absent from the PATH `crabbox-ready` uses. Linking the shim satisfies the probe and then fails the helper. `execPath` is the binary the shim ultimately runs, with npm and npx beside it.

**Why `export PATH` in `crabbox-ready`.** sshd hands non-login commands `/usr/bin:/bin:/usr/sbin:/sbin`, which excludes `/usr/local/bin`, and `npm` is `#!/usr/bin/env node` — without it both checks fail *even with Node installed*. The AWS `crabbox-ready` already exports the same PATH.

**Why `</dev/null` on the guest children.** `BootstrapMacOSOverSSH` feeds this script to `sudo -n /bin/sh -s` on stdin, so anything reading stdin swallows the rest of it while the shell still exits 0 — silent truncation:

```
$ printf 'echo LINE1\n/bin/cat >/dev/null\necho LINE3\n' | /bin/sh -s
LINE1
```

## Contributor-reported verification

The native evidence below was supplied before maintainer integration; it is not an independently rerun native trace for the revised head. The later mixed-prefix and shim changes are supported by executable regression fixtures and successful contributor-head CI.

**On a real Apple-silicon Parallels macOS 26.5.2 guest.** Runtime relocated to `/opt/homebrew/bin`, exported from the guest user's `~/.bash_profile`, nodejs.org blocked via `/etc/hosts`:

- the probe passes before anything runs — the template is genuinely "previously working"
- the first commit fails preparation: `curl: (7) Failed to connect to nodejs.org port 443`
- this branch exits 0, links `/usr/local/bin/{node,npm,npx}` to the preserved runtime, downloads nothing, and the exact `sshReadyCommand` string passes over SSH; `crabbox-ready` passes as root too
- with every runtime removed and the network restored, the pinned installer runs and verifies its checksum

Warmup against a template carrying the baseline reaches readiness in ~30s instead of stalling:

```
provisioned lease=cbx_72ebbbdc12ff vm=f516ef5f-1bea-456f-a202-a2c992a22c93 ip=10.211.55.48
ready ssh=parallels-01@10.211.55.48:22 network=public
warmup complete total=33.39s
```

**In tree.** The macOS stanza is exercised under a real `/bin/sh` with stubbed tools, covering: healthy runtime (no download, preparation continues), non-macOS guest (untouched), no runtime with an unreachable installer (preparation stops), both mixed-prefix arrangements offline, and an asdf-style shim whose manager is absent from the fixed PATH. Each regression test fails on its parent commit. Plus `sh -n` over the generated script for both desktop modes.

```
go test -race -timeout=20m ./internal/cli -run 'TestParallels|TestManagedMacOSNodeBaseline|TestMacOSNode'   ok
go test -race -timeout=20m ./internal/providers/parallels/...                                               ok
go vet ./internal/cli ./internal/providers/parallels/...                                                    clean
node scripts/check-docs-links.mjs                                                                           251 files, internal links ok
```

Note: the full `./internal/cli` suite exceeds a 25m `-timeout` on my machine, but so does stock `main` — 1501.292s on `origin/main` versus 1500.952s on this branch. Pre-existing and unrelated.

## Scope question for the maintainer

Three review rounds each surfaced a narrower case of the same question: how much of the version-manager landscape a provider fix should absorb. The verification step is meant to bound that — it makes the stanza fail safe for managers nobody enumerated, rather than needing a fix per manager.

If you would rather have a smaller surface — **don't preserve user-managed runtimes at all, always install the pinned baseline** — that is a strictly smaller change and I am happy to cut it back. The cost is that templates working offline today would begin requiring `nodejs.org`.

## Commits

| | |
|---|---|
| `50ad7613` | install the macOS Node baseline on the Parallels path |
| `97893f94` | preserve a guest's existing Node runtime |
| `03624656` | keep stdin off guest children |
| `7901d386` `f24d9321` | executable regressions + `docs/providers/parallels.md` (from the parallel session that closed https://github.com/openclaw/crabbox/pull/2388) |
| `7c8aaf6e` | cover the preservation arm |
| `1bfc53c3` | guard each command's destination independently |
| `5c23b949` | resolve runtime-manager shims to real executables |

## Maintainer follow-up

Accepted the preservation-plus-pinned-fallback contract: existing usable runtimes should not acquire a new network dependency merely to remain ready. Integrated current main and added the credited Unreleased changelog entry; no additional production rewrite was needed after the contributor resolved the shim case. Independent isolated Codex review of the complete integrated candidate through P2 was clean.

The prior exact contributor head passed CI. The revised integration head is `e6b2d953f75ad3ecabb41dfed09ae47c6e6dec96`; its full CI passed: https://github.com/openclaw/crabbox/actions/runs/35528883829. All 23 head checks passed or were skipped by their normal conditions. The current-main composition preserves the exact reviewed production/test/docs changes and both published changelog entries. Contributor code was not executed on the maintainer machine. Existing native installation/offline-preservation evidence remains attributed to the contributor, not to a new maintainer hardware run.


saariuslystoned:
## P1 addressed in `97893f94` — the finding was correct

The unconditional installer did introduce a compatibility regression, and I was able to reproduce it on a real Parallels macOS guest before fixing it.

`sharedMacOSNodeInstall()` opens with:

```sh
export PATH="/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin"
if node --version >/dev/null 2>&1 && npm --version >/dev/null 2>&1; then exit 0; fi
```

That PATH excludes `/opt/homebrew/bin`, so on Apple silicon a Homebrew or user-managed runtime is invisible to the installer while remaining visible to the probe, which execs `bash -lc`. The installer therefore downloaded over a healthy runtime, and with nodejs.org unreachable it turned an offline-capable template into one that needs the network to stay ready.

### Fix

Resolve the runtime the way the probe does, before reaching for the installer:

1. standard PATH already resolves `node` and `npm` → nothing to do
2. the guest user's login environment has them → link into `/usr/local/bin` so root, `crabbox-ready` and the probe all agree, and skip the download
3. neither → fall back to the pinned shared installer

The lookup is `su - "$user" -c 'bash -lc "command -v node"'`, which drops privileges **before** sourcing anything, so user-controlled login setup is never executed as root.

One detail worth calling out, because it bit me while testing: this uses `bash -lc` rather than the user's default login shell on purpose. `remotePortableShellInvocation` execs `bash -lc`, and zsh reads different rc files — so `su - user -c 'command -v node'` can miss a runtime the probe resolves. My first attempt at this fix had exactly that bug; a guest with the runtime exported from `~/.zprofile` looked healthy to zsh and broken to the probe. There's now a test pinning the distinction.

Self-linking is guarded, so a broken `/usr/local/bin` install falls through to a clean reinstall rather than pointing at itself.

### Verification — the upgrade scenario the review asked for

Real Parallels macOS guest (Apple silicon), runtime relocated to `/opt/homebrew/bin`, exported from the guest user's `~/.bash_profile`, nodejs.org blocked via `/etc/hosts`:

**Baseline — the template is genuinely "previously working":**
```
$ ssh <guest> 'bash -lc "node --version && npm --version"'
v24.19.0
11.17.0
$ ssh <guest> 'env PATH=/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin sh -c "node --version"'
sh: node: command not found      # invisible to the installer
```

**Previous commit (`50ad7613`) — reproduces the regression:**
```
curl: (7) Failed to connect to nodejs.org port 443 after 1 ms: Couldn't connect to server
v1 exit: 1
```

**This commit (`97893f94`):**
```
v3 exit: 0
/usr/local/bin/node -> /opt/homebrew/bin/node
/usr/local/bin/npm  -> /opt/homebrew/bin/npm
/usr/local/bin/npx  -> /opt/homebrew/bin/npx
pinned installer did NOT run: no /usr/local/lib/crabbox/node-v* prefix
nodejs.org: BLOCKED

READINESS PROBE: PASS (exit 0)     # exact sshReadyCommand string over SSH
crabbox-ready OK as root           # consistent for both callers
```

**Missing-runtime case still works** — every runtime removed, network restored:
```
node-v24.19.0-darwin-arm64.tar.gz: OK
v24.19.0 / 11.17.0
v3 exit: 0
/usr/local/bin/node -> /usr/local/lib/crabbox/node-v24.19.0-darwin-arm64/bin/node
```

### Coverage

Two tests added alongside the existing three: preservation of a user-managed runtime (including that detection mirrors `bash -lc` and never runs the user's login setup as root, and that self-linking is guarded), and that the pinned installer remains the fallback arm rather than running unconditionally. The old-`crabbox-ready` case is still covered by the ordering assertion that the Node handling precedes the readiness gate.

saariuslystoned:
@clawsweeper re-review

Reviewed head was `50ad7613`; current head is `7c8aaf6e`. The P1 is addressed and there are three further commits since.

- `97893f94` — the P1. Resolve the runtime the way the probe does before reaching for the installer: standard PATH → the guest user's login environment, linked into `/usr/local/bin` → the pinned installer as fallback. The lookup is `su - "$user" -c 'bash -lc ...'`, so it drops privileges before sourcing anything and matches `remotePortableShellInvocation`'s `bash -lc` rather than the user's default shell, which reads different rc files. Self-linking is guarded.
- `03624656` — stdin guards. The script is delivered to `sudo -n /bin/sh -s` on stdin, so any child reading stdin silently swallows the rest of it while the shell still exits 0. `</dev/null` on the three lookups and the installer call.
- `7901d386` / `f24d9321` — executable regressions plus `docs/providers/parallels.md`, from the parallel session that closed #2388.
- `7c8aaf6e` — covers the preservation arm, which the stanza test above it did not reach because it stubs `su` to fail.

The upgrade scenario the review named as absent is now exercised twice. On a real Apple-silicon Parallels macOS guest with the runtime at `/opt/homebrew/bin`, exported from `~/.bash_profile`, and nodejs.org blocked via `/etc/hosts`: the probe passes before anything runs, `50ad7613` fails preparation with `curl: (7) Failed to connect to nodejs.org port 443`, and `97893f94` exits 0, links `/usr/local/bin/{node,npm,npx}` to the preserved runtime, downloads nothing, and the exact `sshReadyCommand` string passes over SSH. With every runtime removed and the network restored, the pinned installer still runs and verifies its checksum. The same three arms are covered in-tree under a real `/bin/sh` with stubbed tools.

One note on the review's "evidence reviewed" section: it reports the full `./internal/cli` suite as a concern. It exceeds a 25m `-timeout` on this machine, but so does stock `main` — 1501.292s on `origin/main` at `033bcb47` versus 1500.952s on this branch. Pre-existing and unrelated to this change.

saariuslystoned:
@clawsweeper re-review

Second P1 fixed in `1bfc53c3`. The finding was correct.

`node` and `npm` can live in different prefixes, and the combined guard rejected preservation whenever *either* already occupied its destination — so a template with node at `/usr/local/bin` and npm only in the guest user's login PATH reached the installer and downloaded over a runtime that already satisfied readiness. `npx` already had independent handling; node and npm were the coupled pair.

Each destination is now guarded on its own and only what needs linking is linked. The combined case is kept deliberately and made explicit: if *neither* needs linking, both already sit at `/usr/local/bin` and the standard-PATH probe above still failed, so that install is the broken one and the installer should replace it. That is now an `||` rather than a side effect of the `&&`.

Both mixed-path arrangements are covered with downloads unavailable, as requested. They fail on the parent commit `7c8aaf6e`:

```
--- FAIL: .../node_already_at_its_destination,_npm_elsewhere
    preservation arm failed: exit status 1: REACHED-nodejs.org
--- FAIL: .../npm_already_at_its_destination,_node_elsewhere
    preservation arm failed: exit status 1: REACHED-nodejs.org
```

and pass on `1bfc53c3`. `REACHED-nodejs.org` comes from a stubbed `curl`, so any download attempt fails the test outright. Each case also asserts that a command already at its destination is left as a regular file rather than relinked onto itself, and the runtimes sit in a directory deliberately kept off PATH, standing in for `/opt/homebrew/bin` or `~/.nvm`.

Acceptance commands from the review, on `1bfc53c3`:

```
go test -race -timeout=20m ./internal/cli -run 'TestParallels|TestManagedMacOSNodeBaseline|TestMacOSNode'   ok  13.221s
go test -race -timeout=20m ./internal/providers/parallels/...                                               ok   2.532s
go vet ./internal/cli ./internal/providers/parallels/...                                                    clean
node scripts/check-docs-links.mjs                                                                           251 files, internal links ok
```

saariuslystoned:
@clawsweeper re-review

Third P1 fixed in `5c23b949`. The finding was correct, and this closes the class rather than the instance.

Two changes:

**Resolve the real executable, not the shim.** `node -p process.execPath` returns the binary an asdf/nvm shim ultimately runs, with npm and npx beside it, so what gets linked into `/usr/local/bin` works without the manager on PATH. Falls back to `command -v` when execPath is unavailable.

**Verify rather than assume.** Preservation is now only claimed once the linked result actually runs on the exact PATH `crabbox-ready` uses. Anything still needing the guest user's login context falls through to the pinned installer instead of leaving a helper that disagrees with the probe. That is the general guard: it also covers managers neither of us has thought of.

Also switched `dirname` to `${var%/*}` so the lookup does not depend on a command being present on a minimal PATH.

Regression coverage, offline as requested: `node`/`npm`/`npx` shims that exit 127 unless their manager is on PATH, with the real binaries elsewhere and the manager reachable only from the simulated login shell. It asserts the link resolves to the underlying binary, that no download is attempted, and — the decisive one — that the preserved runtime runs on `crabbox-ready`'s fixed PATH with the manager absent. It fails on the parent `1bfc53c3`:

```
node linked to ".../shims/node", want the underlying binary ".../installs/node"
```

Acceptance on `5c23b949`:

```
go test -race -timeout=20m ./internal/cli -run 'TestParallels|TestManagedMacOSNodeBaseline|TestMacOSNode'   ok  11.986s
go test -race -timeout=20m ./internal/providers/parallels/...                                               ok
go vet ./internal/cli ./internal/providers/parallels/...                                                    clean
node scripts/check-docs-links.mjs                                                                           251 files, internal links ok
```

`docs/providers/parallels.md` now states that shims are resolved to their underlying executables and that an unusable preserved runtime falls back to the installer.

Note for a human reviewer: this is the third P1 in a chain, each a narrower case of the same compatibility question — how much of the version-manager landscape a provider fix should absorb. The verification step above is meant to bound that, since it makes the stanza fail safe for any manager rather than requiring a fix per manager. If the preferred answer is narrower — for example not preserving user-managed runtimes at all and always installing the pinned baseline — that is a smaller change and I am happy to make it instead.

steipete:
Landed in https://github.com/openclaw/crabbox/commit/5bfbe499d1c80ed05ee03ad18660357d80838f40 through the normal protected path. The merged tree exactly matched the reviewed current-main composition; the contributor implementation and credit are preserved.

Parallels macOS preparation now supplies the Node/npm baseline before the legacy readiness shortcut. It preserves usable existing runtimes offline, resolves version-manager shims to their underlying executables, checks both mixed-prefix arrangements, and verifies the fixed-PATH result before falling back to the existing checksum-pinned installer. Linux preparation remains unchanged.

[Exact-head CI](https://github.com/openclaw/crabbox/actions/runs/35528883829) and [post-merge main CI](https://github.com/openclaw/crabbox/actions/runs/35530548345) passed. All substantive source checks passed; one cancelled issue-comment dispatch was classified separately. Post-merge focused race tests passed (95 CLI plus 72 provider tests/subtests, zero skips), followed by vet and a full CLI build. Independent isolated review was clean. Native missing-runtime and offline-preservation evidence remains [contributor-reported](https://github.com/openclaw/crabbox/pull/2387#issuecomment-5750972846), supplemented by the exact-head executable mixed-prefix/shim regressions; this does not claim a new maintainer guest run.

The credited Unreleased entry is present, and https://github.com/openclaw/crabbox/issues/2381 is closed.

PR #2383, opened 2026-09-20
feat(parallels): add replay-safe fixed idempotent lease IDs

saariuslystoned:
## Summary

Implements `IdempotentLeaseIDBackend` for the Parallels provider, so
`crabbox warmup --provider parallels --lease-id cbx_<12 hex>` is accepted and an external
orchestrator can replay a dispatch safely. Before this, the gate at
`internal/cli/run.go:91-93` exited 2 with
`provider=parallels does not support fixed idempotent lease IDs`, and no configuration
could work around it.

Fixes https://github.com/openclaw/crabbox/issues/2382.

> **Revision 2** answers the ClawSweeper review in full — six findings, all of them the same
> defect: identity was bound to mutable names rather than to attested resources. The design
> section below is the revised contract; the revision notes and new native proof are at the
> end.

## Design

The idempotency key is the VM name, not a pre-allocated resource ID. `ParallelsClient.Clone`
already renders a host-unique `crabbox-<lease-id>-<slug>` name, and `prlctl` refuses a
second VM with that name on a host — so a concurrent duplicate create of the same lease is
rejected by Parallels itself, even after a lost reply. That makes this the Incus shape
(`internal/providers/incus/fixed.go`, daemon-unique name) rather than the Proxmox shape
(`newid` VMID pre-allocation discussed in
https://github.com/openclaw/crabbox/issues/1847).

A VM name is host-unique but reusable over time, so it names a slot rather than a resource.
Ownership is therefore bound to two provider-issued identities that configuration cannot
forge:

- **The host.** The provider scope is a digest of the Parallels service's own server and
  hardware identifiers, read from `prlsrvctl info` — the Parallels analogue of the Incus
  daemon certificate. Only attested values enter it, never a fleet entry's name, its
  configured address, or the account used to reach it. Repointing an entry at another
  machine is refused; the same machine answering at a new address still owns and can still
  stop its leases, which matters because otherwise a routine address change would strand
  running VMs with no CLI route to delete them. Hashing keeps host identifiers out of
  claims, labels and errors.
- **The VM incarnation.** Parallels issues a fresh UUID per clone and never reuses one.
  The UUID a successful clone returns is pinned before any further reconciliation, and is
  then required unconditionally for adoption, guest preparation and deletion.

`internal/cli/fixed_lease.go` already owns the claim lock, the
`prepared → acquired → deleting → released` state machine, fingerprint and provider-scope
conflict detection, and TTL expiry. The provider supplies `prepare` and `acquire` callbacks
[...]

PR #2388, opened 2026-09-20
fix(parallels): install the macOS Node readiness baseline

### LINKED ISSUE #2381: Parallels macOS leases hang at readiness because nothing installs the `node`/`npm` baseline the readiness probe requires

saariuslystoned:
Fixes https://github.com/openclaw/crabbox/issues/2381

## Problem

`sshReadyCommand` requires `node --version` and `npm --version` for every macOS
target without an explicit `ReadyCheck`, but nothing on the Parallels path
installs them, so macOS leases never reach readiness.

The mechanism is subtler than a missing package. `parallelsPOSIXEnsureReadyScript`
starts with a shortcut:

```sh
if [ -x /usr/local/bin/crabbox-ready ] && /usr/local/bin/crabbox-ready >/tmp/crabbox-ready.log 2>&1; then
  ...
  exit 0
fi
```

The macOS branch of that generated helper checked only `rsync`, `curl` and the
work root — never Node. On a template that already ships a passing
`crabbox-ready`, guest preparation therefore exits in its first few lines and
does essentially nothing, while the readiness probe keeps failing on `node`.
Appending an installer to the end of preparation would not have helped.

Crabbox already owns a checksum-pinned macOS Node installer
(`recipes/bootstrap/v1/macOSNodeInstall.sh`, generated into
`sharedMacOSNodeInstall`), but it was reachable only from the AWS path.

## Change

Run that same installer in both Parallels preparation paths — the Parallels
Tools path and the `BootstrapMacOSOverSSH` fallback, which both funnel through
`parallelsPOSIXEnsureReadyScript` — before the `crabbox-ready` shortcut.

Preparation is driven through `/bin/sh`, and the installer is `#!/bin/bash` with
`set -euo pipefail`, so it is written to a file and handed to `/bin/bash`:

- `pipefail` never reaches the POSIX shell
- the installer's own `exit 0` ends only the installer, not preparation
- `</dev/null` keeps it from consuming the rest of a script that the SSH
  fallback streams in on stdin
- a failed install stops preparation with a named cause instead of a silent
  readiness timeout

The installer already returns immediately when `node` and `npm` resolve, so
healthy guest runtimes are preserved and no download happens.

The generated macOS `crabbox-ready` now asserts the same baseline on an explicit
PATH, matching the managed macOS contract in `sharedMacOS`. Linux guests are
untouched.

## Diagnostic

This failure surfaced as a transport problem:

```
timed out waiting for SSH on <ip> during bootstrap probe=readiness cause=deadline_exceeded authentication=unknown
```

Transport and authentication were both fine. When the transport probe succeeds
it has run a remote command, which proves authentication, so the timeout now
reports `authentication=ok`, marks the port `:ready` rather than `:auth`, and
names `probe=readiness`.

Naming the specific missing command was considered and left out on purpose:
`runSSHReadinessProbe` deliberately discards remote stderr and keeps only
matcher state (see the comment on `sshReadinessDiagnostic`), so surfacing the
command text would mean walking back that bounded-buffer guarantee. Worth doing
separately if wanted.

## Tests

Committed before the implementation, and failing on the parent commit:

- `TestParallelsMacOSNodeBaselineRunsBeforeReadyShortcut` — the installer is
  embedded verbatim and ordered ahead of the shortcut
- `TestParallelsMacOSNodeBaselineStanzaIsolatesInstaller` — executes the stanza
  under a real `/bin/sh` with stubbed tools: a healthy runtime is left alone and
  preparation continues, a non-macOS guest is untouched, a failed install stops
  preparation, and `pipefail` never leaks into the POSIX shell
- `TestParallelsMacOSReadyScriptMatchesReadinessContract` /
  `TestParallelsLinuxReadyScriptUnchanged`
- `TestParallelsEnsureReadyScriptParsesUnderPOSIXShell` — `sh -n` over the
  generated script for both desktop modes
- `TestWaitForSSHReadyTimeoutReportsProvenAuthentication`

```
go test -race -timeout=20m ./internal/cli -run 'TestParallels|TestManagedMacOSNodeBaseline|TestMacOSNode'
go test -race -timeout=20m ./internal/providers/parallels/...
go vet ./internal/cli ./internal/providers/parallels/...
node scripts/check-docs-links.mjs
```

All pass.

## Native qualification

Run on an Apple-silicon Parallels host against a macOS template with no Node,
using a full clone. Same template, same clone mode, same host on both sides.

On the guest during the stalled run, `node` and `npm` were absent, the template's
`crabbox-ready` returned success, and the readiness contract failed with
`node: command not found` — the shortcut passing while readiness fails is the
whole defect.

Before, on this commit's parent (terminated by hand after 3m of a 20m budget;
it never reached `provisioned`):

```
provisioning provider=parallels lease=... slug=prenode-before clone_mode=full
waiting for <ip>:2222 bootstrap ssh-transport... elapsed=5s remaining=19m55s ports=proxy
...
waiting for <ip>:2222 bootstrap ssh-transport... elapsed=3m9s remaining=16m51s ports=proxy
```

After, on the same template:

```
provisioned lease=... vm=... ip=<ip>
ready ssh=<user>@<ip>:22 network=public workroot=/Users/<user>/crabbox
warmup complete total=37.805s
```

`ready` is the meaningful line: `sshReadyCommand` for a macOS target requires
`node --version >/dev/null && npm --version >/dev/null` over SSH, so the lease
could not reach it without a working baseline. The guest ended up with
`/usr/local/lib/crabbox/node-v24.19.0-darwin-arm64` and links in
`/usr/local/bin`.

## Docs

`docs/providers/parallels.md` "Template requirements" now states that macOS
templates do not need to ship Node, and what preparation installs, mirroring the
equivalent contract in `docs/providers/aws.md`.

## Note

This touches `internal/cli/parallels.go` in a different function from
https://github.com/openclaw/crabbox/pull/2383, which is open against the same
file. Branched from `main` and kept independent; expect at most a small textual
conflict if both land.


saariuslystoned:
Closing in favour of https://github.com/openclaw/crabbox/pull/2387, which fixes the same issue and handles a case this PR got wrong.

Both PRs came from the same fleet and reached the same diagnosis independently: `sshReadyCommand` requires `node`/`npm` for macOS targets, nothing on the Parallels path installed them, and `parallelsPOSIXEnsureReadyScript` exits early on a template that already ships a passing `crabbox-ready`, so preparation did nothing.

#2387 is the better fix. This PR ran the pinned installer unconditionally. Because `macOSNodeInstall.sh` resets PATH to `/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin` while the readiness probe resolves Node through `remotePortableShellInvocation`'s `exec bash -lc` login shell, a template whose Node comes from Homebrew, nvm or asdf is invisible to the installer and visible to the probe. This PR would have downloaded over a healthy runtime, and failed preparation outright when nodejs.org was unreachable — turning an offline-capable template into one that needs the network to stay ready. #2387 resolves the runtime the way the probe does before reaching for the installer, and preserves what it finds.

The parts of this PR that were not in #2387 have been handed over as patches: executable regressions that run the stanza under a real `/bin/sh` (the string assertions could not catch `pipefail` leaking into the POSIX shell, or the installer's `exit 0` ending preparation), an `sh -n` parse check over the generated script, and the `docs/providers/parallels.md` template-requirements update.

The readiness-diagnostic change here is not part of that handover. It is orthogonal to the Parallels fix — a readiness-only failure reports `authentication=unknown` even after the transport probe has proven authentication, on both the direct and `SSHConfigProxy` routes — and `internal/cli/ssh.go` affects every provider, so it belongs in its own PR rather than riding along with a provider fix.

PR #2391, opened 2026-09-20
fix(ssh): report proven authentication on readiness timeouts

saariuslystoned:
## Problem

A lease that is reachable and authenticates, but whose readiness command
fails, times out like this:

```
timed out waiting for SSH on <ip> during bootstrap probe=readiness cause=deadline_exceeded authentication=unknown
```

Transport and authentication are both fine. The readiness command is the only
thing failing. But the message reports authentication as unknown, so the obvious
next moves are to check networking, the security group, the key and the SSH
service — none of which is wrong.

`waitForSSHReadyWithProbeContext` already has the evidence to say otherwise.
When readiness fails on the direct route it runs a transport probe, and that
probe executes a remote command over SSH. If it succeeds, reachability and
authentication are proven, not unknown.

Two things then misreported it:

- the port was recorded as `<port>:auth` and `authentication=unknown` was
  emitted regardless
- `lastProbe` was set back to `"transport"` after the transport probe
  succeeded, so the timeout named the probe that had just worked

## Change

Track whether any remote command has succeeded. When one has, report
`authentication=ok`, mark the port `:ready`, and name `probe=readiness`:

```
timed out waiting for SSH on <ip> during bootstrap probe=readiness cause=deadline_exceeded authentication=ok ports=22:ready
```

Nothing is inferred. `authentication=unknown` is still reported whenever no
remote command has succeeded — an unreachable host, a refused port, a rejected
key all read exactly as before.

### The proxy route

The `SSHConfigProxy` branch never ran a transport probe at all, so a
readiness-only failure there was indistinguishable from an unreachable host. It
now probes transport once readiness has failed, matching the direct route.

This matters more than it looks: a lease reached through a `ProxyCommand` takes
this route, so the case that most needs the distinction was the one that could
never make it. The extra probe runs only after readiness has already failed, so
a healthy lease is unaffected.

## What this does not do

It does not name the failing command. `runSSHReadinessProbe` deliberately
discards remote stderr and keeps only matcher state — see the comment on
`sshReadinessDiagnostic`, which retains "only matching state, never stderr
bytes". Surfacing the command text would mean walking back that bounded-buffer
guarantee, which is a larger decision than this change. Worth doing separately
if maintainers want it.

## Review follow-ups

Both findings from the first review are fixed, each with its regression
committed before the fix.

**Set the active stage before running the proxy transport probe.** The probe ran
while `lastProbe` was still `readiness`, so a probe that blocked until the
deadline was reported as `probe=readiness` when transport had consumed it. It
now sets the stage to `transport` before invoking and restores `readiness` once
transport answers.

**Propagate fatal errors from the proxy transport probe.** A host-key rejection
or workspace-owner setup error was discarded and retried into a generic timeout.
It now goes through `sshReadinessError` before continuing.

Both were the same underlying mistake: the call was added to the proxy branch
without carrying over the contract the direct branch already keeps for the
identical call. The two branches now match.

## Contributor-reported tests at d25aac2429ecc1f8be790ebe9713d4af83a7c7fc

`TestWaitForSSHReadyTimeoutReportsProvenAuthentication` over both routes, and
`TestWaitForSSHReadyProxyTransportProbeFollowsDirectContract` for the two
follow-ups. Each was committed before its fix and fails on its parent:

```
--- FAIL: .../direct
    timeout does not blame readiness: ... probe=transport ... authentication=unknown ports=<port>:auth
--- FAIL: .../proxy
    timeout still reports authentication as unknown after a successful transport probe:
    ... authentication=unknown ports=proxy
--- FAIL: .../blocked_transport_owns_the_reported_stage
    error does not contain "probe=transport": ... probe=readiness ... ports=proxy
--- FAIL: .../fatal_transport_error_stops_the_wait
    error does not contain "SSH host-key verification failed": ... timed out waiting for SSH ...
```

```
go test -race -timeout=20m ./internal/cli -run 'TestWaitForSSHReady|TestSSHReadiness|TestSSHWaitProgress|TestProbeSSH'
go vet ./internal/cli
```

Both pass, the readiness suite over three consecutive `-race` runs. Existing
coverage asserting `authentication=unknown` where nothing was proven is
unchanged and still passes.

## Contributor-reported native proxy behavior at d25aac2429ecc1f8be790ebe9713d4af83a7c7fc

Captured on a macOS guest reached over real SSH, on the proxy route — the route
this PR adds code to, and the one a `ProxyCommand` lease actually takes.

The guest was cloned from a snapshot taken before its Node baseline existed, so
`node` is genuinely absent and the readiness command fails for a real reason,
while transport answers normally. The reported binary is the original contributor revision, cut from `main` without the separate Node-installation change.

```
provisioning provider=parallels lease=<lease> host=<host> source=<template> clone_mode=full
waiting for <guest-ip>:2222 bootstrap ssh-transport... elapsed=6s remaining=19m54s ports=proxy:ready
waiting for <guest-ip>:2222 bootstrap ssh-transport... elapsed=19m34s remaining=26s ports=proxy:ready
waiting for <guest-ip>:2222 bootstrap ssh-transport... elapsed=19m59s remaining=1s ports=proxy:ready
timed out waiting for SSH on <guest-ip> during bootstrap probe=readiness cause=deadline_exceeded authentication=ok ports=proxy:ready; next_action=check lease status, then stop or replace the lease if SSH stays unreachable
```

Addresses, hostnames, user and lease identifiers are redacted.

Three things to note. `ports=proxy:ready` appears from the first progress line,
six seconds in: the transport probe this PR adds to the proxy route ran and
succeeded, which on unpatched code cannot happen because that route never probed
transport at all — it prints `ports=proxy`. The timeout reports
`authentication=ok`, which is the change under review, and `probe=readiness`
rather than the transport probe that had just succeeded. Before this PR the same
run ends in `authentication=unknown ports=proxy`.

The direct route is covered by unit tests rather than a live transcript. The
reporting itself — the `authenticated` flag, the `authentication=` field and the
timeout message — is shared by both routes and is exercised end to end above;
what differs is only where the flag is set. A live direct-route capture needs a
guest that is SSH-reachable while failing readiness, which the fixture available
here could not hold open long enough to record. Happy to add one if a reviewer
wants it.

## Origin

Found while fixing https://github.com/openclaw/crabbox/issues/2381, where macOS
Parallels leases stall at readiness because nothing installs the `node`/`npm`
baseline the probe requires. That failure surfaces as the transport-shaped
timeout above. The provider fix is
https://github.com/openclaw/crabbox/pull/2387; this is split out because
`internal/cli/ssh.go` affects every provider and the diagnostic stands on its
own.

## Maintainer follow-up

Review found and fixed another edge case: after a completed direct-port scan, a closed fallback must not overwrite the readiness stage proved by a successful transport command on another port. Active in-flight probes still retain their own stage.

Added installed-OpenSSH tests for direct connections and actual ProxyJump forwarding, with successful transport and no-success controls. They use pinned synthetic host keys, private SSH configuration, no account keys or agent, and a session endpoint that executes only fixed harmless fixture commands. The direct-success case includes a closed fallback. An explicit credential-free Linux CI step records JSON execution evidence and requires the root test plus all four cases to run and pass without skips.

Formatting, diff checks, and independent isolated Codex review through P2 passed. These new native tests have not been executed locally; the revised head has now passed isolated CI, including the explicit real-OpenSSH step. The earlier hardware transcript remains contributor-reported evidence for the original revision, not proof of these maintainer additions. The Unreleased changelog includes contributor credit.

The first revised CI run passed the CLI package but found that the exact workflow-step contract had not been updated for the new required SSH step. The follow-up adds that step to the contract and regression mutations guarding its no-skip and proxy-case assertions; independent review passed. Revised CI passed: https://github.com/openclaw/crabbox/actions/runs/35528370246. The explicit JSON gate recorded run/pass events for the root test and all four direct/proxy success/control cases with no skips. Ubuntu OpenSSH 9.6p1 executed the fixed fixture commands; both proxy cases recorded two real forwarded sessions. All 22 head checks are complete and successful or intentionally skipped.

Revised head: `ab6c8542bc655f8dab6bb008bca5b7a0b261a7a2`.


saariuslystoned:
Following up on the side observation relayed from this work — a Parallels macOS guest whose sshd never started on the direct (host-local) route, with `prlctl exec` still working. I investigated it on aiworker-01 against disposable full clones of `aiworker-01-macos-clean-base` (guest macOS 26.5.2).

**It did not reproduce.** The direct route provisions a working guest end to end:

```
provisioned lease=cbx_a6186b99b509 vm=762f3df8-276e-42a7-8c27-d80a64e483c9 ip=10.211.55.66
running on 10.211.55.66 echo CRABBOX_E2E_OK
CRABBOX_E2E_OK
run summary lease=27.943s bootstrap=223ms command=393ms exit=0
```

`prepareGuest` (`internal/providers/parallels/backend.go`) is route-agnostic — direct and proxy run the same script and differ only in prlctl transport — so there is no code path by which the direct route alone disables sshd.

### Most likely explanation: clone mode

`--parallels-clone-mode` defaults to `linked`, and on this Apple silicon host a linked clone boots but never reports an IP; Parallels Tools never came up either, across 40 attempts over ~5 minutes:

```
timed out waiting for Parallels VM a99dcbd5-73c4-47dd-8b3c-4543f43084e3 IP; last_state=running
```

`full` works reliably. "VM running, alive under `prlctl exec`, unreachable over SSH" is exactly what a linked clone looks like from the caller's side — the guest is fine and sshd is fine, there is just no reachable address. Worth checking whether the failing run passed `--parallels-clone-mode full`; the transcript in this PR's description did (`clone_mode=full`), so the sshd observation likely came from a different invocation.

### Two things not worth chasing

**`Load failed: 5: Input/output error`** in `/tmp/crabbox-remote-login.log` is benign. It is just `launchctl load -w` against an already-loaded `ssh.plist`; sshd stays up and reachable. The block does work when sshd is genuinely down — unloading it with `launchctl bootout system/com.openssh.sshd` reproduces the reported state, and re-running ensure-ready recovers it fully (`788 0 com.openssh.sshd`, `TCP *:22 (LISTEN)`, `Remote Login: On`), with the log empty because `load -w` then succeeds.

**`systemsetup -setremotelogin`** cannot work over `prlctl exec` at all:

```
setremotelogin: Turning Remote Login on or off requires Full Disk Access privileges.
```

The Tools exec context lacks the TCC grant on macOS 26. Without `-f` it also prompts on stdin and spins indefinitely non-interactively. That explains why manual attempts to enable it that way failed.

### Two real defects found along the way

Neither explains the original report, but both are in this code path:

- openclaw/crabbox#2396 — `prlctl exec` flattens argv, so the leading `set -eu` in the generated guest-prep script is consumed as arguments to an inner shell and the body runs with abort-on-error never in effect. Affects `EnsureGuestReady` and `InstallSSHKey`. Unit tests mock the runner and cannot see it.
- openclaw/crabbox#2397 — macOS `crabbox-ready` never verifies sshd is listening, while every `launchctl` call is `|| true`, so a guest with no sshd would still report ready and surface later as a generic `waitForSSHReady` timeout. That is the readiness-diagnostic failure mode this PR is about, reached from the provider side.

Scope: single host, one template, macOS 26.5.2. The proxy route was not exercised, so route-specific quoting through `shellWords` is untested, and I did not reproduce the original failure — this is a non-reproduction plus two adjacent findings, not a refutation of what was observed.

steipete:
Landed in https://github.com/openclaw/crabbox/commit/55875abc35733978a13b0ec0fc5f27a77792c323 through the normal protected merge path. The merge tree matched the reviewed current-main composition.

The timeout now distinguishes successful SSH authentication from a guest-readiness failure. Proxy routes receive the same diagnostic transport check, and a completed scan's closed fallback port no longer replaces a readiness-stage diagnosis after another port authenticated successfully. Existing active-probe cancellation and fatal-error handling are retained.

Verification: [exact-head CI](https://github.com/openclaw/crabbox/actions/runs/35528370246) and [post-merge main CI](https://github.com/openclaw/crabbox/actions/runs/35529938490) passed. All substantive source checks passed; the one cancelled check was a verified issue-comment metadata dispatch. The required installed-OpenSSH gate recorded run/pass events for the root and all four direct/ProxyJump success/control cases, with no skips and real forwarded-session/command evidence. Independent isolated reviews were clean. Post-merge macOS validation passed all 11 new diagnostic test/subtest events, 120 workflow-contract test/subtest events, vet, and the full CLI build.

Local caveat: the broader macOS readiness selector also selected two existing workspace-owner fixture groups and failed seven root/subtest events there. That run is retained as failed, not counted as a broad pass. Source inspection found unchanged early setup paths returning before the new diagnostic logic; no baseline rerun or definitive underlying environment diagnosis is claimed. Full main CI remained green. Contributor hardware evidence stays attributed to its original revision; no new cloud or guest provisioning was performed by the maintainer.

PR #2385, opened 2026-09-20
fix(parallels): hold a host capacity reservation across count and clone

saariuslystoned:
Fixes https://github.com/openclaw/crabbox/issues/2384

## Problem

A Parallels fleet host's `maxVMs` limit was a check-then-act. Each fork counted the host's live `crabbox-` VMs from a fresh `ListVMs` and then cloned into it, with nothing holding the host between the two steps. `crabbox shard` provisions every shard in its own goroutine (`internal/cli/shard.go:290-302`), so N concurrent forks all read the same pre-clone inventory, all pass the gate, and all clone.

Measured on a real Parallels host configured `maxVMs: 2`, with `crabbox shard --count 8`: **8 VMs created, 0 shards refused.**

## Change

`ReserveParallelsFleetCapacity` selects a fleet host and holds a per-host `flock` from the capacity count until the caller releases it. The Parallels backend's acquire path holds that reservation until the clone has completed and is visible to the next `ListVMs`, then releases it — bootstrap after the clone runs outside the reservation so it does not keep other forks waiting.

A host with no `maxVMs` has no limit to enforce and is not serialized, so the unconfigured case stays fully parallel.

`SelectParallelsFleetConfig` keeps its signature for the two callers that only inspect a host and never create a VM — `Doctor` (`internal/providers/parallels/backend.go:364`) and `checkpoint fork --dry-run` (`internal/cli/checkpoint.go:1757`) — and is now documented as advisory.

The lock follows the repo's existing primitives: `internal/cli/claim_lock.go:34`, `internal/cli/fixed_lease.go:26`, `internal/cli/checkpoint_store.go:232`, all `gofrs/flock`, including the "never unlink this lock" rule so a waiter keeps using the same inode after unlock.

## Contributor-reported result at b7251c177f1d2412b37dde59744e0f258df74d9b

Same host, same fleet config, same command:

| Build | `crabbox-` VMs on the host | Shards provisioned | Shards refused |
|---|---|---|---|
| `033bcb47` | 8 | 8 | 0 |
| original contributor revision | **2** | **2** | **6** (exit 5, `Parallels host <name> is at maxVMs capacity`) |

## Contributor-reported verification

```
go build ./...
go vet ./...
gofmt -l internal/
go test ./internal/providers/parallels/ -count=1
go test ./internal/cli/ -run 'Parallels|Shard|CheckpointFork' -count=1
go test -race -count=3 -run Parallels ./internal/cli/ ./internal/providers/parallels/
```

[...]

PR #2392, opened 2026-09-20
fix(parallels): cap direct-host VMs with parallels.maxVMs

### LINKED ISSUE #2386: Parallels host override silently discards maxVMs, and a direct host has no capacity limit
## Summary

`maxVMs` exists only as a field on `parallels.hosts[]` entries. Overriding the host with `CRABBOX_PARALLELS_HOST` or `--parallels-host` discards the fleet list, so the capacity limit silently stops applying, and there is no direct-host equivalent to restore it.

The override discarding the fleet is clearly deliberate — `internal/providers/parallels/provider.go:129-138` says so:

```go
if core.FlagWasSet(fs, "parallels-host") {
	cfg.Parallels.Host = *v.Host
	// An explicit host is a direct-host override, not another fleet hint.
	// Leaving configured fleet candidates here silently replaces the flag.
	cfg.Parallels.Hosts = nil
	cfg.Parallels.SelectedHost = ""
}
```

`internal/cli/config.go:4546-4554` does the same for the environment variable. That part is not the complaint.

The complaint is the consequence: `parallelsHostMaxVMs` (`internal/cli/parallels.go:1251-1258`) resolves the limit by scanning `cfg.Parallels.Hosts`, so with the fleet nil it returns 0, and `parallelsHostWithinCapacity` (`:1260`) treats 0 as unlimited and returns `true` unconditionally. An operator who had a limit now has none, with no warning.

## Why this is easy to hit

The credential guard's own remediation text points users straight at the action that drops the limit:

```
provider=parallels refuses repository-configured parallels.host with inherited credentials;
set CRABBOX_PARALLELS_HOST or --parallels-host to explicitly approve the credential destination
```

A repository-configured fleet with `maxVMs` therefore has two states: refused outright, or accepted with the limit silently removed. The only way to keep both the limit and the approval is to move the fleet into user config, which the message does not mention.

## Observed

Configuring a fleet host with `maxVMs: 2` and then approving the destination with `CRABBOX_PARALLELS_HOST`, `crabbox shard --count 8` created **8 VMs with 0 capacity refusals** — not because the limit was raced, but because it was no longer in effect. Removing the environment variable and moving the same fleet into user config, the same command refused 6 shards with `at maxVMs capacity`.

The tell is refusals, not VM count: a low VM count with zero `at maxVMs capacity` errors means the limit was never applied.

## Impact

`maxVMs` reads as a safety limit on a local hypervisor, where overfilling wedges the machine rather than just costing money. Losing it silently, as a side effect of following the CLI's own advice, is a surprising failure mode. This is also a trap for anyone reproducing capacity behaviour — it can make a capacity bug look fixed, or a fix look broken.

## Suggested options

Either or both:

1. Warn when a host override discards fleet entries that carried a `maxVMs`, naming the limit being dropped.
2. Give a direct host a capacity limit of its own — a top-level `parallels.maxVMs` with a `CRABBOX_PARALLELS_MAX_VMS` override — so approving a destination does not require giving up the limit.

Extending the guard's message to mention that user config is not subject to it would also help on its own.

## Related

https://github.com/openclaw/crabbox/issues/2384 is a separate defect in the same area — the fleet limit is a check-then-act when it *is* in effect. This issue is about the limit not being in effect at all. They are independent: fixing one does not address the other.

saariuslystoned:
## Summary

Fixes https://github.com/openclaw/crabbox/issues/2386.

`maxVMs` exists only on `parallels.hosts[]` entries. Overriding the host with `--parallels-host` or `CRABBOX_PARALLELS_HOST` clears both `cfg.Parallels.Hosts` and `cfg.Parallels.SelectedHost`, so `parallelsHostMaxVMs` matches nothing and returns 0. A direct host therefore has no capacity limit at all.

This adds the missing replacement: a top-level `parallels.maxVMs` with a `CRABBOX_PARALLELS_MAX_VMS` environment override, applying only on the direct-host path.

The fleet-discard behavior on override is unchanged and deliberate — this PR does not touch that code path or its comment, and `TestApplyFlagsExplicitHostBypassesConfiguredFleet` still passes.

## Updated for https://github.com/openclaw/crabbox/pull/2385

Rebased onto current `main` as a single clean commit (no merge commit). https://github.com/openclaw/crabbox/pull/2385 extracted the limit lookup into `parallelsHostMaxVMs`, which **both** `lockParallelsFleetCapacity` and `parallelsHostWithinCapacity` now consult.

The direct-host fallback lives in that resolver, not in the inventory check. That distinction is load-bearing: `lockParallelsFleetCapacity` skips the reservation entirely when the resolver returns `<= 0`, so supplying the limit only to the count would leave a configured direct host advisory-only — concurrent forks would all observe the same pre-clone inventory and clone past the limit. Resolving it in `parallelsHostMaxVMs` makes locking and counting honor the setting together.

This is ClawSweeper's P2 finding, and it was correct — thank you. The whole production change against current main is now:

```go
 func parallelsHostMaxVMs(cfg Config) int {
 	for _, host := range cfg.Parallels.Hosts {
 		if cfg.Parallels.SelectedHost == firstNonBlank(host.Name, host.Host, "local") {
 			return host.MaxVMs
 		}
 	}
-	return 0
+	// No fleet entry selected, so this is the direct host. A matched fleet entry
+	// returns above even at zero, so this is not a default for fleet entries.
+	return cfg.Parallels.MaxVMs
 }
```

plus the config plumbing for the new field. A direct host already flows through the entire selection, reservation, and capacity path because `ParallelsCandidateConfigs` returns `[]Config{cfg}` when `Hosts` is empty — only the limit lookup was missing.

## Precedence contract

A selected fleet entry's own `maxVMs` always wins; the top-level `parallels.maxVMs` applies **only when no fleet entry was selected** (the direct-host path) and is **not** a default for fleet entries that omit `maxVMs`. A matched entry returns above even at zero, so a fleet host with `maxVMs: 0` stays unlimited.

Absent config the value is 0, which means unlimited, exactly as today. The change is strictly opt-in.

## Changes

- `internal/cli/config.go`: `MaxVMs int` on `ParallelsConfig`; `maxVMs` on `fileParallelsConfig` plus the file→config mapping; `CRABBOX_PARALLELS_MAX_VMS` in `applyEnv`. All mirror the existing per-host field.
- `internal/cli/parallels.go`: `parallelsHostMaxVMs` returns `cfg.Parallels.MaxVMs` when no fleet entry matched.
- `docs/providers/parallels.md`: documents `parallels.maxVMs`, `CRABBOX_PARALLELS_MAX_VMS`, the precedence sentence, that a limited direct host takes the same reservation as a limited fleet host, and that an explicit host override discards the fleet's `maxVMs` limits.

The contributor left changelog maintenance to the maintainer; the follow-up below adds the feature entry.

## Regression tests

`TestParallelsDirectHostCapacityBoundsConcurrentForks` in `internal/cli/parallels_capacity_test.go`, alongside the fleet counterpart from https://github.com/openclaw/crabbox/pull/2385 and reusing its stateful `parallelsCapacityFakeHost` (clones become visible to later lists, and the host records peak concurrent `crabbox-` VMs). 8 concurrent forks run the same reserve → count → clone → release sequence as `acquireOnce`, against a direct host configured `maxVMs: 2`.

It is deliberately built to fail against the count-only shape, so it covers the reservation half rather than just the count:

**A. No fix (current main behavior):**
```
--- FAIL: TestParallelsDirectHostCapacityBoundsConcurrentForks (0.06s)
    parallels_capacity_test.go:363: direct host overfilled: peak crabbox VMs=8 maxVMs=2 clones=8 refused=0
```

**B. Count-only fix — limit supplied to `parallelsHostWithinCapacity` but not to the resolver (the shape ClawSweeper flagged):**
```
--- FAIL: TestParallelsDirectHostCapacityBoundsConcurrentForks (0.06s)
    parallels_capacity_test.go:363: direct host overfilled: peak crabbox VMs=8 maxVMs=2 clones=8 refused=0
```
Still 8 clones on a `maxVMs: 2` host, because the reservation is skipped.

**C. With the fix as submitted:** PASS — 2 clones, 6 refusals with `at maxVMs capacity`, peak 2.

Also added:
- `TestParallelsDirectHostUnlimitedBypassesReservation` — an unset direct host stays lock-free and fully parallel, keeping the change opt-in (mirrors `TestParallelsUnlimitedCapacityBypassesReservation`).
- `TestSelectParallelsFleetConfigAppliesDirectHostMaxVMs` in `internal/cli/parallels_test.go` — table-driven over six precedence cases: direct host at limit, below limit, unset, fleet limit wins over a lower top-level value, fleet limit applies over a higher one, and top-level-is-not-a-fleet-default.

## Contributor-reported verification

Run on the rebased head `e45e0760`:

| Command | Result |
| --- | --- |
| `go build ./...` | pass |
| `go vet ./...` | pass |
| `gofmt -l internal/ cmd/` | no output |
| `go test -race -count=1 -run Parallels ./internal/cli/ ./internal/providers/parallels/` | `ok internal/cli 5.712s`, `ok internal/providers/parallels 3.027s` |
| `go test ./internal/providers/parallels/ -count=1` | `ok internal/providers/parallels 1.854s` |

The full `./internal/cli/` package suite was not run locally (40+ minutes on this machine); CI covers it here.

## Real-host proof

Now supplied, on a physical Apple silicon Mac running Parallels 27.0.1 — full redacted logs in https://github.com/openclaw/crabbox/pull/2392#issuecomment-5751442990.

The host already held 2 `crabbox-` VMs from an unrelated workload, so every limit below was enforced against a genuine live inventory count. Same host, same session, `main` @ `687f5741` vs this PR @ `e45e0760`:

| # | Scenario | before (`main`) | after (this PR) |
|---|---|---|---|
| A/B | direct host, `maxVMs=2`, host already holds 2 | exit 1, passed the gate, **created a 3rd VM** | **exit 5**, `is at maxVMs capacity`, 0 created |
| C | fleet `maxVMs: 1` + top-level `maxVMs=9` | — | **exit 5** — per-host wins over the higher top-level |
| D | `maxVMs` unset | — | passes the gate, unlimited preserved |
| E | 6 concurrent warmups, `maxVMs=4`, 2 pre-existing | `admitted=6 refused=0` | `admitted=2 refused=4`, final inventory exactly 4 |
| F | `CRABBOX_PARALLELS_HOST` override discarding a fleet `maxVMs: 1`, over SSH to a remote Mac | passed the gate and provisioned | **exit 5** naming the overriding host |

F is the scenario in the issue title, run from a second machine against the Parallels host over SSH.

Two things I want to state plainly rather than let the numbers imply: in E's before-run all 6 forks cleared the gate but only 1 VM materialized, because concurrent `prlctl clone` against one source contends at the Parallels level — the evidence is `admitted=6 refused=0` vs `admitted=2 refused=4`, not a VM count, and I am not claiming `main` produced 8 VMs. And guest VMs on this host fail to boot for a pre-existing reason unrelated to this change, identically on both binaries, strictly after the clone and the capacity gate; it is visible in the logs and affects none of the results.

Every VM created was deleted and the host was returned to its exact starting inventory.

## What this does not cover

- **It does not warn when an override discards a positive fleet limit.** That was the other option on https://github.com/openclaw/crabbox/issues/2386 — the override still silently drops configured `maxVMs` values, and a user who wants a cap on the overriding host has to set the new one.
- **It does not make the top-level value a default for fleet hosts.** A fleet entry that omits `maxVMs` stays unlimited, as it is today.
- **Reservation scope is inherited from https://github.com/openclaw/crabbox/pull/2385 and unchanged.** Callers with different state directories, or on different machines, still race against the same host.

## Note on the ClawSweeper recommendation

ClawSweeper originally recommended the warn-only approach on the issue and applied `clawsweeper:no-new-fix-pr` / `clawsweeper:needs-product-decision`. This PR knowingly takes the other option, mainly because the direct host already flows through the full capacity path, so this was the smaller change — whereas the warn-only path has no stderr or warning channel at either override site. Its follow-up review then recommended approving the opt-in direct-host limit once it used main's reservation mechanism, which is what this revision does.

The configuration contract is still a product call and entirely yours: happy to rework this as a warning instead, adjust the precedence, or close it if you would rather settle the shape first.

## Maintainer decision and repair

Accepted the opt-in direct-host limit, not a warning-only replacement. A selected fleet entry retains its own limit, including unlimited; the top-level setting is not an implicit fleet default. Existing explicit-host override behavior remains unchanged.

Review found a configuration-layering edge case: explicit YAML `maxVMs: 0` was ignored after an inherited positive limit. The maintainer repair distinguishes an omitted value from explicit zero, adds seven YAML/environment layering cases, and clarifies that nonpositive values are unlimited, stopped inventoried Crabbox VMs count, and reservations coordinate only callers sharing state and the configured host/account. The setting is YAML/environment-only. Added the credited Unreleased feature entry.

Independent isolated Codex review through P2 is clean after staging the repair. The revised head is `d73147692fb5a90a7170dac4de063c51e9651326`; its full isolated CI passed: https://github.com/openclaw/crabbox/actions/runs/35528910059. The final composition with current main was independently reviewed through P2 with no findings; its non-changelog patch is unchanged, and existing release-note entries are preserved. No contributor code was executed locally. The native evidence above remains contributor-reported for `e45e0760`; reservation/count/admission mechanics are unchanged by this configuration-only repair.


saariuslystoned:
## Real-host proof on Parallels hardware

Ran the requested after-fix evidence on a physical Apple silicon Mac. Host details and identifiers are redacted; counts, exit codes, and messages are verbatim.

**Environment**

| | |
|---|---|
| Host | Apple silicon Mac (arm64), macOS 26.5.2 |
| Parallels | `prlctl version 27.0.1 (58670)` |
| "before" binary | built from `main` @ `687f5741` |
| "after" binary | built from this PR @ `e45e0760` |
| Source VM | a macOS base VM, linked clone from an existing snapshot |

Both binaries were built with the same toolchain and run against the same host in the same session. The `maxVMs` limit was supplied through `CRABBOX_PARALLELS_MAX_VMS`, exercising the new environment override; case C supplies a fleet entry through YAML.

**The host was not empty.** It already held **2 `crabbox-` VMs belonging to an unrelated workload**, which is what makes this a real inventory test rather than a synthetic one: every limit below is enforced against a genuine live count, not a count the test arranged. Those two VMs were never touched, and every VM these runs created was deleted afterwards — the host was returned to its exact starting inventory.

### Sequential cases

| # | Scenario | Binary | Result |
|---|---|---|---|
| A | direct host, `maxVMs=2`, host already holds 2 | **after** | **exit 5**, `Parallels host local is at maxVMs capacity`, 0 VMs created |
| B | identical config | **before** | **exit 1**, passed the capacity gate, **created a 3rd VM despite `maxVMs=2`** |
| C | fleet entry `maxVMs: 1` **and** top-level `maxVMs=9` | **after** | **exit 5** refused — the per-host limit wins over the higher top-level value |
| D | `maxVMs` unset | **after** | passed the capacity gate — unlimited preserved, change stays opt-in |

Case B is the defect reproduced on hardware:

```
CASE: B. BEFORE(main 687f5741) + env maxVMs=2 -> expect NOT refused (the bug)
  crabbox- VMs on host before: 2
  exit code: 1
  capacity gate: PASSED   -> provisioning provider=parallels lease=cbx_<redacted> slug=<redacted> host=local
  crabbox- VMs on host after:  3
  NEW VMs created by this case:
      + {<redacted-uuid>}  crabbox-cbx-<redacted>-<redacted>
```

and case A is the same config under this PR:

```
CASE: A. FIXED + parallels.maxVMs=2 via env, direct host already holds 2 crabbox VMs -> expect REFUSED
  crabbox- VMs on host before: 2
  exit code: 5
  capacity gate: REFUSED  -> Parallels host local is at maxVMs capacity
  crabbox- VMs on host after:  2
```

### Concurrency — the reservation half

6 simultaneous `warmup` processes against the direct host, `CRABBOX_PARALLELS_MAX_VMS=4`, 2 `crabbox-` VMs already present, so **at most 2 new admissions are correct**.

**After (this PR):**
```
=== CONCURRENCY: 6 simultaneous warmups, direct host, CRABBOX_PARALLELS_MAX_VMS=4
=== crabbox- VMs already on host: 2  (so at most 2 new clones may be admitted)
  fork 1: REFUSED (exit 5)
  fork 2: REFUSED (exit 5)
  fork 3: ADMITTED (exit 1) slug=<redacted>
  fork 4: ADMITTED (exit 1) slug=<redacted>
  fork 5: REFUSED (exit 5)
  fork 6: REFUSED (exit 5)

=== RESULT: admitted=2 refused=4
=== crabbox- VMs before=2 after=4  limit=4
=== peak did not exceed limit: YES
```

Exactly 2 admitted, 4 refused, final inventory exactly at the limit.

**Before (`main`), identical config, host reset to the same 2-VM baseline first:**
```
=== RESULT: admitted=6 refused=0
```

All 6 passed the capacity gate; nothing was bounded.

One honest caveat on that last run: although all 6 were admitted, only 1 VM actually materialized, because 6 concurrent `prlctl clone` operations against one source snapshot contend at the Parallels level and most clones failed *after* clearing the gate. **The evidence here is `admitted=6 refused=0` versus `admitted=2 refused=4`, not the resulting VM count** — I am not claiming `main` produced 8 VMs. Case B is the clean single-threaded demonstration that an admitted fork does clone past the limit.

### The override path from the issue

This is the scenario in https://github.com/openclaw/crabbox/issues/2386's title, run from a second machine against the Parallels host over SSH — a genuine remote host, not local execution. Config declares a fleet entry with `maxVMs: 1`; `CRABBOX_PARALLELS_HOST` then overrides it, discarding that fleet; `CRABBOX_PARALLELS_MAX_VMS=2` is what should apply.

**Before:**
```
  exit code: 1
  capacity gate: PASSED  -> provisioning provider=parallels lease=cbx_<redacted> slug=<redacted> host=<redacted-host>
```
The override discarded the fleet's `maxVMs: 1`, no limit replaced it, and provisioning proceeded.

**After:**
```
  exit code: 5
  capacity gate: REFUSED -> Parallels host <redacted-host> is at maxVMs capacity
```

### Disclosure

Guest VMs on this host fail to boot (`Failed to start the VM: Unable to start the virtual machine`). That is a pre-existing condition of the host, unrelated to this change, and it reproduces identically on both binaries. It occurs strictly *after* the clone and the capacity gate, so it does not affect any result above — the capacity decision, the clone, and the VM-count deltas all happen before the guest is started. Flagging it rather than trimming it out of the logs.

### Cleanup

Every VM created by these runs was deleted. Final host inventory was verified identical to the starting inventory: the 2 unrelated `crabbox-` VMs, plus the base VM and one unrelated non-Crabbox VM, all untouched.

saariuslystoned:
@clawsweeper re-review

steipete:
Verified landing: https://github.com/openclaw/crabbox/commit/8074965275e4444c18db373794d8dd9d5c816d40. The opt-in direct-host capacity contract is now on main, with explicit YAML zero correctly overriding an inherited positive value, nonpositive values retaining unlimited behavior, and fleet-entry capacity kept separate. Documentation and the credited Unreleased entry are included.

Post-merge focused race coverage passed 177 test/subtest events with no skips, including all seven configuration-layer cases; vet and a full CLI build passed. The earlier native Parallels evidence remains contributor-attributed to revision e45e07603ae2a81b1b1f91a5a893c01230bb3b88, including admissions/refusals, fleet precedence, and restored inventory; no additional guest run is claimed here. Thanks @saariuslystoned.

The original main CI run https://github.com/openclaw/crabbox/actions/runs/35530797077 failed in the controller warmup-start fixture, not a capacity assertion. That failure remains recorded. The test-only repair landed through https://github.com/openclaw/crabbox/pull/2402; [subsequent full main CI](https://github.com/openclaw/crabbox/actions/runs/35535400432) passed with the capacity production code unchanged. The missing config-display field was separately fixed in https://github.com/openclaw/crabbox/pull/2400.

PR #2395, opened 2026-09-20
fix(cli): tolerate a lost race when creating lease SSH directories

saariuslystoned:
## Summary

`ensureLeaseSSHDirectories` and `walkDirectoryPathWithoutSymlinks` each build
their path one component at a time with an `os.Lstat` followed by a create when
the component is missing. That is a check-then-create: two callers that both
find a component missing will both try to create it, and the loser fails.

```text
create private lease SSH directory: mkdir <state>/crabbox: file exists
```

Any concurrent first-time lease work on a fresh state directory can hit this —
the per-lease key directory `crabbox/testboxes/<lease-id>` is built through this
path, and the shared `crabbox` and `testboxes` components are created by
whichever lease gets there first. It is not provider-specific; I hit it running
concurrent fixed-ID Parallels acquisitions in
https://github.com/openclaw/crabbox/pull/2383, which is why it is split out
here.

## Fix

Losing the race is not a failure, so both sites now tolerate `fs.ErrExist` and
fall through to the validation that already followed the create.

Both create sites retain their following `Lstat` and ordinary-directory checks.
The lease-directory path also retains its existing ownership and platform-specific
privacy checks. This repairs benign concurrent creation without changing those
validation rules or claiming stronger filesystem-race guarantees.

Nothing else changes: the create mode, the ownership checks, the symlink
refusals and the error text for genuine failures are all untouched.

## Contributor-reported tests on `403c89c5`

`internal/cli/lease_test.go`:

- `TestEnsureLeaseSSHDirectoriesToleratesConcurrentCreation` — eight concurrent
  first-time callers all succeed. This fails on `main` with the `file exists`
  error above.
- `TestEnsureLeaseSSHDirectoriesRefusesSymlinkComponent` — a symlink planted at
  a component is refused.
- `TestEnsureLeaseSSHDirectoriesRefusesFileComponent` — a plain file at a
  component is refused.

The two refusal tests pass both before and after the change. They are there to
pin the security property across it, not just to assert it afterwards: applying
only the test file to a clean tree shows the concurrency test failing while both
refusals already pass.

## Real behavior

Two concurrent `crabbox warmup` leases against a real Parallels host, sharing one
fresh `XDG_STATE_HOME`, so both build the shared lease SSH tree and their own key
directories under it at the same time. Addresses, host and guest account names
and template names are redacted.

```text
$ XDG_STATE_HOME=<fresh> crabbox warmup --provider parallels ... &   # x2

leased cbx_a6ce7b2b507a slug=tidal-hermit provider=parallels server=3dbd511d-a67e-4276-aa99-0bbb5660fd58 ip=<guest-a> idle_timeout=30m0s
ready ssh=<guest-user>@<guest-a>:22 network=public workroot=<work-root>
warmup complete total=33.902s

leased cbx_b6c1bddcba62 slug=blue-crab provider=parallels server=473ed2b7-133e-4b6c-8483-0838944b2a12 ip=<guest-b> idle_timeout=30m0s
ready ssh=<guest-user>@<guest-b>:22 network=public workroot=<work-root>
warmup complete total=34.885s
```

The tree they created concurrently, with both per-lease key directories present
and the shared parents still private:

```text
per-lease key directories: 2
drwx------  <state>/crabbox
drwx------  <state>/crabbox/testboxes
```

Both released cleanly and the host was left with only its base template.

### What this evidence does and does not show

It shows the changed code path carrying real concurrent leases end to end, with
permissions intact. It is not a before/after discriminator: I could not make the
failure reproduce through the CLI. The window between the `Lstat` and the create
is microseconds, while separate `crabbox` processes arrive milliseconds apart, so
48 concurrent CLI leases against the unfixed binary (4 rounds x 12) produced zero
failures. The deterministic reproduction is the in-process test, which fails on
`main` and passes here; I would rather say that plainly than present two passing
runs as a comparison.

## Contributor-reported verification on `403c89c5`

```sh
gofmt -l internal/
go vet ./...
go test -race -count=1 -timeout=20m -run 'EnsureLeaseSSHDirectories|Testbox|LeaseSSH|SSHRoot|PrivateDirectory|SSHTransport' ./internal/cli/
```

All clean. The full `internal/cli` package does not fit a 20m `-race` budget on
my machine, which is a property of the package rather than of this change, so
the race run above is filtered to the lease, key and private-directory tests
this touches.

I also ran the wider `Lease|SSH|Testbox|Claim|Transport` set without `-race`.
Two tests fail there on my machine —
`TestRunCommandWritesBrokeredReusedAWSLeaseOutputBeforeCommand` and
`TestRunCommandInjectsReservedMetadataIntoStaticSSH` — each spending ~120s
before failing at `blocked_stage=ssh` against `static_127-0-0-1`. They want a
real SSH listener on localhost, which this machine does not serve. I confirmed
they are pre-existing rather than caused by this change by running the same two
tests on the base commit without it: identical failures, 242.72s there versus
242.71s here.

No configuration or secret implications.

## Maintainer follow-up

Kept the two existence-error guards and existing subsequent validation. Prepared state roots with the existing platform fixture helper, handled unavailable Windows symlink fixtures, and added ordinary concurrent creation coverage for the missing-parent path. These tests do not force a deterministic scheduling interleaving or establish stronger replacement-race guarantees; the invalid-component cases check pre-existing components.

Extended the existing disposable local-container CI fixture with four concurrent processes invoking the actual built CLI against distinct fixed lease IDs and one fresh shared state root. Each successful lease must run a command and stop. Cleanup checks require absent containers and keys and released fixed-ID tombstones without live resource metadata, matching the provider contract. The CI event gate requires the concurrency subtest to run and pass without skips. Native Windows/macOS gates also require the ordinary directory-concurrency cases.

Integrated current main and added the credited Unreleased fix entry. Independent isolated Codex review through P2 is clean after correcting an initial test expectation that fixed-ID claims disappear completely. No contributor code, Docker lifecycle, or tests were executed locally. The new runtime checks have now passed in secretless CI at `585cae43aaeb2fec15626139b0d4dfc62ba6103a`; exact-head results follow below.

## Follow-up on the native lifecycle gate

Two fixes on top of `test(cli): prove concurrent lease SSH directory creation`,
addressing the review of that commit:

- **Workflow fixture.** The gate requires run and pass events for
  `TestLocalContainerProviderE2E/concurrent-cli-warmups` as well as its parent,
  but the fixture's successful scenario emitted only parent events, so the
  Scripts gate failed on its own success case. The successful fixture now emits
  both, and two negative scenarios keep the subtest's failure modes covered:
  concurrency events missing entirely, and the subtest skipped. Both are
  rejected, which is what shows the gate checks the subtest rather than passing
  vacuously. `node --test scripts/connector-e2e-smokes-workflow.test.js` — 18
  pass, 0 fail.
- **Documented prerequisite.** The subtest drives an exact binary through an
  absolute `CRABBOX_BIN`, so the documented local invocation
  `go test -tags localcontainer ./cmd/crabbox` fails once Docker is available.
  `docs/providers/local-container.md` now documents building the CLI and setting
  that path.

## Final exact-head CI proof

At `585cae43aaeb2fec15626139b0d4dfc62ba6103a`, [full CI](https://github.com/openclaw/crabbox/actions/runs/35533582792) and [Connector E2E Smokes](https://github.com/openclaw/crabbox/actions/runs/35533582552) succeeded. The built-CLI concurrency child emitted run/pass events and completed four warmups, commands, and stops against one fresh state root. Its cleanup assertions confirmed absent containers and keys, with only released fixed-ID tombstones retained.

The native Windows and macOS gates each emitted run/pass for both ordinary directory-concurrency tests; all 24 required Windows and 19 required macOS cases ran and passed without skips. This proves successful concurrent operation on those tested paths, not a forced filesystem interleaving or stronger replacement-race guarantees.

The earlier Scripts failure was the positive workflow fixture missing the newly required child events. The contributor corrected that fixture and documented the built-binary prerequisite; the final gate also rejects a skip appearing after a pass. All final source checks passed. The maintainer's isolated P2 review is clean, and current-main composition preserves the reviewed non-changelog patch byte-for-byte.


steipete:
Verified landing: https://github.com/openclaw/crabbox/commit/c161f932797dc312487be6a44410e0c5c4753418 matches the expected reviewed composition. [Full main CI](https://github.com/openclaw/crabbox/actions/runs/35535565946) and [Connector E2E Smokes](https://github.com/openclaw/crabbox/actions/runs/35535565906) passed. The final source-check snapshot has no failed or pending source gates; two cancelled issue-comment ClawSweeper dispatches are metadata-only.

The exact merged revision's native local-container log confirms both the E2E root and concurrency child ran/passed, with four actual built-CLI warmups sharing fresh state. All ran commands and stopped; assertions verified absent containers and keys, with only released fixed-ID tombstones retained. No skips or failures occurred in those required tests. The reviewed PR head also passed all 24 Windows and 19 macOS required native cases without skips. Local post-merge ordinary concurrency tests passed, alongside the 252-event focused combined-main suite, vet, and full CLI build.

The production fix tolerates only an existing-component creation result, then retains subsequent directory validation. It does not claim stronger replacement-race guarantees or a forced scheduling interleaving. The earlier Scripts failure was repaired by updating the positive workflow fixture to emit the newly required child events, with missing/skip cases covered. Documentation and the credited changelog entry are on main. Thanks @saariuslystoned for the original fix, native Parallels evidence, and workflow/documentation follow-up.

PR #2399, opened 2026-09-20
fix(parallels): verify the macOS SSH listener during guest preparation

### LINKED ISSUE #2397: parallels(macOS): crabbox-ready never verifies sshd is listening, so guest prep can report ready with no SSH
## Summary

On macOS Parallels guests, `crabbox-ready` never checks that sshd is listening, while every `launchctl` call that enables remote login is suffixed `|| true`. A guest whose sshd failed to start would therefore still be marked ready and exit 0, and the caller would only discover the problem later as a generic `waitForSSHReady` timeout.

Found while investigating an unverified report of a Parallels macOS guest with no sshd. That specific failure did not reproduce — see openclaw/crabbox#2396 for the non-reproduction, the investigation context, and a separate defect in the same code path.

Environment: aiworker-01 (Apple silicon), Parallels, guest macOS 26.5.2 (25F84), template `aiworker-01-macos-clean-base`, guest account `parallels-01`. Tested against disposable full clones; template untouched.

---

## The gap

The macOS branch of `parallelsPOSIXEnsureReadyScript` generates this readiness check:

```sh
#!/bin/sh
set -eu
rsync --version >/dev/null
curl --version >/dev/null
test -w '/Users/parallels-01/crabbox'
```

No SSH probe. And the remote-login block immediately above it swallows every failure:

https://github.com/openclaw/crabbox/blob/55875abc35733978a13b0ec0fc5f27a77792c323/internal/cli/parallels.go#L935-L939

```sh
/bin/launchctl load -w /System/Library/LaunchDaemons/ssh.plist >"$remote_login_log" 2>&1 ||
  /bin/launchctl bootstrap system /System/Library/LaunchDaemons/ssh.plist >>"$remote_login_log" 2>&1 || true
/bin/launchctl enable system/com.openssh.sshd >>"$remote_login_log" 2>&1 || true
/bin/launchctl kickstart -k system/com.openssh.sshd >>"$remote_login_log" 2>&1 || true
```

So the one thing the guest exists to provide — a reachable sshd — is the one thing guest prep does not verify before declaring success.

The EC2 macOS bootstrap already gets this right in its `READY` block (`internal/cli/bootstrap_generated.go`):

```sh
ssh_ready=0
for port in ...; do
  if nc -z 127.0.0.1 "$port"; then ssh_ready=1; break; fi
done
test "$ssh_ready" -eq 1
```

## Suggested fix

Add the same `nc -z 127.0.0.1 <port>` probe to the Parallels macOS `crabbox-ready`, so readiness fails at the point of failure with a clear cause instead of surfacing as an SSH timeout later.

## Why not simply make the `launchctl` calls fatal

Because their failures are expected on healthy guests. On a guest where sshd is already up, `/tmp/crabbox-remote-login.log` contains:

```
Load failed: 5: Input/output error
Try running `launchctl bootstrap` as root for richer errors.
```

That is just `launchctl load -w` against an already-loaded `ssh.plist`. sshd stays up and SSH stays reachable. Making these calls fatal would turn a routine no-op into a provisioning failure. (Flagging this explicitly because the log line reads alarmingly and is a natural thing to blame when debugging.)

The block does work when sshd is genuinely down. Unloading it with `launchctl bootout system/com.openssh.sshd` produces a guest that is alive via `prlctl exec` with port 22 dead and `Remote Login: Off`; re-running ensure-ready recovers it completely:

```
--- launchctl ---   788   0   com.openssh.sshd
--- listener ---    launchd  1  root  TCP *:22 (LISTEN)
--- remotelogin --- Remote Login: On
```

with the log empty, because `load -w` succeeds when the service really is not loaded. A TCP connect and auth handshake from the host confirmed reachability afterwards.

So the correct change is to *verify the outcome*, not to fail on the individual attempts.

## Related: do not reach for `systemsetup` here

Recording this so the fix above is not "re-enable remote login with `systemsetup`":

```
$ prlctl exec $V /usr/sbin/systemsetup -f -setremotelogin off
setremotelogin: Turning Remote Login on or off requires Full Disk Access privileges.
```

The Parallels Tools exec context lacks the required TCC grant on macOS 26, so `systemsetup -setremotelogin` cannot work over `prlctl exec` in either direction. Without `-f` it also prompts on stdin and will spin indefinitely in a non-interactive run. `systemsetup` currently appears only in the EC2 bootstrap, not in the Parallels path; it should stay that way.

## Scope / caveats

Single host, one template, macOS 26.5.2 guest. The proxy route (`--parallels-host`) was not exercised. I did not observe a guest that actually failed this way in practice — sshd came up correctly in every run — so this is a latent gap in the readiness contract rather than an observed production failure.

saariuslystoned:
## Problem

macOS guest preparation enables remote login through `launchctl`, but every call is suffixed `|| true`, and the readiness helper it writes checks only `rsync`, `curl`, `node`, `npm` and a writable work root:

```sh
#!/bin/sh
set -eu
export PATH="/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin"
rsync --version >/dev/null
curl --version >/dev/null
node --version >/dev/null
npm --version >/dev/null
test -w /Users/.../crabbox
```

Nothing verifies sshd. A guest whose sshd never came up is declared ready and exits 0; the failure surfaces later as a generic `waitForSSHReady` timeout that names neither the cause nor the remedy. The one thing the guest exists to provide is the one thing preparation does not check.

Fixes https://github.com/openclaw/crabbox/issues/2397.

## Change

Add an SSH listener probe to the macOS readiness helper:

```sh
ssh_ready=0
for port in '2222' '22'; do
  if nc -z 127.0.0.1 "$port" >/dev/null 2>&1; then
    ssh_ready=1
    break
  fi
done
test "$ssh_ready" -eq 1
```

and gate the early exit on the same check, so a guest carrying a `crabbox-ready` written by an older crabbox is re-verified instead of short-circuiting past remote-login setup:

```sh
if [ -x /usr/local/bin/crabbox-ready ] && /usr/local/bin/crabbox-ready >/tmp/crabbox-ready.log 2>&1 && crabbox_ssh_listening; then
```

### Any candidate port, not the configured one

The probe accepts any candidate port. This matters: templates commonly serve on 22 while the lease is configured for 2222, and crabbox already falls back host-side (`internal/cli/ssh.go`). A probe restricted to the configured port would fail guests that provision successfully today — including the template every measurement below was taken on. Candidates come from `sshPortCandidates`, the same derivation the host-side probe uses, so the two cannot drift apart.

### The launchctl calls stay non-fatal

Deliberately. `Load failed: 5: Input/output error` is the ordinary result of `launchctl load -w` against an already-loaded `ssh.plist`; sshd stays up and reachable. Failing on it would reject healthy guests. This change verifies the *outcome* rather than the individual attempts.

### Linux is untouched

`crabbox_ssh_listening` returns success when `sw_vers` is absent, so the Linux branch — which manages ssh through systemd — behaves exactly as before.

## Verification

Acceptance criteria from the issue review:

```
go vet ./internal/cli ./internal/providers/parallels/...                      OK
go test -race -timeout=20m ./internal/cli -run 'TestParallels'                ok  80.462s
go test -race -timeout=20m ./internal/providers/parallels/...                 ok   7.606s
```

Two regression tests, committed before the fix at `b441070e` and failing on that parent:

- `TestParallelsEnsureGuestReadyVerifiesMacOSSSHListener` — the helper probes a listener and tries both candidate ports.
- `TestParallelsEnsureGuestReadyRechecksMacOSSSHListenerWhenHelperExists` — the early exit is gated on the listener, covering the older-template bypass.

The generated script passes `sh -n`.

### Native verification

Apple silicon host, guest macOS 26.5.2 (25F84), full clone of a stock template, direct (host-local) route:

```
provisioning provider=parallels lease=cbx_9fa3a49b04f9 clone_mode=full
using ssh port 22 for 10.211.55.69 (configured 2222 not ready)
provisioned lease=cbx_9fa3a49b04f9 vm=f96d6e68-639a-4d25-80cc-02e6279e3865 ip=10.211.55.69
running on 10.211.55.69 echo LISTENER_FIX_OK
LISTENER_FIX_OK
run summary lease=29.914s bootstrap=349ms command=414ms exit=0
```

That run exercises the port-mismatch case directly — configured 2222, guest serving on 22 — and provisions normally, which is the regression this change most needed to avoid.

The probe was also checked in both directions on a live guest, healthy and then with sshd unloaded via `launchctl bootout system/com.openssh.sshd`:

```
--- sshd UP: expect rc=0 ---
rc=0
--- sshd DOWN: expect rc=1 ---
rc=1
```

## Notes

Single host, one template, one Parallels version. The proxy route (`--parallels-host`) was not exercised natively; it shares this generator, and the transport is unchanged by this PR.

Changelog left to maintainers per `AGENTS.md`.


steipete:
Prepared the existing PR for landing at `c264d04530c09c28f8011409357346fc27d54635`: integrated current main, added the credited Unreleased entry and provider documentation, and clarified that the new check establishes listener availability, not SSH authentication. The later authenticated-readiness gate remains unchanged.

The production listener implementation is unchanged from the contributor revision. Independent isolated Codex review through P2 is clean. Existing contributor-reported native macOS proof remains attributed to the earlier revision; no new guest run, remote-host run, or cleanup observation is claimed by this follow-up. Updated-head CI is required before landing.

The separate stdin-transport proposal at https://github.com/openclaw/crabbox/pull/2401 is not included; its Linux and remote-route compatibility questions remain separate.

steipete:
Verified landing: https://github.com/openclaw/crabbox/commit/f7bed55d50de9b4be32ed1f52dc4984ba6e883f1 exactly matches the reviewed tree. [Full post-merge CI](https://github.com/openclaw/crabbox/actions/runs/35537415163) and [Connector E2E Smokes](https://github.com/openclaw/crabbox/actions/runs/35537415151) passed. All substantive source checks completed successfully; cancelled issue-comment ClawSweeper dispatches were classified separately as metadata-only.

The preparation check now catches a missing macOS listener earlier while preserving configured/fallback port selection, harmless launchctl outcomes, Linux behavior, and the later authenticated SSH gate. Post-merge focused race coverage passed 107 CLI and 72 provider test/subtest events with no skips, plus vet and full CLI build.

Native macOS evidence remains attributed to contributor revision 9782fcc77859f64656c897871b7b770b5b284885, whose production implementation is unchanged in the landed candidate. No fresh maintainer guest run, remote-host proof, or additional cleanup observation is claimed. Documentation and the credited Unreleased entry are on main; https://github.com/openclaw/crabbox/issues/2397 is closed by the merge. The distinct stdin-transport proposal at https://github.com/openclaw/crabbox/pull/2401 remains separate. Thanks @saariuslystoned.

PR #2401, opened 2026-09-20
fix(parallels): send guest prep scripts over stdin

### LINKED ISSUE #2396: parallels: prlctl exec flattens argv, silently dropping set -eu from guest prep scripts
## Summary

`prlctl exec` does not preserve argv boundaries: it joins its arguments into one string and re-parses that string with a shell inside the guest. Because `EnsureGuestReady` passes the whole generated script as a single argv element after `/bin/sh -lc`, the script's leading `set -eu` is consumed as arguments to an inner shell and the rest of the script runs with abort-on-error never in effect.

Found while investigating an unverified report that crabbox's direct (host-local) SSH route produced a Parallels macOS guest with no sshd. That specific failure did not reproduce (see below). A separate defect in the same code path — macOS `crabbox-ready` never verifying sshd — is tracked in openclaw/crabbox#2397.

Environment: aiworker-01 (Apple silicon), Parallels, guest macOS 26.5.2 (25F84), template `aiworker-01-macos-clean-base`, guest account `parallels-01`. Tested against disposable full clones; template untouched.

---

## The defect

https://github.com/openclaw/crabbox/blob/55875abc35733978a13b0ec0fc5f27a77792c323/internal/cli/parallels.go#L532

```go
result, err := c.prlctl(ctx, nil, "exec", vmID, "/bin/sh", "-lc", parallelsPOSIXEnsureReadyScript(...))
```

Demonstrated by installing an argv-dumping script in the guest (stdin *is* preserved by `prlctl exec`) and calling it with two arguments:

```
$ prlctl exec $V /bin/sh /tmp/args.sh 'a b' c
argc=3
argv[1]=[a]
argv[2]=[b]
argv[3]=[c]
```

Two arguments in, three out. Corroborating probes:

```
$ prlctl exec $V /bin/sh -c 'echo hello world'      # prints an empty line, not "hello world"
$ prlctl exec $V /bin/sh -c 'echo A; echo B'        # prints blank, then "B"
```

The guest-side error from a third probe shows the reconstructed command line directly:

```
bash: -c: line 0: syntax error near unexpected token `one'
bash: -c: line 0: `/bin/sh -c printf "argc_extra=%s\n" "$#"; ... done one two '
```

## Consequence

The generated script begins with `set -eu`. Because the whole script is flattened, the guest effectively runs `/bin/sh -lc set` with `-eu` as positional arguments — consuming line 1 — and every remaining line executes in the guest's *outer* bash, with `set -eu` never in effect.

Confirmed by running the real generated script and observing `BASH_EXECUTION_STRING=set` in the guest's environment dump, while the rest of the script still took effect (`/usr/local/bin/crabbox-ready` and the work root were both created).

Guest prep therefore runs with abort-on-error disabled. Today that is mostly benign, because the final `/usr/local/bin/crabbox-ready` still determines the exit status — but it defeats guards the script is written to rely on, e.g. `[ -x "$kickstart" ]` in the macOS `--desktop` branch, which is intended to abort and instead falls through.

`InstallSSHKey` has the same shape and the same exposure:

https://github.com/openclaw/crabbox/blob/55875abc35733978a13b0ec0fc5f27a77792c323/internal/cli/parallels.go#L422-L423

Existing unit tests mock the command runner, so they assert on the argv crabbox *intends* to pass and cannot observe this.

## Suggested fix

Feed the script to the guest over **stdin** (`prlctl exec <vm> /bin/sh` with the script on stdin) rather than as an argv element. stdin is preserved faithfully — that is how every probe script in this investigation was run. `BootstrapMacOSOverSSH` already uses the stdin shape (`sudo -n /bin/sh -s`).

---

## On the original report: not reproduced

The direct (host-local) route provisions a working macOS guest end to end:

```
provisioned lease=cbx_a6186b99b509 vm=762f3df8-276e-42a7-8c27-d80a64e483c9 ip=10.211.55.66
running on 10.211.55.66 echo CRABBOX_E2E_OK
CRABBOX_E2E_OK
run summary lease=27.943s bootstrap=223ms command=393ms exit=0
```

`prepareGuest` (`internal/providers/parallels/backend.go`) is route-agnostic — direct and proxy run the same script and differ only in prlctl transport — so there is no code path by which the direct route alone disables sshd.

### Likely explanation: `--parallels-clone-mode` default is unusable on Apple silicon

Now tracked separately as openclaw/crabbox#2398.

The flag defaults to `linked`. On this Apple silicon host a linked clone boots but never reports an IP, and Parallels Tools never came up either (40 attempts over ~5 min):

```
timed out waiting for Parallels VM a99dcbd5-73c4-47dd-8b3c-4543f43084e3 IP; last_state=running
```

`full` works reliably (lease provisioned in ~28s, SSH reachable, command exit 0). "VM running but unreachable over SSH" is very likely what the original report actually was. Consider erroring early with a clear message for `linked` on Apple silicon rather than timing out at IP discovery.

---

## Scope / caveats

Single host, one template, macOS 26.5.2 guest. The proxy route (`--parallels-host`) was not exercised, so route-specific quoting through `shellWords` is untested. The originally reported sshd failure was not reproduced; this defect and openclaw/crabbox#2397 are adjacent findings, not a confirmed explanation of it.

saariuslystoned:
## Problem

`prlctl exec` does not preserve argv boundaries. It joins its arguments into one string and re-parses that string with a shell inside the guest. `EnsureGuestReady` and `InstallSSHKey` both handed the whole generated POSIX script over as a single argv element after `/bin/sh -lc`, so the guest effectively ran `/bin/sh -lc set` with `-eu` as positional arguments — consuming line 1 — and executed every remaining line in its **outer** shell, with `set -eu` never in effect.

Guest preparation therefore ran with abort-on-error disabled, defeating guards the scripts are written to rely on, such as `[ -x "$kickstart" ]` in the macOS `--desktop` branch.

Fixes https://github.com/openclaw/crabbox/issues/2396.

## What this does

Deliver both scripts on **stdin**, which `prlctl exec` preserves verbatim — the shape `BootstrapMacOSOverSSH` already uses (`sudo -n /bin/sh -s`).

`prlctlWithStdin` carries the reader through both routes. `prlctl` is now a thin wrapper over it, so host selection, key handling, the remote `PATH=` prefix and the password-filtered child environment are untouched on the remote/proxy route; ssh forwards stdin to `prlctl` on the Parallels host.

Stdin delivery imposes a rule on the scripts themselves: any child that reads stdin swallows the remainder of the script while the shell still exits 0. https://github.com/openclaw/crabbox/pull/2387 established `</dev/null` for the macOS Node children; this extends it to **every** child in both scripts. Three groups matter most:

- **The Linux branch**, which reaches stdin for the first time here — `apt-get update`, both `apt-get install` calls, `systemctl daemon-reload` / `enable --now`.
- **The readiness gate**, `/usr/local/bin/crabbox-ready </dev/null >/tmp/crabbox-ready.log 2>&1`, which runs mid-script: anything it consumed would be lost.
- **The probes guarding that gate** — the desktop VNC checks (`nc -z 127.0.0.1 5900`, `systemctl is-active`) and, after the rebase, the macOS SSH-listener check from https://github.com/openclaw/crabbox/pull/2399.

Commands that are intentionally non-fatal keep their `|| true`.

## Commits

| | |
|---|---|
| `bc6ea357` | failing regression: both scripts must arrive on stdin and not in argv, plus the stdin-consumption rule |
| `47459e3e` | the fix |

Rebased onto current `main` (`2ad31a5a`); the rebase was clean, and on it `go vet ./...` is clean and `go test -race ./internal/cli -run 'TestParallels|TestManagedMacOSNodeBaseline|TestMacOSNode'` and `go test -race ./internal/providers/parallels/` both pass. Previously rebased onto `f7bed55d` after https://github.com/openclaw/crabbox/pull/2399 merged. That PR added the macOS SSH-listener probe `crabbox_ssh_listening`, which runs **mid-script** inside the readiness gate, so its `nc -z 127.0.0.1 "$port"` is now covered by the same `</dev/null` rule and asserted by the regression test. The `nc` inside the generated `crabbox-ready` helper needs no guard: the prep script already invokes that helper with stdin on `/dev/null`.

The regression commit fails on its parent, and only through its own two tests — the rest of `TestParallels` is green there.

The runner is mocked, so the tests cannot observe guest-side reconstruction; they pin the transport crabbox chooses, which is the part that is wrong. Native evidence below covers the guest side.

### In tree

```
go vet ./internal/cli ./internal/providers/parallels/...                                                  clean
go test -race -timeout=20m ./internal/cli -run 'TestParallels'                                            ok  10.9s
go test -race -timeout=20m ./internal/cli -run 'TestParallels|TestManagedMacOSNodeBaseline|TestMacOSNode'  ok  18.1s
go test -race -timeout=20m ./internal/providers/parallels/...                                             ok   5.9s
gofmt                                                                                                     clean
```

### Native, on a real Apple-silicon Parallels macOS 26.5.2 guest

Run twice: once against the pre-rebase head (the transport probes below), and again against the rebased head `e29afd08`, so the combination with https://github.com/openclaw/crabbox/pull/2399's listener probe is covered rather than assumed.

Disposable full clone of the host's macOS template, deleted afterwards; template left stopped with all 7 snapshots and its current snapshot unchanged.

**The mechanism, both shapes, same guest.** Old argv shape — `set -eu` is swallowed, the rest leaks, exit 0:

```
$ prlctl exec $V /bin/sh -lc "$(printf 'set -eu\nfalse\necho SHOULD_NOT_PRINT\n')"
...
BASH_EXECUTION_STRING=set
...
SHOULD_NOT_PRINT
exit=0
```

New stdin shape — `set -eu` is in effect, nothing leaks:

```
$ printf 'set -eu\nfalse\necho SHOULD_NOT_PRINT\n' | prlctl exec $V /bin/sh -s
exit=1
```

Argv flattening is inherent to `prlctl exec` and is not fixed by moving the script to stdin — it is simply no longer in the way. Positional arguments still split:

```
$ printf 'for a in "$@"; do printf "[%s]\n" "$a"; done\n' | prlctl exec $V /bin/sh -s 'a b' c
[a]
[b]
[c]
```

Exit status propagates (`exit 7` → `exit=7`), and a 60 KB script over stdin arrives whole.

**The real generated scripts**, dumped from this branch and piped to `prlctl exec <vm> /bin/sh -s` on a fresh clone: `install-key` exit 0, `ensure-ready` exit 0.

**End to end**, this branch's binary (`GOOS=darwin GOARCH=arm64`):

```
run summary lease=30.769s bootstrap=318ms sync=0s command=394ms total=893ms end_to_end=31.985s exit=0
run summary lease=31.256s bootstrap=324ms sync=0s command=422ms total=933ms end_to_end=32.512s exit=0
```

**After the rebase**, on a fresh full clone, the generated scripts from `e29afd08` piped to `prlctl exec <vm> /bin/sh -s`:

```
-- install-key over stdin                       exit=0
-- ensure-ready over stdin                      exit=0      (includes the #2399 listener probe)
-- guest-side state
   Connection to 127.0.0.1 port 22 succeeded!
   SSHD_LISTENING
   READY_OK                                                 (/usr/local/bin/crabbox-ready passes as root)
-- set -eu really in effect on this guest shell  exit=1      (no leak)
```

Template left stopped with all 7 snapshots and its current snapshot unchanged; every clone deleted.

### An unrelated flake I hit, with a control

Some runs on that host fail at `prlctl exec` with `PrlJob_GetRetCode: Invalid argument` / `PrlJob_GetResult: Invalid argument`, at whichever exec comes first. I ran `origin/main` built the same way as a control, alternating. Pre-rebase: this branch 2 pass / 2 fail, `origin/main` 1 pass / 2 fail. Post-rebase against `f7bed55d`: both binaries passed round 1 and both failed round 2, on the same round. It is pre-existing and not introduced here, but worth knowing if CI or another native run trips on it.

### Native, remote-host route (`--parallels-host`)

Earlier revisions of this body listed the remote route as unexercised; this section covers it. Crabbox ran on an operator Mac and drove a **remote Apple-silicon Parallels host** (macOS 26.5.2, `prlctl` 27.0.1) over `--parallels-host`, cloning a macOS template with `--parallels-clone-mode full`. The branch is rebased onto current `main` (`2ad31a5a`). The control is an unmodified `main` binary built the same way.

```
crabbox run --provider parallels --target macos \
  --parallels-host <host> --parallels-host-user <user> \
  --parallels-source <macOS template> --parallels-clone-mode full \
  --parallels-user <guest user> --no-sync -- /bin/sh -c 'echo GUEST_OK; sw_vers -productVersion; ls -ld /var/lib/crabbox/bootstrapped'
```

While each lease ran, a sampler on the Parallels host recorded the argv of every `prlctl exec` process, so the transport is observed on the host rather than inferred.

**Fault injection without touching the template.** `CRABBOX_WORK_ROOT=/var/lib/crabbox/ssh.username` is a path that `InstallSSHKey` creates as a regular root-owned file just before guest prep. In `EnsureGuestReady`, `mkdir -p "$work_root"` then fails mid-script. The final `crabbox-ready` still passes, because `test -w` on a file is true for root. The fault can only be caught if `set -eu` is really in effect.

| Run | Binary | Fault | Guest-prep argv seen on the host | Result |
|---|---|---|---|---|
| happy | this PR | none | `prlctl exec <vm> /bin/sh -s` | lease ready, `GUEST_OK`, exit 0 |
| fault | this PR | work root is a file | `prlctl exec <vm> /bin/sh -s` | **`parallels guest prep: exit status 1: mkdir: /var/lib/crabbox/ssh.username: File exists`** |
| fault | `main` | same | `prlctl exec <vm> /bin/sh -lc set -eu\012user=…\012work_root=…` (whole script in argv) | prep **passes**, lease reports `provisioned`, then fails later as `create remote workdir: exit status 1` |

Happy path, this PR:

```
provisioned lease=cbx_5080a8f918bb vm=5933f005-… ip=10.211.55.86
GUEST_OK
26.5.2
-rw-rw-rw-  1 root  wheel  0 Sep 20 23:27 /var/lib/crabbox/bootstrapped
run summary lease=41.471s bootstrap=526ms sync=0s command=1.14s total=2.171s end_to_end=44.262s sync_skipped=true exit=0
lease cleanup stopped=true policy=auto lease=cbx_5080a8f918bb
```

Fault, this PR. It fails at the step that failed, and crabbox deletes the clone:

```
provisioning provider=parallels lease=cbx_c26822d8cad9 ... clone_mode=full keep=false
parallels guest prep: exit status 1: mkdir: /var/lib/crabbox/ssh.username: File exists
```

Fault, `main`. The failure is swallowed and surfaces later under the wrong name:

```
provisioned lease=cbx_75c5a7d9bc84 vm=c4e8b504-… ip=10.211.55.88
  workdir=/var/lib/crabbox/ssh.username/cbx_75c5a7d9bc84/crabbox
create remote workdir: exit status 1
```

So ssh does forward stdin to the remote `prlctl` intact. Both scripts reached the guest whole: the happy path provisioned and ran, and `set -eu` aborted the fault run at the right line.

One PR fault attempt died with the pre-existing `PrlJob_GetRetCode: Invalid argument` flake described above, while another VM was running on the host. The retry is the run shown.

**Cleanup.** Every clone was deleted, with a readback of `prlctl list --all` after each run. The template was left stopped, with the same snapshot count and the same current snapshot as before, compared against a pre-run readback.

**Linux:** there is no Linux Parallels template on this host either, so the Linux `apt-get` path still has no native coverage. That limit is unchanged, and accepting it is the maintainer's call.

## Scope limits

Please read the native evidence as covering exactly this and no more:

- **Exercised:** the local (host-resident `prlctl`) route **and** the remote route (`--parallels-host`, section above), both on a **macOS** guest, on one Apple-silicon host and one template.
- **Reasoned about, not exercised:** **Linux guests.** There is no Linux Parallels template on that host, so the `apt-get` stdin-consumption path — the single largest behavioral surface this change opens — has no native coverage at all. It is asserted in tree by string match only. Dropping `-l` also means a Linux guest now runs the script under `/bin/sh` proper (dash on Debian/Ubuntu) rather than whatever outer shell the flattened form landed in; `TestParallelsEnsureReadyScriptParsesUnderPOSIXShell` covers that syntactically, and on a Linux CI runner that check is dash.
- **Behavior change worth a maintainer's eye:** `set -eu` genuinely taking effect is the point of the fix, but it means preparation can now abort where it previously ran on. Anything meant to stay non-fatal must carry `|| true`; I audited both scripts and left the existing ones alone, but a guest-side failure that used to be silently survivable will now surface as a failed prep.

Contributor PR — no `CHANGELOG.md` edit, per `AGENTS.md`.


saariuslystoned:
@clawsweeper re-review

Rebased onto current `main` (`2ad31a5a`). The PR body now has native evidence for the remote-host route (`--parallels-host`): a normal lease, plus a fault-injected lease that fails at the failing step on this branch and is swallowed on `main`. It also confirms there is no Linux template on the host, so the Linux limit stands.

steipete:
Landed via https://github.com/openclaw/crabbox/pull/2431: your commits (authorship intact) rebased onto `main` after the caller-context diagnostics change in https://github.com/openclaw/crabbox/pull/2415 conflicted with this branch, plus a docs/changelog commit thanking you. Verified with the full race suite on a remote Crabbox AWS gate and CI. Thanks @saariuslystoned!

PR #2403, opened 2026-09-20
fix: preserve Azure and Hetzner heartbeat timeout overrides

steipete:
## Summary

Direct Azure and Hetzner heartbeats could keep a stored 30-minute idle timeout even when the caller explicitly requested 90 minutes. The public heartbeat path supplies `TouchRequest.IdleTimeoutOverride`, but the legacy shared wrapper retained only the server and state, dropping that intent before label encoding.

Carry the complete touch request through the shared best-effort method and use the existing override-aware label helper. An omitted override still preserves the stored policy and TTL cap. The adapters now own client construction and resource identifier selection; the shared method owns label updates and nonfatal warning output. Delete the old core dispatcher and its redundant wrapper.

This also removes an unconditional AWS `i-` identifier heuristic ahead of Azure routing. An Azure adapter request stays with Azure rather than selecting another cloud from its resource identifier. Azure still uses CloudID-or-Name, and Hetzner still uses its numeric ID. AWS, GCP, and Proxmox already have independent touch implementations, which are unchanged.

## Verification

The complete shared/Azure/Hetzner race suite passed 850 tests/subtests with no skips. The focused touch regressions passed 15 events, covering provider selection, explicit replacement versus omission, resource-key selection, and client/write failures; four existing core label-policy tests also passed. CLI/provider vet and a full CLI build passed. The pinned CI deadcode tool passed the exact `-test ./...` scope for Linux/amd64 with empty output. Independent isolated Codex review through P2 is clean.

All local commands used a private, credential-free environment with external networking denied. The initial local compile caught one leftover unused import, which was removed before the passing runs. An initial offline `go run` of the analyzer could not perform module metadata lookup; the same pinned cached analyzer binary subsequently passed.

No native Azure or Hetzner allocation/API run is claimed. Tests exercise the existing injected provider-client boundary and the exact updated metadata payload; production client implementations and service APIs are unchanged. Full exact-head CI remains required before landing.

## Scope

No new flags, configuration, credentials, retry behavior, claim mutation, or deletion behavior. Existing best-effort factory/write failures still return the touched server with a warning. Provider-authoring documentation records the ownership boundary, and the linked Unreleased entry is included. The final changelog-only commit preserves the tested source unchanged.


saariuslystoned:
### Real-behavior Hetzner proof for this PR

The review flagged that the adapters here are covered only by fake-client tests, with no
after-fix heartbeat through a real setup showing **persisted** timeout replacement and
omission. I had a funded Hetzner account available, so here is that run against the real
Hetzner Cloud API. Head under test: `3df4438edc81235973d8eb2ff59cb717a137cbb7`, contrasted
against `main` @ `f7bed55d`.

**Method.** One real `cx23` server allocated through crabbox's own lease path (so the labels
and claim are genuine), then heartbeated with binaries built from each commit. Labels were
read back **out-of-band** with `curl` against `api.hetzner.cloud/v1/servers/<id>` — not through
crabbox's own output — so the readback is independent of the code under test. The lease was
created with `--ttl 6h` on purpose: `expires_at = min(created+ttl, last_touched+idle)`, so at
the default 90m TTL a 90m idle replacement would have been clamped and invisible.

All rows are the same server, same lease, within ~2.5 minutes:

| # | binary | command | CLI reported | persisted `idle_timeout_secs` | `expires_at` |
|---|---|---|---|---|---|
| baseline | PR head, `warmup` | idle omitted | `30m` | **1800** | 22:08:05 |
| **C** | **main** | `heartbeat --idle-timeout 90m` | `30m0s` ❌ | **1800 — discarded** ❌ | 22:09:44 |
| **1** | **PR head** | `heartbeat --idle-timeout 90m` | `1h30m0s` ✅ | **5400 — replaced** ✅ | **23:09:57** |
| **2** | **PR head** | `heartbeat` (no flag) | `1h30m0s` ✅ | **5400 — preserved** ✅ | 23:10:11 |
| control | main | `heartbeat` (no flag) | `1h30m0s` | 5400 — preserved | 23:10:25 |

**Case 1 — replacement.** Rows C and 1 are the *same command against the same lease in the same
stored state*, 13 seconds apart, differing only in the binary. On `main` the explicit
`--idle-timeout 90m` is silently dropped — exit code 0, no warning, persisted label stays
`1800`. On this PR the same command persists `idle_timeout_secs=5400` and `expires_at` moves out
by a full 90 minutes. `expires_at - last_touched_at` equals `idle_timeout_secs` exactly in every
row.

**Case 2 — omission.** With the stored value at 5400 and the config default at 1800, a bare
`heartbeat` keeps **5400** rather than resetting to the config default, while `last_touched_at`
advances (21:39:57 → 21:40:11) and `expires_at` tracks it — so the heartbeat is doing real work,
not no-oping.

**Control — no regression.** The bare heartbeat on `main` also preserves 5400. Omission was
already correct; this PR changes only the explicit-replacement path and does not trade one
behavior for the other.

One implementation note worth recording: the run had to use an isolated config to force the
**direct** Hetzner path. With a coordinator registered, `loadBackend` returns a
`coordinatorLeaseBackend` and returns before `sshBackend.Touch`, so a brokered heartbeat never
exercises this code at all — worth knowing for anyone trying to reproduce this.

#### What this does not cover

- **The Azure half of this PR is still unproven.** No Azure allocation or API call was made. The
  Azure adapter change is structurally identical, but that is not evidence and I'd rather not
  have it read as such.
- No long-running expiry test — the lease was destroyed minutes after creation, so the persisted
  deadline is proven but actual reaper enforcement at 90m was not observed.
- Single region (`nbg1`), single server type, Linux only.

The server was destroyed immediately after the run; deletion confirmed out-of-band (API GET
returns `404 not_found`, account inventory back to its pre-run state). Total spend ≈ EUR 0.0005.
Full raw terminal output for every step is captured in our run packet; happy to paste any
individual step inline if useful.

PR #2425, opened 2026-09-21
fix(nvidia-brev): confirm asynchronous stops before reuse

vincentkoc:
## What Problem This Solves

Fixes an issue where stopping a retained NVIDIA Brev workspace reported success before the provider finished stopping it. Immediate reuse could attempt SSH against a workspace that was still shutting down.

## Why This Change Was Made

Brev's stop command only acknowledges an asynchronous request. Persist the stopping intent before dispatch, clear the SSH endpoint, and wait for organization-scoped inventory to confirm STOPPED. Interrupted or failed observation keeps a retryable claim; reuse respects the pending intent even while inventory briefly still reports RUNNING. Already-stopping and stopped workspaces do not receive duplicate stop requests.

The provider owns reconciliation; core's existing inactive-endpoint rule now includes stopping. Production delta: +58 lines for this lifecycle boundary. No config additions or new secret handling.

## User Impact

Stop success now means the workspace is stopped. Cancellation, provider errors, and organization changes preserve custody. Retrying stop reconciles the claim, and reusing a stopping workspace waits before restarting it.

## Evidence

- Focused NVIDIA Brev and conditional-claim tests passed.
- Regression cases cover delayed stop, cancellation during the request or observation, missing/failed inventory, organization change, repeated stop, stale claims, and reuse with a pending stop.
- `go test -race ./internal/providers/nvidiabrev -count=1 -timeout=3m` passed; hosted CI is pending.
- Direct upstream contract checked at Brev v0.6.335: https://github.com/brevdev/brev-cli/blob/1733b3d5822b4fdec88682c7a2358a45fbd66e32/pkg/cmd/stop/stop.go#L183-L196 and https://github.com/brevdev/brev-cli/blob/1733b3d5822b4fdec88682c7a2358a45fbd66e32/pkg/cmd/start/start.go#L397-L421.
- Live GPU smoke has not run: an authenticated Brev environment is not available in this task yet.

Part of the first-class Brev support work. Separate follow-ups cover current certificate SSH/authentication, ordinary status, retained host/container selection, pause/resume, recovery inventory, and live smoke coverage.

PR #2429, opened 2026-09-21
fix(nvidia-brev): resolve certificate routes with native SSH

vincentkoc:
Brev emits certificate-only `Match exec` routes that Crabbox's partial `Host` parser cannot represent reliably. This change delegates route interpretation to native OpenSSH and removes the partial parser.

The provider captures configuration inside its organization checks. Commands, sync, copying, forwarding, and interactive sessions use the captured bytes and reevaluate the certificate hook for each transport. A later Brev refresh cannot redirect the resolved operation. Session-owned snapshots remain available to ProxyJump until cleanup; printed native SSH/editor commands intentionally use Brev's current file.

Shared transport ownership preserves Pond and VNC/WebVNC listeners while excluding inherited forwards and remote commands. Missing configuration or failed required certificate renewal rejects dispatch. Route discovery now reports bounded, credential-redacted hook stderr even when `ssh -G` returns success without selecting the required route. No new user configuration or credential storage is added.

Validation at `0f574978378f429e002692535ec607e3e905ceb9`:
- Provider race suite passed (4.633s): `go test -race ./internal/providers/nvidiabrev -count=1 -timeout=3m`.
- Real OpenSSH transport proof passed (4.059s): `go test -race ./internal/cli -run '^TestSSHTransportCapturedConfigPreventsAliasReassignment$' -count=1 -timeout=2m -v`. Two named commands and an interactive shell reached endpoint A; after the source configuration reassigned the alias, endpoint B received zero connections.
- Native fixtures cover certificate routes, optional static-key fallback, repeated and failed renewal, user overrides, captured metadata, ProxyJump lifetime, forwarding, and printed commands. New diagnostic tests cover a real failing hook, redaction before truncation, and wrapped error identity.
- [Hosted CI](https://github.com/openclaw/crabbox/actions/runs/35597446827) passed for this head: source/test jobs passed in attempt 1; only Direct Go install needed attempt 2 after a sum.golang.org HTTP/2 download error, and that retry passed.
- Earlier transport revision passed the broader provider/SSH/Pond/VNC race suite. `git diff --check` passed.
- Production Go delta: +6 net lines; tests: +296 net lines. The production increase preserves bounded, actionable dependency diagnostics and route ownership; raw configuration bytes are excluded from JSON.
[...]

PR #2430, opened 2026-09-21
fix(nvidia-brev): preserve retained connection targets

vincentkoc:
A Brev lease created with `--nvidia-brev-target host` can silently switch to the container in a fresh Crabbox process. Claimed projections replace the stored target with the runtime default, and SSH resolution independently uses that default. Heartbeats can also erase the label.

Resolve one effective connection target from explicit current input, then the retained lease, then the default. Track accepted file/environment/flag assignments so an explicit `container` still overrides a stored `host`; preserve that explicit default in follow-up commands. Claimed projections and SSH resolution use the same precedence, and incomplete heartbeat metadata preserves the stored target.

Validation:
- `go test ./internal/providers/nvidiabrev ./internal/cli -run NvidiaBrev -count=1 -timeout=3m` passed.
- Regression follows acquire-host, fresh-process status/ready-probe/run/list, heartbeat with missing target, stop, fresh-process resume, and explicit switch to container, including the resulting follow-up command.
- File/environment tests cover an explicit default and ignored empty assignments; existing flag metadata tests cover accepted visits.
- `go generate ./internal/cli/config_nvidia_brev.go` and `git diff --check` passed. Independent source review found no actionable issue.

Production delta: +30 handwritten Go lines and +2 generated lines, restoring the retained connection-selection contract. No new public config/default surface, credentials, or provider API is introduced. Authenticated GPU smoke remains pending account selection.

Related lifecycle and SSH fixes: https://github.com/openclaw/crabbox/pull/2425 and https://github.com/openclaw/crabbox/pull/2429.


steipete:
Landed via #2517 (merged as c0c14f406), which rebuilt this change on current main with a full remote race-suite gate. Your fix and analysis are preserved with `Co-authored-by` credit in the squash commit and a changelog thank-you. Closing this original in favour of the superseding PR. Thanks @vincentkoc!

PR #2463, opened 2026-09-21
fix(lambda): preserve acquisition causes after cleanup failure

steipete:
## Summary

Lambda acquisition rollback formatted the primary error as text while wrapping only the cleanup error. An SSH timeout could therefore lose its exit code 5 and cancellation could lose its discoverable cause.

Reuse `shared.JoinAcquireCleanupError` rather than the private wrapper. Both causes and primary exit-code precedence survive, and cleanup failure explicitly vetoes an automatic fresh-allocation retry in that acquisition. Simply preserving the original typed timeout without that marker would allow a retry despite cleanup debt. Successful rollback still permits the existing bootstrap retry.

The diagnostic prefix stays intact; joined errors use newlines and the shared retry wrapper emits a cleanup warning. This keeps cleanup details visible when the CLI prints the primary typed error. If recovery-claim persistence and rollback both fail, retain the claim-write error too; previously that error was silently discarded. Typed errors are considered in acquisition, cleanup, then claim-write order. Consequently an otherwise untyped failure can now report the claim-write exit code 2 instead of fallback 1. Successful rollback still returns the original failure unchanged. No mutation ordering, claim authority, ambiguity reconciliation, or fixed-ID behavior changes.

## Verification

- Baseline regression cases failed for both `keep` modes: timeout exit code became 1 or cleanup code 9 instead of 5; cancellation identity was lost.
- Full Lambda/shared race suites and vet pass.
- Tests cover termination and key-cleanup failures, primary/cleanup identities and code precedence, retained local claim/key state, no fresh allocation after failure, and retry after successful rollback.
- New deterministic regressions block the claim directory with a regular file after acquisition: all three combined-failure cases failed before the follow-up, and the successful-rollback control passed. The fixed full Lambda/shared race suites and vet pass, including typed precedence and successful-cleanup retry eligibility.
- Complete independent Codex review of the entire updated candidate found no actionable issues through P2.

## Qualification status

Native lifecycle qualification is pending because Lambda test credentials are not configured in the current verification environment. No provider API or resource operations were performed. This remains a draft; unit evidence is not presented as a native pass. No credentials, account permissions, dependencies, or release machinery change.


steipete:
Maintainer note: this PR and its Unreleased entry were prepared by steipete as maintainer work. The repository rule reserves contributor changelog edits for maintainers; it does not prohibit maintainer entries. I am retaining the changelog entry. The separate runtime/native proof request remains open, and the PR stays draft—unit evidence is not being substituted for that qualification.

